PR #33, opened 2019-11-01
Add Prettier to format code

EvanHahn:
This adds [Prettier][1], a code style formatter. This will help keep the code consistent, and helps prevent [bikeshedding][2] about code formatting (such as tabs versus spaces, single quotes versus double quotes, etc).

You can run the formatter manually with `yarn run format`, but it also runs automatically on Git commit, following [Prettier's docs][3].

Most of the changes in this commit are due to Prettier. The files that meaningfully changed were:

* `.huskyrc.json`
* `.prettierignore`
* `.prettierrc.json`
* `package.json`

[1]: https://prettier.io/
[2]: https://en.wiktionary.org/wiki/bikeshedding
[3]: https://prettier.io/docs/en/precommit.html#option-2-pretty-quick

### DIFF
--- a/.huskyrc.json
+++ b/.huskyrc.json
@@ -1,0 +1,5 @@
+{
+  "hooks": {
+    "pre-commit": "pretty-quick --staged"
+  }
+}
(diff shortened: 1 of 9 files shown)


positiveimpact:
This definitely makes sense to have consistent code formatting

PR #37, opened 2019-11-04
add json file w/ first project, add project images

positiveimpact:
Created a json file containing the data of an example project. This json data can be used for testing and will eventually be replaced by data from our postgres database


data_temp/projects.json
```diff
@@ -0,0 +1,95 @@
+{
```
EvanHahn: It doesn't look like Prettier formatted this JSON file. Did you run `yarn format`? Or did it not automatically run when you ran `git commit`?

positiveimpact: It doesn't seem like it ran. It might be because I used the git extension interface in VS Code to commit it. I made a commit that prettifies it.

positiveimpact:
@EvanHahn 
**1. I would see this in 3 steps:**

1. For the overview on the main page we'll need `name, image, organisation_name, organisation_image, location, impact, recurring_yearly, labels`

2. For the overview on the project's page, we'll additionally need `number_of_followers, creation_date, collaborators_accepted, helpful_skills, helpful_contacts`

3. For the "Project" tab on the project's page, we'll need the rest of the fields. 

I'd say 1. and 2. are part of the MVP

**2. Good point. I don't think it actually needs to be committed/merged.**

Issue #40, opened 2019-11-05
Create Landing Page Layout


ReeceLangerock:
Create a layout component for the landing page. Will include:

Header component
Grid section for projects 
- currently only featured, may add more in future
Browse all button

Issue #39, opened 2019-11-05
Create project preview component


ReeceLangerock:
Create a component that encompasses the elements in the below image

![image](https://user-images.githubusercontent.com/23420334/68172132-0051eb00-ff3c-11e9-976d-511461650f2b.png)

Issue #38, opened 2019-11-05
Create generic button component


ReeceLangerock:
Using Material UI as a base, create a generic button component to be used across the app.

 Needs outline and solid style options

EvanHahn:
Closing in favor of #65. We'll just use [Material UI's Button component](https://material-ui.com/components/buttons/#button).

PR #48, opened 2019-11-22
Feature/landing page layout

ReeceLangerock:
Addresses these tickets part of in full: #40 #38 #39


public/placeholder.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="23.5" height="30.821" viewBox="0 0 23.5 30.821"><defs><style>.a{fill:#484848;}</style></defs><g transform="translate(-76)"><g transform="translate(76)"><g transform="translate(0)"><path class="a" d="M87.75,0C81.271,0,76,4.861,76,10.836a10.2,10.2,0,0,0,1.754,5.7l9.327,13.861a1,1,0,0,0,.832.427h.008a1,1,0,0,0,.832-.439l9.089-14A10.193,10.193,0,0,0,99.5,10.836C99.5,4.861,94.229,0,87.75,0Zm8.412,15.46L87.9,28.184l-8.48-12.6a8.507,8.507,0,0,1-1.473-4.746c0-4.979,4.406-9.042,9.8-9.042s9.8,4.063,9.8,9.042A8.5,8.5,0,0,1,96.162,15.46Z" transform="translate(-76)"/></g></g><g transform="translate(81.875 5.418)"><g transform="translate(0)"><path class="a" d="M171.875,90a5.436,5.436,0,1,0,5.875,5.418A5.667,5.667,0,0,0,171.875,90Zm0,9.042a3.635,3.635,0,1,1,3.923-3.624A3.789,3.789,0,0,1,171.875,99.042Z" transform="translate(-166 -90)"/></g></g></g></svg>
```
EvanHahn: What's the licensing on these SVGs? Did you make them yourself?

src/components/general/Header.js
```diff
@@ -6,10 +6,10 @@ const Header = () => (
...
     <img src="https://climateconnect.earth/logo.png"></img>
     <LinkContainer>
-      <Link href="forum" text="Forum" passHref></Link>
-      <Link href="browse" text="Browse" passHref></Link>
-      <Link href="create" text="Create A Project" passHref></Link>
-      <Link href="signin" text="SIGN IN" passHref type="button"></Link>
+      <Link href="forum" text="Forum"></Link>
+      <Link href="browse" text="Browse"></Link>
+      <Link href="create" text="Create A Project"></Link>
+      <Link href="signin" text="SIGN IN" type="button"></Link>
```
EvanHahn: Nitpick: this capitalization is stylistic, not semantic. Could we do one of the following?

1. Add `text-transform: uppercase` or similar
1. Add a TODO/issue to do that?

src/components/general/Button.js
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from "react";
+import styled from "styled-components";
+
+export default class Button extends Component {
+  render() {
+    const { type, text } = this.props;
```
EvanHahn: It seems slightly more idiomatic to make the text be passed in as children instead of as an explicit prop. In other words, the interface would look like this:

```jsx
<Button type="whatever">Hello world!</Button>
```

If you like that, I _think_ we'd just need to replace `text` with `children` here and update usages.

What do you think?

src/components/project/ProjectPreview.js
```diff
@@ -0,0 +1,83 @@
+import React from "react";
+import styled from "styled-components";
+import Button from "../general/Button";
+
+export default function ProjectPreview({ data }) {
+  const rand = Math.ceil(Math.random() * 300);
```
EvanHahn: Would you mind adding a comment explaining that this is temporary?

test/components/layouts/indexlayout.test.js
```diff
@@ -0,0 +1,26 @@
...
+  it("contains a browse button", () => {
+    const wrapper = shallow(<IndexLayout />);
+    expect(wrapper.find(Link)).toHaveLength(1);
+    expect(wrapper.find("button")).toHaveLength(1);
+    expect(wrapper.find("button").text()).toEqual("Browse");
+  });
+
+  it("contains correct grid components", () => {
+    // const wrapper = shallow(<IndexLayout />);
+    // add test here when Grid Container component is completed
```
EvanHahn: Would you mind adding `TODO` here for greppability?

test/components/layouts/indexlayout.test.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { shallow } from "enzyme";
+import IndexLayout from "./../../../src/layouts/IndexLayout";
+import Link from "next/link";
+
+describe("Layout Component Tests", () => {
```
EvanHahn: Nitpick: I don't think the word "tests" adds anything here...could we change it to just `describe("Layout component", ...` or `describe("IndexLayout", ...`?

src/reset.css
```diff
@@ -9,3 +9,58 @@ body {
   margin: 0;
   font-family: "Segoe UI", Tahoma, Geneva, Verdana, sans-serif;
 }
+button {
+  font-family: "Segoe UI", Tahoma, Geneva, Verdana, sans-serif;
```
EvanHahn: I think this should be `font-family: inherit`. That way, if we change the font in the `html`/`body`, we won't forget to update the button.

src/components/project/ProjectPreview.js
```diff
@@ -0,0 +1,83 @@
+import React from "react";
+import styled from "styled-components";
+import Button from "../general/Button";
+
+export default function ProjectPreview({ data }) {
```
EvanHahn: Could `data` have a more descriptive name, like `project`?

src/layouts/IndexLayout.js
```diff
@@ -0,0 +1,56 @@
...
+    return <ProjectPreview data={project} key={project.name}></ProjectPreview>;
+  });
+}
+
+export default function index() {
+  return (
+    <Container>
+      <h1>Share and Collaborate on projects to reach the maximum positive impact on the world</h1>
+
+      <h3>Featured</h3>
```
EvanHahn: Because there's an `h1` above and `h3`s in `ProjectPreview`, should this be an `h2` instead?

And if the styling on the `h2` is wrong for this spot, we should change the styles, not the tag—it's usually harder to add accessibility later.

src/layouts/IndexLayout.js
```diff
@@ -0,0 +1,56 @@
+import React from "react";
+import styled from "styled-components";
+import Link from "next/link";
+import TEMP_FEATURED_DATA from "./../../projects.json";
+import ProjectPreview from "./../components/project/ProjectPreview";
+import Button from "../components/general/Button.js";
+
+function renderProjects() {
```
EvanHahn: Could this instead be a component that returns a fragment?

```jsx
function ProjectPreviewList() {
  return (
    <>
      {TEMP_FEATURED_DATA.projects.map(project => (
        <ProjectPreview ... />
      )}
    </>
  );
}
```

src/layouts/IndexLayout.js
```diff
@@ -0,0 +1,56 @@
...
+  // TEMPORARY DATA
+  return TEMP_FEATURED_DATA.projects.map(project => {
+    return <ProjectPreview data={project} key={project.name}></ProjectPreview>;
```
EvanHahn: Projects will eventually have unique IDs. Could we do one of the following?

1. Add fake `id`s to the sample project data, and then use that for the `key` here
1. Add a TODO, saying that we intend to use the `id` for the `key` instead

src/layouts/IndexLayout.js
```diff
@@ -0,0 +1,56 @@
...
+        {renderProjects()}
+        {/* Added to preview multiple projects */}
+        {renderProjects()}
+        {renderProjects()}
+        {renderProjects()}
+      </ProjectGrid>
+
+      <Link href="Browse">
+        <Button text="Browse" type="big">
+          {" "}
```
EvanHahn: Why is this here?

src/layouts/IndexLayout.js
```diff
@@ -0,0 +1,56 @@
...
+import Button from "../components/general/Button.js";
+
+function renderProjects() {
+  // TEMPORARY DATA
+  return TEMP_FEATURED_DATA.projects.map(project => {
+    return <ProjectPreview data={project} key={project.name}></ProjectPreview>;
+  });
+}
+
+export default function index() {
```
EvanHahn: Why is this `index` and not `IndexLayout`?

.gitignore
```diff
@@ -23,3 +23,5 @@
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+
+projects.json
```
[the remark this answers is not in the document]
ReeceLangerock: added

public/placeholder.svg
ReeceLangerock: Not sure, I took them from the mockup. @positiveimpact ?

src/components/general/Button.js
ReeceLangerock: Yeah that way is better / simpler. I was doing the same as the Link component which can be changed as well.

src/components/general/Header.js
ReeceLangerock: Good call, did option 1 for `type="button"`

src/layouts/IndexLayout.js
ReeceLangerock: prettier being weird?

src/layouts/IndexLayout.js
ReeceLangerock: That seems like a bit of overkill, I generally like to keep render functions like this in the same component. Especially when this will most likely be the only one and it's fairly simple.

ReeceLangerock:
Forgot that I hadn't written tests for these components, I'll do that before re-requesting a merge.

public/placeholder.svg
positiveimpact: All the icons are from flaticon.com. When we launch, we'll have either to pay a monthly base fee of 7.50€ for using all the icons or credit the author of each icon. Alternatively, I could also look for a resource for fully free icons.

positiveimpact: I've had a look at free icons and couldn't find many good ones. However, if we have to credit multiple authors, our footer will become really packed. Most icons I used are by https://www.flaticon.com/authors/smashicons. I suggest we credit them in the footer and I make sure all icons are by this author.

EvanHahn: I have a preference for free icons. Have we tried icons from Fork Awesome, or IconFinder, or IconMonstr, or just using emojis?

positiveimpact: Thanks for the suggestions, I'll try if I can switch out all the icons with free icons!

test/components/general/header.test.js
```diff
@@ -16,15 +16,17 @@ describe("Header Component Tests", () => {
...
-      { href: "forum", passHref: true, text: "Forum" },
-      { href: "browse", passHref: true, text: "Browse" },
-      { href: "create", passHref: true, text: "Create A Project" },
-      { href: "signin", passHref: true, text: "SIGN IN", type: "button" }
+      { href: "forum" },
+      { href: "browse" },
+      { href: "create" },
+      { href: "signin", type: "button" }
     ];
+    const expectedTexts = ["Forum", "Browse", "Create A Project", "SIGN IN"];
```
EvanHahn: Does this test still pass with the capitalized text?

ReeceLangerock: It did not, made a change so it does. I noticed I forgot tests for the link component, so i'll add that later and add tests for the correct text-transform class

Issue #41, opened 2019-11-05
Create progress component


ReeceLangerock:
Relates to #14 

Create a generic progress display component

positiveimpact:
I put the progress component in the design, but tbh I wasn't able to think of a good way to determine what to show there. We will still need this component, for example for [this page](https://xd.adobe.com/view/395518bf-7e51-4eb2-6ff2-3f1b57fdaa17-4f64/screen/7ecacdee-ef87-420b-af02-5308a40be120/Create-Account-Interface-Step-2).

EvanHahn:
Closing in favor of #65. Material UI already has a [progress component](https://material-ui.com/components/progress/).

PR #53, opened 2019-12-09
First page of create a project flow

plantfansam:
This is my first JS in quite some time, so please don't hesitate to tear this thing apart! For reference, [here's the mockup I was working off of](https://xd.adobe.com/view/395518bf-7e51-4eb2-6ff2-3f1b57fdaa17-4f64/screen/8080cb18-a8ac-43d2-ad78-8512918a0e29/Create-Project-Step-1/). It looks like this:

image

Here's what the PR does:

* Adds a new `<CreateProjectLayout>` component, rendered at `/create`, whose state is manipulated by the organization, project location, and project name inputs
* Adds the light gray color used in the border to the theme
* A small correction from `organisation` -> `organization` (couldn't help myself)
* Adds an `onClick` handler to the `Button` component
* Adds a rudimentary `ProgressBar` component (see #41)
* Updates yarn.lock after updates to `package.json`

The PR _does not_:

* represent a pixel-perfect rendering of the mockup
* have good tests of the child components' state manipulation
* properly integrate into whatever gridding system we want to use (material ui? native css grids?)

Known issues:

* Making a selection from the dropdown causes the container to grow slightly in width


projects.json
```diff
@@ -4,7 +4,7 @@
       "name": "CO2-Labels for University Canteen",
       "image": "/data_temp/images/canteen.png",
       "organisation_name": "sneep Erlangen e.V.",
-      "organisation_image": "/data_temp/images/sneep.png",
+      "organization_image": "/data_temp/images/sneep.png",
```
EvanHahn: IMO, we should undo this change, as it is (1) inconsistent with `organisation_name` above (2) unrelated to the overall goal of this PR. Updating this probably also means we'll update `organization_image` elsewhere in this PR.

Relatedly: as we work on this project, I think we'll need to decide whether we internally use American English or British English for various spellings. I don't think it really matters as long as we pick one, and that decision is much less important than what we display _publicly_ (which should, ideally, be internationalized).

src/components/general/Button.js
```diff
@@ -4,7 +4,11 @@ import styled from "styled-components";
 export default class Button extends Component {
   render() {
     const { type, children } = this.props;
-    return <StyledButton type={type}>{children}</StyledButton>;
+    return (
+      <StyledButton onClick={this.props.clickHandler} type={type}>
```
EvanHahn: I think that this should be `onClick` instead of `clickHandler` for consistency.

We should also add this to  destructuring on line 6, instead of grabbing it here. In other words, I think this should be:

```jsx
const { type, children, onClick } = this.props;
return (
  <StyledButton onClick={onClick} type={type}>
  // ...
```

package.json
```diff
@@ -23,17 +23,20 @@
   },
   "dependencies": {
     "@material-ui/core": "^4.5.2",
+    "@material-ui/icons": "^4.5.1",
```
EvanHahn: I don't see this module used. Should we remove it?

package.json
```diff
@@ -23,17 +23,20 @@
...
     "@zeit/next-css": "^1.0.1",
+    "material-ui-icons": "^1.0.0-beta.36",
```
EvanHahn: Same as above—I don't see this module used. Should we remove it?

test/components/general/button.test.js
```diff
@@ -12,4 +12,15 @@ describe("Button Component", () => {
     const wrapper = shallow(<Button type="big">Button</Button>);
     expect(wrapper.props()).toEqual({ children: "Button", type: "big" });
   });
+
+  it("sets onClick to the value of the clickHandler prop", () => {
```
EvanHahn: Nitpick: I don't think this is really the thing you're testing here—would a spec title like "calls the click handler when clicked" be better?

src/components/general/ProgressBar.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+}));
+
+export default function ProgressBar(props) {
+  const classes = useStyles();
+
+  return (
+    <div className={classes.root}>
+      {props.progressPercentage}%
+      <LinearProgress variant="determinate" value={props.progressPercentage} />
```
EvanHahn: Out of curiosity, what does `variant="determinate"` do? [The docs I looked at](https://material-ui.com/api/linear-progress/#linearprogress-api) weren't super clear.

src/components/project/ProjectBasicsPane.js
```diff
@@ -0,0 +1,32 @@
...
+import FormControl from "@material-ui/core/FormControl";
+import TextField from "@material-ui/core/TextField";
+import OrganizationSelect from "../../components/general/OrganizationSelect.js";
+
+export default function ProjectBasicsPane(props) {
+  return (
+    <>
+      <OrganizationSelect
+        organizations={props.organizations}
+        handleChange={props._handleOrganizationChange}
```
EvanHahn: It feels a little weird to pass a property prefixed with an underscore like this—that typically indicates that something is private. What's the reasoning for doing this?

src/layouts/CreateProjectLayout.js
```diff
@@ -0,0 +1,80 @@
...
+import TEMP_CATEGORY_DATA from "./../../tmp-data/categories.json";
+
+class CreateProjectLayout extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      organizations: TEMP_ORGANIZATION_DATA.organizations,
+      categories: TEMP_CATEGORY_DATA.categories,
+      progressPercentage: 0,
+      organization_id: "",
```
EvanHahn: JavaScript identifiers are typically `camelCased`. Is there a reason we're using underscores here?

test/.eslintrc.json
```diff
@@ -1,5 +1,9 @@
 {
   "env": {
+    "es6": true,
+    "node": true,
+    "browser": true,
```
EvanHahn: The `node` and `browser` options seem mutually exclusive to me. Given that the tests (currently) run in Node, I think we should omit `browser`, but let me know if there's a reason to have both that I'm missing.

test/components/general/progressbar.test.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+import { shallow } from "enzyme";
+import ProgressBar from "../../../src/components/general/ProgressBar";
+
+describe("ProgressBar component", () => {
+  it("renders the percentage as text", () => {
+    let percentage = 30;
+    const wrapper = shallow(<ProgressBar progressPercentage={percentage}></ProgressBar>);
+    expect(wrapper.text()).toContain(`${percentage}%`);
```
EvanHahn: Nitpick: IMO, we should hard-code `30%` here. For example, imagine we changed the interface to accept a fraction between 0 and 1—we wouldn't want to render `.3%`.

src/components/general/ProgressBar.js
```diff
@@ -0,0 +1,24 @@
...
+    textAlign: "center"
+  }
+}));
+
+export default function ProgressBar(props) {
+  const classes = useStyles();
+
+  return (
+    <div className={classes.root}>
+      {props.progressPercentage}%
```
EvanHahn: Should this be rounded? For example, what if someone passes a percentage of `1.23`? Should that render as `1.23%`, or `1%`?

src/components/project/ProjectBasicsPane.js
```diff
@@ -0,0 +1,32 @@
...
+        value={props.organization_id}
+      ></OrganizationSelect>
+      <FormControl fullWidth>
+        <TextField
+          id="project_name"
```
EvanHahn: HTML element IDs are global, and subtle weird things can happen when multiple elements have the same ID. Could we give these random IDs instead?

I could see defining a function called `randomIdForLabel` to make the intent clearer.

```jsx
function randomIdForInputLabel() {
  return `${Date.now().toString()}-${Math.random().toString()}`;
}

// ...

<TextField
  id={randomIdForInputLabel()}
  // ...
/>
```

The same comment applies for other `id`s in this PR.

src/components/project/ProjectBasicsPane.js
```diff
@@ -0,0 +1,32 @@
+import React from "react";
+import FormControl from "@material-ui/core/FormControl";
+import TextField from "@material-ui/core/TextField";
+import OrganizationSelect from "../../components/general/OrganizationSelect.js";
```
EvanHahn: Nitpick: we shouldn't add the file extension for JS imports—in other words, we can remove the `.js` suffix. This isn't a huge deal, but lets us more easily change the file extension in the future. We might do that if we switch to TypeScript as one example.

This nitpick applies to a few other places in this PR, too.

src/layouts/CreateProjectLayout.js
```diff
@@ -0,0 +1,80 @@
...
+  _handleProjectNameChange = event => {
+    this.setState({ project_name: event.target.value });
+  };
+
+  _handleProjectLocationChange = event => {
+    this.setState({ project_location: event.target.value });
+  };
+
+  _incrementPane = () => {
+    this.setState({ active_pane: this.state.active_pane + 1 });
```
EvanHahn: If the new state depends on the previous state (like it does here), it's best to call `setState` with a function.

```js
this.setState((state) => ({
  activePaneIndex: state.activePaneIndex + 1
}))
```

In short, this is useful if multiple `setState`s get called in rapid succession. You can read more about that [in React's docs](https://reactjs.org/docs/react-component.html#setstate).

src/layouts/CreateProjectLayout.js
```diff
@@ -0,0 +1,80 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      organizations: TEMP_ORGANIZATION_DATA.organizations,
+      categories: TEMP_CATEGORY_DATA.categories,
+      progressPercentage: 0,
+      organization_id: "",
+      project_name: "",
+      project_location: "",
+      active_pane: 0
```
EvanHahn: Should this be called `activePaneIndex` instead?

test/components/layouts/CreateProjectLayout.test.js
```diff
@@ -0,0 +1,16 @@
...
+describe("Create project layout", () => {
+  it("contains correct header text", () => {
+    const wrapper = shallow(<CreateProjectLayout />);
+    expect(wrapper.find("h1")).toHaveLength(1);
+    expect(wrapper.find("h1").text()).toEqual("Create a Project");
+  });
+
+  it("Updates state when organization dropdown value is changed", () => {
+    // const wrapper = mount(<CreateProjectLayout />);
+    // todo - what is a good way to test this?
```
EvanHahn: IMO, we should just remove this test for now.

test/components/general/OrganizationSelect.test.js
```diff
@@ -0,0 +1,24 @@
...
+      organization_image: "tmp-data/assets/images/sneep.png",
+      id: 44
+    },
+    {
+      name: "Climate Action Tech",
+      organization_image: "tmp-data/assets/images/climate-action-tech.png",
+      id: 45
+    }
+  ];
+  it("renders each organization as a MenuItem", () => {
```
EvanHahn: I'm new to React snapshot testing. Is this the kind of spec title that is used? It seems like we could completely reorganize the component to not render `MenuItem`s and the test could still pass if we updated the snapshot.

src/layouts/CreateProjectLayout.js
```diff
@@ -0,0 +1,80 @@
+import React, { Component } from "react";
+import Grid from "@material-ui/core/Grid";
+import styled from "styled-components";
+
+import ProgressBar from "../../src/components/general/ProgressBar.js";
+import ProjectPaneContents from "../../src/components/project/ProjectPaneContents.js";
```
EvanHahn: Where is this component? I wasn't able to find it.

src/components/general/OrganizationSelect.js
```diff
@@ -0,0 +1,31 @@
...
+import InputLabel from "@material-ui/core/InputLabel";
+import MenuItem from "@material-ui/core/MenuItem";
+import Select from "@material-ui/core/Select";
+
+export default function OrganizationSelect(props) {
+  return (
+    <FormControl fullWidth>
+      <InputLabel id="organization-select-label">Organization</InputLabel>
+      <Select
+        labelId="organization-select-label"
```
EvanHahn: Why is this needed?

src/layouts/CreateProjectLayout.js
```diff
@@ -0,0 +1,80 @@
+import React, { Component } from "react";
+import Grid from "@material-ui/core/Grid";
+import styled from "styled-components";
+
+import ProgressBar from "../../src/components/general/ProgressBar.js";
```
EvanHahn: It looks like we're backing out to the project root before coming back in. Could we shorten this import?

```jsx
 import ProgressBar from "../components/general/ProgressBar"; 
```

This comment applies to a few other imports in this PR, too.

EvanHahn:
@plantfansam There hasn't been movement on this PR in ~a month. If it's alright with you, I'll take it over and get it merged. If not, let me know in the next couple of days—and no problem if you'd rather finish it up!

positiveimpact:
Closed as we've done many changes to the code infrastructure since this PR has been made. Will remake this using the Form component from #86

PR #54, opened 2019-12-14
Add ESLint rule for self-closing React tags

EvanHahn:
Components without children can be self-closed to avoid unnecessary extra closing tag. For example:

```jsx
// This:
<Foo />

// Instead of this:
<Foo></Foo>
```

This enables the [`react/self-closing-comp` ESLint rule][1] to enforce this, and fixes this throughout the codebase.

[1]: https://github.com/yannickcr/eslint-plugin-react/blob/HEAD/docs/rules/self-closing-comp.md

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,6 +19,7 @@ module.exports = {
   plugins: ["react"],
   rules: {
     "react/prop-types": "off",
+    "react/self-closing-comp": "error",
     "react/prefer-stateless-function": "error"
   },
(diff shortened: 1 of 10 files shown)


plantfansam:
I think we're gonna need to do a little rebasing on this ... @EvanHahn yell at me if the repo dismisses my :+1: and I can re-:+1:

EvanHahn:
Rebased and merged! Thanks @plantfansam.

PR #66, opened 2019-12-22
Use Material UI across the site

### LINKED ISSUE #65: Use Material UI across the site

EvanHahn:
Screenshot

(The logo looks strange in the screenshot, but that's only an issue in the screenshot.)

This updates the app to use Material UI across the site. We had previously installed it as a dependency, but this uses it for all of our UI.

A few other things changed as a result of this commit:

* Removed `styled-components`, `reset.css`, and many styles in favor of Material UI's CSS setup
* Removed our custom `Button` component in favor of Material UI's built-in one
* Removed the Browse page (easy to add later)
* Removed `IndexLayout`—just define the layout in `pages/index.js`
* Fixed the logo URL
* Fixes a warning in development where `className`s were mismatched until the first hot-reload, causing styling errors

Closes #65.


src/components/layout.js
```diff
@@ -1,40 +1,41 @@
...
+
   return (
     <>
       <Head>
         <title>{title}</title>
         <link rel="icon" href="/favicon.ico" />
       </Head>
       <ThemeProvider theme={theme}>
-        <LayoutContainer>
+        <Container maxWidth="lg">
```
positiveimpact: Why did you limit the width of this container? I feel like it makes sense if the header and footer stretch over the whole page.

src/components/project/ProjectPreview.js
```diff
@@ -1,87 +1,56 @@
...
+  ListItemText
+} from "@material-ui/core";
+import PlaceIcon from "@material-ui/icons/Place";
+import BarChartIcon from "@material-ui/icons/BarChart";
+import { makeStyles } from "@material-ui/core/styles";
+
+const useStyles = makeStyles({
+  media: {
+    height: 140
+  }
```
positiveimpact: I think displaying the image to the left over the whole height and with a limited width is a better approach, because it uses the space of the card more efficiently. Also by just showing a slim portion of the image over the whole width, the most important part of the image might be cut out. So I suggest putting the image to the left over the whole height and the text to the right.

src/components/project/ProjectPreview.js
```diff
@@ -1,87 +1,56 @@
...
+              </ListItemIcon>
+              <ListItemText primary={project.location} />
+            </ListItem>
+            <ListItem>
+              <ListItemIcon>
+                <BarChartIcon />
+              </ListItemIcon>
+              <ListItemText primary={`Impact: ${project.impact}`} />
+            </ListItem>
+          </List>
```
positiveimpact: Why did you remove the organisation name?

test/components/project/projectpreview.test.js
```diff
@@ -1,35 +1,15 @@
...
-    expect(spans.at(1).text()).toEqual(project.location);
-    expect(spans.at(2).text()).toEqual(project.impact.toString());
-  });
-
-  it("contains correct button", () => {
-    expect(wrapper.find(Button)).toHaveLength(1);
-    expect(wrapper.find(Button).props()).toEqual({ children: "Get Involved", type: "outlined" });
+    const wrapper = mount(<ProjectPreview project={project} />);
+    expect(wrapper.text()).toContain(project.name);
+    expect(wrapper.text()).toContain(project.location);
```
positiveimpact: If you choose to readd the organisation name, you should also add it to this test

PR #68, opened 2020-01-16
Make the header responsive

EvanHahn:
This makes the `<Header />` component responsive on narrow screens (like mobile phones). On wide screens, it uses the buttons as before. On narrow screens, it renders an expandable drawer.

Screencast of what it looks like on iPhone

Screencast of what it looks like on dekstop

If you're on a narrow screen (such as a phone), we briefly render the "real" buttons before JavaScript can kick in and hide them. I felt that this was acceptable for environments with slow connections (they should be able to click the buttons while the page is loading) or environments without JavaScript. However, it's a bit wonky, so we should consider filing an issue or fixing it now.

### DIFF
--- a/src/components/general/Header.js
+++ b/src/components/general/Header.js
@@ -1,7 +1,36 @@
-import React from "react";
+import React, { useState } from "react";
+import noop from "lodash/noop";
 import Link from "next/link";
-import { Box, Container, Button } from "@material-ui/core";
+import {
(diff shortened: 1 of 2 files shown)


positiveimpact:
I'll create an issue for not showing the buttons until javascript kicks in. It will become more relevant once there are more links than the a phone screen can fit next to each other

PR #70, opened 2020-01-21
Added a coming soon message on Sign In page.

ddhanesha:
Screen Shot 2020-01-20 at 4 47 43 PM

@EvanHahn as discussed I added a banner to show coming soon banner. Let me know if there are any issues with it.

### DIFF
--- a/pages/signin.js
+++ b/pages/signin.js
@@ -4,7 +4,7 @@ import { Typography } from "@material-ui/core";
 
 const Home = () => (
   <Layout title="Sign In">
-    <Typography align="center" variant="h5" color="primary">
+    <Typography align="center" variant="h5" color="primary" component="h2">
       This page is coming soon!
     </Typography>
   </Layout>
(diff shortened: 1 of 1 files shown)


pages/signin.js
```diff
@@ -1,9 +1,12 @@
 import React from "react";
 import Layout from "../src/components/layout";
+import { Typography } from "@material-ui/core";
 
 const Home = () => (
   <Layout title="Sign In">
-    <h1>Sign In</h1>
+    <Typography align="center" variant="h5" color="primary">
```
EvanHahn: This should look like an `h5` but it should be the next level down from the existing header for accessibility, so I believe this should actually be an `h2`.

To achieve that, I believe you'll keep `variant="h5"` but add `component="h2"`. Worth double-checking that in the browser, though.

ddhanesha: Yep, you are right. Added `component="h2"`

PR #49, opened 2019-11-29
About us page

positiveimpact:
I implemented the about page which will simultaneously be used as a landing page during development.

I was unsure about the following things:
1. Where to store the "info" and "members" jsons, that are currently in AboutLayout.js
2. How to best go about deploying this during development. During development we only want this page  and a few select other pages to be accessible (onepager.pdf, faq).
3. On which level to import fontawesome. It will most likely be used on many pages, but maybe not on all. For now I only imported it on about.js as it's only being used there atm. But eventually all icons will be switched out by fontawesome icons, so maybe it would make sense to directly put it in layout.js.

And I haven't written test yet, but I decided to first get feedback on this and write test later, incase I need to change  a lot of things.

PR #72, opened 2020-01-29
Style ProjectPreview Cards

### LINKED ISSUE #13: Implement project preview list on the front page
This page: https://xd.adobe.com/view/395518bf-7e51-4eb2-6ff2-3f1b57fdaa17-4f64/

positiveimpact:
Restyled the ProjectPreview cards. Made them more friendly by putting the info under the image. That makes it much easier to display them next to each other on different screen sizes. 
Images like sneep.png and canteen.png will be put into public/images once #49 gets merged.
Closes #13


projects.json
```diff
@@ -3,9 +3,9 @@
     {
       "id": "78b6677c-28d2-4b2b-9825-4479706d2402",
       "name": "CO2-Labels for University Canteen",
-      "image": "/data_temp/images/canteen.png",
+      "image": "canteen.png",
```
EvanHahn: Should these use absolute URLs (like `/canteen.png`)?

src/components/layout.js
```diff
@@ -26,7 +26,7 @@ export default function Layout({ title, children }) {
         <Header />
         <Container maxWidth="lg" component="main">
           <Container maxWidth="sm">
-            <Typography variant="h1" className={classes.mainHeading}>
+            <Typography variant="h5" className={classes.mainHeading}>
```
EvanHahn: This doesn't seem related to this PR. Why did this change?

src/components/project/ProjectMetaData.js
```diff
@@ -0,0 +1,41 @@
...
+});
+
+export default function ProjectMetaData({ project }) {
+  const classes = useStyles();
+
+  return (
+    <Box>
+      <Box>
+        <span className={classes.cardIconBox}>
+          <img src={project.organisation_image} className={`${classes.orgLogo} `} />
```
EvanHahn: Nitpick: I think this can just be `className={classes.orgLogo}`.

src/components/project/ProjectMetaData.js
```diff
@@ -0,0 +1,41 @@
...
+    display: "inline-block"
+  },
+  cardIcon: {
+    verticalAlign: "bottom",
+    marginBottom: -2,
+    marginTop: 2
+  }
+});
+
+export default function ProjectMetaData({ project }) {
```
EvanHahn: I'm fine to have this in a separate component, but it seems like it's really only useful in the `ProjectPreview` component. Why a separate component/file?

src/components/project/ProjectPreview.js
```diff
@@ -1,29 +1,31 @@
...
+  return {
+    root: {
+      "&:hover": {
+        cursor: "pointer"
+      },
+      "-webkit-user-select": "none",
+      "-moz-user-select": "none",
+      "-ms-user-select": "none",
+      userSelect: "none",
+      backgroundColor: "#fafafa",
```
EvanHahn: Should this color come from a theme?

src/components/layout.js
positiveimpact: I changed it because I thought it looked oversized compared to the new card/font sizes. I added component="h1" to it. Would you like me to move it to a different PR?

src/components/project/ProjectMetaData.js
positiveimpact: I thought it makes it a bit clearer. If you think it makes more sense to not create new files for components, that are only used once, I will move it back to the main file.

src/components/project/ProjectPreview.js
positiveimpact: That was the color where I had trouble accessing it and you suggested it might be in CssBaseline. I changed it to backgroundColor: "inherit" which resolves to the same color.

positiveimpact:
Thanks for the reviews! I made changes and left comments about changes I didn't make (yet). 
Let me know if you think further changes should be made.

PR #83, opened 2020-02-06
Feature/add links to footer

positiveimpact:
I added the "impressum" link as well as social media links to the footer.
Additionally I made the footer always stay at the bottom of the page. Before it wasn't of the bottom if the height of the content of the page was smaller than the screen height. 
To accomplish this I had to add an additional div to layout.js and AboutLayout.js. Since every layout needs this div, I decided to create the file LayoutWrapper.js which is used by both layout files. This way we only have to make changes in one place if we want to make any changes to that div in the future.


public/data/projects.json
```diff
@@ -5,7 +5,7 @@
       "name": "CO2-Labels for University Canteen",
       "image": "/images/canteen.jpg",
       "organisation_name": "sneep Erlangen e.V.",
-      "organisation_image": "/images/sneep.png",
+      "organisation_image": "images/sneep.png",
```
EvanHahn: Should this be reverted?

public/icons/facebook.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
```
EvanHahn: Should we use the [Material UI icon](https://material-ui.com/components/material-icons/) for Facebook instead? The same comment applies for the other icons—Twitter, GitHub, Instagram, etc.

src/components/general/Footer.js
```diff
@@ -1,22 +1,90 @@
 import React from "react";
+import Link from "next/link";
 import { Box } from "@material-ui/core";
 import { makeStyles } from "@material-ui/core/styles";
 
 const useStyles = makeStyles(theme => ({
-  root: {
+  footer: {
```
EvanHahn: Why did this name change?

src/components/layouts/AboutLayout.js
```diff
@@ -2,32 +2,28 @@ import React from "react";
...
-          className={`${classes.main} ${classes.directlyUnderHeader}`}
-        >
-          {children}
-        </Container>
-        <Footer />
-      </ThemeProvider>
-    </>
+    <LayoutWrapper theme={aboutTheme}>
+      <Header />
+      <Container maxWidth={false} component="main" className={`${classes.main}`}>
```
EvanHahn: Nitpick: could this just be `className={classes.main}`?

src/components/layouts/LayoutWrapper.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import Head from "next/head";
+import { makeStyles, ThemeProvider } from "@material-ui/core/styles";
+
+const useStyles = makeStyles({
+  leaveSpaceForFooter: {
+    position: "relative",
+    paddingBottom: 68,
+    minHeight: "100vh"
```
EvanHahn: Nice!

src/components/layouts/LayoutWrapper.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import Head from "next/head";
+import { makeStyles, ThemeProvider } from "@material-ui/core/styles";
+
+const useStyles = makeStyles({
+  leaveSpaceForFooter: {
+    position: "relative",
+    paddingBottom: 68,
```
EvanHahn: Should this use `theme`?

public/data/projects.json
positiveimpact: Oh got I keep screwing this up for some reason. Reverted it.

public/icons/facebook.svg
positiveimpact: Makes sense. Switched to Material UI Icons

src/components/general/Footer.js
positiveimpact: I built a wrapper around it, removed it again and forgot to change it back. Reverted.

positiveimpact:
Thanks for the review! Implemented all suggested changes:
-added the "/" back in in organisation_image
-switched to material UI icons
-reverted className of root item to "root"
-Fixed the Nitpick
-used theme.spacing for height of footer and padding of body

I also removed some unused CSS that I found in Footer.js

Are there any other remarks? Otherwise I think this is ready to merge.

PR #85, opened 2020-02-06
Create env variable that enables us to deploy just the static pages from the dev branch

### LINKED ISSUE #73: Create env variable that enables us to deploy just the static pages from the dev branch
Right now we're deploying pre-launch-website branch to climateconnect.earth. On that branch all pages besides a few necessary static pages are commented out. 
The problem with this setup is, that we need to always  manually comment out new pages, if we merge in develop and this might lead to manual errors. Therefore we should create an env variable PRE_LAUNCH. 

In PRE_LAUNCH mode, if you access the main page should serve the /about page. Also none of the links in the header besides the logo should be shown. The only other pages, that should be accessible are /impressum and /faq. The footer can be shown as usual.
This depends on #49

ddhanesha:
resolves #73 

I have added a `PRE_LAUNCH` environment variable to `.env` file. This will help us deploying static pages from the dev branch without commenting out these pages. I have installed `dotenv` library to use env variables fetched from `.env` file. 

If the PRE_LAUNCH === 'true' it will show the about page into as a main page. 
Screen Shot 2020-02-06 at 8 30 03 AM

if PRE_LAUNCH === 'false' it will show the default project page we had before. 
Screen Shot 2020-02-06 at 8 30 37 AM

Please feel free to make any suggestions you have for this PR.

### DIFF
--- a/src/components/general/Header.js
+++ b/src/components/general/Header.js
@@ -66,20 +66,18 @@ export default function Header() {
 
   return (
     <Box component="header" className={classes.root}>
-      {process.env.PRE_LAUNCH === "true" ? (
-        <Container className={classes.container}>
-          <img src="/images/logo.png" alt="Climate Connect" className={classes.logo} />
-        </Container>
-      ) : (
-        <Container className={classes.container}>
-          <Link href="/">
-            <a>
-              <img src="/images/logo.png" alt="Climate Connect" className={classes.logo} />
-            </a>
-          </Link>
-          {isNarrowScreen ? <NarrowScreenLinks /> : <NormalScreenLinks />}
-        </Container>
-      )}
+      <Container className={classes.container}>
+        <Link href="/">
+          <a>
+            <img src="/images/logo.png" alt="Climate Connect" className={classes.logo} />
+          </a>
+        </Link>
+        {process.env.PRE_LAUNCH === "true" ? (
+          <></>
+        ) : (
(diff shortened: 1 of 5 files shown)


src/components/general/Header.js
```diff
@@ -66,15 +66,20 @@ export default function Header() {
...
+          <img src="/images/logo.png" alt="Climate Connect" className={classes.logo} />
+        </Container>
+      ) : (
+        <Container className={classes.container}>
+          <Link href="/">
+            <a>
+              <img src="/images/logo.png" alt="Climate Connect" className={classes.logo} />
+            </a>
+          </Link>
+          {isNarrowScreen ? <NarrowScreenLinks /> : <NormalScreenLinks />}
```
positiveimpact: I think here it would be easier to understand if you just wrapped an if else ( ? : ) function around `{isNarrowScreen ? <NarrowScreenLinks /> : <NormalScreenLinks />}`.
Something like this: 
`process.env.PRE_LAUNCH === true ? </> : (isNarrowScreen ? <NarrowScreenLinks /> : <NormalScreenLinks />)`
I think this makes sense because the image should still be a link going to "/" in the PRE_LAUNCH version and therefore this is the only difference between PRE_LAUNCH=true and PRE_LAUNCH=false

ddhanesha: Makes sense, updated code as requested.

positiveimpact:
Looks good. Merged!

PR #86, opened 2020-02-11
Sign up personal info

positiveimpact:
Created the first 2 pages of the sign up workflow.
To achieve that, I created a generic Form component which allows us to easily make forms in the same style.


pages/addinfo.js
```diff
@@ -0,0 +1,65 @@
...
+    {
+      required: true,
+      label: "Country",
+      type: "text",
+      select: {
+        values: countries,
+        defaultValue: ""
+      }
+    },
+    /*{
```
EvanHahn: Should we remove this?

src/components/general/SelectField.js
```diff
@@ -0,0 +1,50 @@
...
+    marginTop: theme.spacing(2)
+  },
+  maxHeight: {
+    maxHeight: 100
+  }
+}));
+
+export default function SelectField({ field }) {
+  const classes = useStyles();
+  console.log(field);
```
EvanHahn: Should this be here?

PR #98, opened 2020-02-18
Feature/organization page

### LINKED ISSUE #74: Create organisation page
Create a page with an overview over an organisation. 
This should include a picture, a description, a list of the projects of that organisation as well as a list of their members.

positiveimpact:
* Created page for organization accounts
* Created ProfilePreview component, which can be shown whereever we want to show an Avatar with a name and adapted the code to use it whereever suitable.
* Adapted profile page to be as similar as possible to organization page
* Created list of possible organization types - those will become important when adding the ability to create/edit a project

Closes #74


src/components/profile/ProfilePreviews.js
```diff
@@ -0,0 +1,27 @@
...
+    listStyleType: "none"
+  }
+});
+
+export default function ProfilePreviews({ profiles }) {
+  const classes = useStyles();
+
+  // TODO: use `profile.id` instead of index when using real profiles
+  return (
+    <Grid container component="ul" className={`${classes.reset} ${classes.root}`} spacing={2}>
```
EvanHahn: There's no `classes.root` here—should we remove it?

public/data/projects.json
```diff
@@ -4,6 +4,7 @@
       "id": "78b6677c-28d2-4b2b-9825-4479706d2402",
       "name": "CO2-Labels for University Canteen",
       "image": "/images/canteen.jpg",
+      "organization_url": "sneeperlangen",
```
EvanHahn: Should this start with a `/`? (Not sure.)

Issue #123, opened 2020-03-18
Create API route for logging in


positiveimpact:
Called by:
- The login interface

Functionality:
- If the correct information got entered, send jwt to user

Restrictions:
- logged out users only

ddhanesha:
Assigning this to myself as discussed!

PR #133, opened 2020-03-23
Installed and added django-rest-knox to backend project for token auth.

### LINKED ISSUE #110: Set up API token authentication for users
Is related to #123

ddhanesha:
Resolves #110 

I have installed django-rest-knox libarry for token authentication. I have updated settings file so that it includes knox in the installed apps, use token auth for all rest frameworks API. 

When you merge the request to develop or pull the branch locally you will have to run migration to create database tables. Knox creates database tables to store token securely. 

This will enable us to work on #123 ticket.

### DIFF
--- a/backend/climateconnect_main/settings.py
+++ b/backend/climateconnect_main/settings.py
@@ -43,7 +43,8 @@ LIBRARY_APPS = [
     'django.contrib.sessions',
     'django.contrib.messages',
     'django.contrib.staticfiles',
-    'rest_framework'
+    'rest_framework',
+    'knox'
 ]
 
 INSTALLED_APPS = CUSTOM_APPS + LIBRARY_APPS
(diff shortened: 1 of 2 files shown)


backend/climateconnect_main/settings.py
```diff
@@ -131,3 +132,10 @@
 # https://docs.djangoproject.com/en/2.2/howto/static-files/
 
 STATIC_URL = '/static/'
+
+REST_FRAMEWORK = {
+    'DEFAULT_AUTHENTICATION_CLASSES': [
+        'knox.auth.TokenAuthentication',
+        'rest_framework.authentication.SessionAuthentication'
```
positiveimpact: Does this mean, that session authentication is the fallback?

ddhanesha: Yes a fallback. Also easy to test APIs locally without frontend integration.

PR #148, opened 2020-04-04
Added a availability and skill tables.

### LINKED ISSUE #4: Create database table to store users
Sketches for layout(s):
- [profiles.json](https://github.com/climateconnect/climateconnect/blob/develop/public/data/profiles.json)
- [profile_types.json](https://github.com/climateconnect/climateconnect/blob/develop/public/data/profile_types.json)
- [profile_info_metadata.json](https://github.com/climateconnect/climateconnect/blob/develop/public/data/profile_info_metadata.json)

ddhanesha:
Resolves Partially #4 

This PR is to add generic tables we will be using for each User.


backend/climateconnect_api/models/common.py
```diff
@@ -0,0 +1,65 @@
...
+        return "%s" % self.name
+
+
+class Skill(models.Model):
+    name = models.CharField(
+        help_text="Name of the skill",
+        verbose_name="Name",
+        max_length=512
+    )
+
```
positiveimpact, line 45: Maybe it makes sense to additionally have a key or id field incase we want to change the name of a skill later?

ddhanesha: Djnago automatically creates a primary key in the postgres database.

ddhanesha: So no need to explicitly define it in model class

PR #149, opened 2020-04-04
API to get data for logged in user

### LINKED ISSUE #141: Create API route for getting data of the logged in user
Called by: 
- any page if the user is logged in

Arguments:
- `token` of the logged in user
- optional filter for only getting certain properties

Functionality:
- returns the entry of the logged in user from the users table, if filter is set, it only returns the filtered properties

Restrictions:
- You can only get the data of your own logged in account

ddhanesha:
Resolved #141 

I have added an API that will fetch user's profile data. Currently I don't have any filters added. I can add filters once I get some idea of what type of filters do we want to include.

You can test the API by logging in with a user and then go to `/api/my_profile` adding token authentication. Please review #148 because this PR is based of off #148 PR.


backend/climateconnect_api/views/user_views.py
```diff
@@ -45,18 +52,38 @@ def post(self, request):
...
-            username=request.data['email'], password=request.data['password'],
+            username=request.data['email'],
             email=request.data['email'], first_name=request.data['first_name'],
             last_name=request.data['last_name'], is_active=True
         )
 
+        user.set_password(request.data['password'])
+        user.save()
+
+        url_slug = (user.first_name + user.last_name).lower()
```
positiveimpact, line 63: This can stay like this for now, but maybe add a TODO comment to add some numbers incase the user slug already exists

PR #150, opened 2020-04-07
Added organization tables.

### LINKED ISSUE #3: Create database table(s) to store organizations
Sketches for layout(s):
- [organization.json](https://github.com/climateconnect/climateconnect/blob/develop/public/data/organizations.json)
- [organization_info_metadata.json](https://github.com/climateconnect/climateconnect/blob/develop/public/data/organization_info_metadata.json)
- [organization_types.json](https://github.com/climateconnect/climateconnect/blob/develop/public/data/organization_types.json)

ddhanesha:
resolves #3 

I have added organization tables in the PR. I created a new django app inside our backend directory to keep organization related work in separate directory. Pull down latest changes locally and run `./manage.py migrate` to migrate new database tables to your local postgres instance. Please take a look and let me know if you have any questions.


backend/climateconnect_main/settings.py
```diff
@@ -34,6 +34,7 @@
 
 CUSTOM_APPS = [
     'climateconnect_api',
+    'organization'
```
positiveimpact, line 37: Why do you register organization as a custom app?

backend/organization/models/organization.py
```diff
@@ -0,0 +1,151 @@
...
+
+    key = models.CharField(
+        help_text="Points to key of the type.",
+        verbose_name="Key",
+        max_length=512,
+        null=True,
+        blank=True
+    )
+
+    # TODO (Dip): Confirm about additional info column.
```
positiveimpact, line 19: As suggested on slack, I would add an `additionalInfo` column here, which should point to none, one or many fields of the `organization` model.
This column specifies which additional information the organization has to specify on their page if they choose this type.

backend/organization/models/organization.py
```diff
@@ -0,0 +1,151 @@
...
+        blank=True
+    )
+
+    short_description = models.TextField(
+        help_text="Short description about an organization",
+        verbose_name="Short Description",
+        null=True,
+        blank=True
+    )
+
```
positiveimpact, line 144: Here the possible additionalInfo columns should be added. so far those are `school`(for organization type `studentorganization`), `city` (for organization type `citygovernment`), `state` (for organization type `stategovernment`), `country`, `organ` (Both for organization type `federalgovernment`).

We would have to make sure, that they are only shown if the organization actually still has that type. Alternatively they could be stored in an additional table.

backend/organization/models/tags.py
```diff
@@ -0,0 +1,70 @@
+from django.db import models
+
+from organization.models import Organization
+
+
+class OrganizationTags(models.Model):
```
positiveimpact, line 6: Isn't this the same as `OrganizationTypes`? I think `OrganizationTypes` didn't exist in Evan's model, so I thought you put it instead of `OrganizationTags`

backend/climateconnect_main/settings.py
ddhanesha: Kept organization as a separate app directory in django for better code management. When you do that you have to add a directory to CUSTOM_APPS to allow other directories to recognize database models, functions and APIs. Read more: https://docs.djangoproject.com/en/3.0/intro/tutorial01/#creating-the-polls-app

backend/organization/models/tags.py
ddhanesha: Yeah I realized that after the code push. Removing `OrganizationTypes`.

backend/organization/models/organization.py
ddhanesha: Yeah we can just show those fields if the organization is in that type!

ddhanesha: Added all the columns here.

ddhanesha:
@positiveimpact merging this PR. If you find anything missing or wrong. Please let me know and I will make changes with a new PR.

PR #182, opened 2020-05-03
Feature/user roles

### LINKED ISSUE #181: create database table to store roles, organization members and project memebers

ddhanesha:
Resolves #181 .

Please review and let me know if something needed to be changed or you have questions.


backend/organization/admin.py
```diff
@@ -28,3 +28,19 @@ class ProjectAdmin(admin.ModelAdmin):
...
 
 admin.site.register(Project, ProjectAdmin)
+
+
+class ProjectMemberAdmin(admin.ModelAdmin):
+    search_fields = ('user__name', 'user__id', 'project__name',)
+    list_filter = ('role__name',)
+
+
+admin.site.register(ProjectMember, ProjectMemberAdmin)
```
positiveimpact, line 38: Does this mean, that they can do something through the django admin interface/api directly?

ddhanesha: No it is just for us to manage these tables.

Issue #121, opened 2020-03-18
Create API route(s) to get organization(s)


positiveimpact:
Called by:
- organization page
- #122 

Functionality:
- get organization(s) by id
- get organizations by search/filter terms
- optionally also return users and projects connected to this organization

Restrictions:
- none

ddhanesha:
`optionally also return users and projects connected to this organization`

I think this should be a separate API Route.

ddhanesha:
For restriction, should we do the same as we did for members API? Show some information to anonymous user and complete information to logged in users?

PR #185, opened 2020-05-11
Added APIs to get organizations and projects.

### LINKED ISSUE #119: Create API route(s) to get project(s)
Called by:
- #14
- #18 
- the index.js page

Functionality:
- get project by id
- get project by search/filter terms
- get most relevant projects (for front page)

Restrictions:
- none

ddhanesha:
Resolved #119 and #121. 

I think project APIs are still work in progress because I have not added filters by organizations. But review APIs and let me know if you have any questions. If you want to integrate this right now I can create another issue to work on adding filters and other things.

### DIFF
--- a/backend/climateconnect_main/urls.py
+++ b/backend/climateconnect_main/urls.py
@@ -14,16 +14,13 @@ Including another URLconf
     2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
 """
 from django.contrib import admin
-from django.urls import path
+from django.urls import path, include
 from climateconnect_api.views import (
     status_views, user_views
 )
 from knox import views as knox_views
 from django.conf import settings
 from django.conf.urls.static import static
-
-
-
 
 urlpatterns = [
     path('admin/', admin.site.urls),
@@ -32,5 +29,8 @@ urlpatterns = [
     path('logout/', knox_views.LogoutView.as_view(), name='logout-api'),
     path('signup/', user_views.SignUpView.as_view(), name="signup-api"),
     path('api/my_profile/', user_views.PersonalProfileView.as_view(), name='user-profile-api'),
-    path('api/members/', user_views.MemberProfilesView.as_view(), name="member-profiles-api")
-]+ static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
+    path('api/members/', user_views.MemberProfilesView.as_view(), name="member-profiles-api"),
+    path('api/', include('organization.urls')),
+]
+
+urlpatterns + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
--- a/backend/organization/serializers/project.py
+++ b/backend/organization/serializers/project.py
@@ -1,7 +1,9 @@
 from rest_framework import serializers
 
-from organization.models import Project
+from organization.models import (Project, ProjectMember)
 from climateconnect_api.serializers.common import SkillSerializer
+from climateconnect_api.serializers.user import UserProfileStubSerializer
+from climateconnect_api.serializers.role import RoleSerializer
 
 
 class ProjectSerializer(serializers.ModelSerializer):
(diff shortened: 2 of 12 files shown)


backend/climateconnect_main/urls.py
```diff
@@ -14,23 +14,23 @@
...
     path('admin/', admin.site.urls),
     path('ping/', status_views.PingPongView.as_view(), name='ping-pong-api'),
     path('login/', user_views.LoginView.as_view(), name='login-api'),
     path('logout/', knox_views.LogoutView.as_view(), name='logout-api'),
     path('signup/', user_views.SignUpView.as_view(), name="signup-api"),
     path('api/my_profile/', user_views.PersonalProfileView.as_view(), name='user-profile-api'),
-    path('api/members/', user_views.MemberProfilesView.as_view(), name="member-profiles-api")
-]+ static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
+    path('api/members/', user_views.MemberProfilesView.as_view(), name="member-profiles-api"),
+    path('api/', include('organization.urls')),
```
positiveimpact, line 33: Does this mean if the organization's name was 'greenpeace' the url would be `/api/greenpeace`? I think it would make most sense for it to be `/api/organizations/greenpeace`.

backend/organization/serializers/project.py
```diff
@@ -0,0 +1,12 @@
+from rest_framework import serializers
+
+from organization.models import Project
+from climateconnect_api.serializers.common import SkillSerializer
+
+
+class ProjectSerializer(serializers.ModelSerializer):
+    skills = SkillSerializer(many=True)
+
```
positiveimpact, line 9: We'll also display organization members. This could be included here, but could also do done in a seperate route. If we want to do it in a seperate route, we should add an issue for that.

backend/organization/views/organization_views.py
```diff
@@ -0,0 +1,17 @@
...
+from rest_framework.generics import ListAPIView
+from rest_framework.filters import SearchFilter
+from rest_framework.pagination import PageNumberPagination
+
+from organization.serializers.organization import OrganizationSerializer
+from organization.models import Organization
+
+
+class OrganizationAPIView(ListAPIView):
+    permission_classes = [IsAuthenticated]
```
positiveimpact, line 11: Here I'd do the same thing as with profiles: Users who aren't logged in should be able to partially view organization pages in my opinion.

backend/organization/views/organization_views.py
```diff
@@ -0,0 +1,17 @@
...
+    filter_backends = [SearchFilter]
+    pagination_class = PageNumberPagination
+    # `^` is for startswith method. User can search organization by name.
+    search_fields = ['^name']
```
positiveimpact, line 15: I would use `url_slug` as the search field.

backend/organization/views/project_views.py
```diff
@@ -0,0 +1,25 @@
...
+from rest_framework.filters import SearchFilter
+
+from organization.models import Project
+from organization.serializers.project import ProjectSerializer
+
+
+class ListProjectsView(ListAPIView):
+    permission_classes = [IsAuthenticated]
+    filter_backends = [SearchFilter]
+    search_fields = ['^name']
```
positiveimpact, line 13: I would use `url_slug` as the search field.

backend/organization/views/project_views.py
```diff
@@ -0,0 +1,25 @@
...
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.pagination import PageNumberPagination
+from rest_framework.filters import SearchFilter
+
+from organization.models import Project
+from organization.serializers.project import ProjectSerializer
+
+
+class ListProjectsView(ListAPIView):
+    permission_classes = [IsAuthenticated]
```
positiveimpact, line 11: We definitely need to show projects to users who aren't logged in. Here we could also do 2 different serializers for logged in and not logged in users.

backend/climateconnect_main/urls.py
ddhanesha: No this will includes urls.py file in organization directory. Its a way to include URLs in other files. Its common approach taken in Django for better API URL management.

backend/organization/serializers/project.py
ddhanesha: Organization members or Project members? We have two tables. I would prefer it to be a new API. It will make backend response faster I believe.

positiveimpact: Oh I meant to say project members. Sounds good!

ddhanesha: @positiveimpact Is organization members API will also be open for everyone (i.e.: permission=allowany)?

ddhanesha:
Made all requested changes. Added Project members API as well. Currently its permission class is set is_authenticated. but we can change it later if it should also be open.

PR #191, opened 2020-05-25
Create API routes for creating and editing projects

### LINKED ISSUE #11: Create API routes for creating and editing projects
Called by: 
- #52

Functionality: 
- Insert or edits a personal project or a project of an organization, in which you are an administrator, in the projects table

Restricted to:
- logged in users
- if they are creating a project of an organization: admins of that organization

ddhanesha:
Resolves #11 

I have created two APIs to create organizational projects and personal projects. I will add testing steps once its ready for review.


ddhanesha:
This PR is ready for review. @positiveimpact there are lot of changes here. So if you want we can setup a call to go over all the changes.

backend/organization/utility/project.py
```diff
@@ -0,0 +1,61 @@
...
+    if 'country' in data:
+        project.country = data['country']
+    if 'city' in data:
+        project.city = data['city']
+    if 'image' in data:
+        project.image = data['image']
+    if 'description' in data:
+        project.description = data['description']
+
+    project.url_slug = project.name + str(project.id)
```
positiveimpact, line 49: We could consider making easier to remember urls in the long run, so people can also tell somebody "Just go to climateconnect.earth/xyz to view our project". But in the short run I think this is fine :)

backend/organization/views/project_views.py
```diff
@@ -23,14 +37,102 @@ def get_serializer_class(self):
...
-    serializer_class = ProjectSerializer
-    pagination_class = None
+"""
+Note: I am splitting organization project creation and personal project creation into
+two separate APIs. It makes more sense do two APIs so that we can manage user accessibility
+by project or organization member roles.
+"""
+
+
+class OrganizationProjectsView(APIView):
```
positiveimpact, line 47: Shouldn't this be called something like `OrganizationsCreateProjectView`? I think the name doesn't really make it clear that this view is to create a project.

backend/organization/views/project_views.py
```diff
@@ -23,14 +37,102 @@ def get_serializer_class(self):
...
+        project_creation_pre_checks(request.data)
+        project = create_new_project(request.data)
+
+        if 'order' in request.data:
+            order = request.data['order']
+            ProjectParents.objects.create(
+                project=project, parent_organization=organization,
+                parent_user=request.user, order=order
+            )
+
```
positiveimpact, line 65: In the "share project" interfaces you can already add your team members. Should it be possible to pass team members to this API and have them be created by this view like the projectparents or should the frontend do an additional request for that? I would vote for the first option. The same thing applies to the personal project create API.

backend/organization/views/project_views.py
```diff
@@ -23,14 +37,102 @@ def get_serializer_class(self):
...
+        project = create_new_project(request.data)
+        if 'order' in request.data:
+            order = request.data['order']
+            ProjectParents.objects.create(
+                project=project, parent_user=request.user, order=order
+            )
+
+        # Here a user is creating an individual project so we will be adding them as a project member
+        role = Role.objects.get(name="Project Administrator")
+        ProjectMember.objects.create(
```
positiveimpact, line 84: Here we could directly add the other members, that the user specified in the "share project" interace.

backend/organization/views/project_views.py
```diff
@@ -23,14 +37,102 @@ def get_serializer_class(self):
...
+            project=project, user=request.user, role=role
+        )
+        serializer = ProjectSerializer(project)
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+
+class ProjectAPIView(APIView):
+    permission_classes = [OrganizationProjectCreationPermission]
```
positiveimpact, line 92: Does setting the permission to this mean that users who don't fulfil this permission also can't do the get request to view a project?

backend/organization/admin.py
```diff
@@ -43,4 +44,11 @@ class OrganizationMemberAdmin(admin.ModelAdmin):
     list_filter = ('role__name',)
 
 
-admin.site.register(OrganizationMember, OrganizationMemberAdmin)
\ No newline at end of file
+admin.site.register(OrganizationMember, OrganizationMemberAdmin)
+
+
+class ProjectParentsAdmin(admin.ModelAdmin):
+    search_fields = ('project__name', 'organization__name')
```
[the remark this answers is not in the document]
ddhanesha: Oh this is on django admin. Its for internal use only. For us and whoever else can access django admin.

backend/organization/utility/project.py
ddhanesha: Yeah agreed. For now I think this works.

backend/organization/views/project_views.py
ddhanesha: Good call. Yeah I will change the name

backend/organization/views/project_views.py
ddhanesha: Yeah I think with create API we can add project members in the same API. For editing each project we should create a separate API to add project members. Hope that makes sense.

backend/organization/views/project_views.py
ddhanesha: For create API will do that.

backend/organization/views/project_views.py
ddhanesha: Anyone who is authenticated can read a project. Only people with the permission will be able to edit project.

ddhanesha:
@positiveimpact this PR is ready for a second review. If everything looks good we can merge the PR.

positiveimpact:
Looks good to me! Merging.

PR #219, opened 2020-06-18
Feature/connect project and project list page to backend

### LINKED ISSUE #156: Connect "project" page with backend
Functionality:
- Use actual project data rather than mock data

Calls routes from:
- #119

### LINKED ISSUE #157: Connect frontpage with backend
Functionality:
- Use actual project data rather than mock data
- Make requests according to the filter, that was applied

Calls routes from:
- #119
- #120
- #121

positiveimpact:
This is now ready for review. 
## Summary: 
This PR makes the project page and the `projects` tab of the index page catch their data from our API rather than static local json files.

### Backend changes
- renames the `profiles_image` field of the `users` table to `image` for consistent naming between `users` and `organizations` table
- removes redundant `organization_image` field from `organizations` table - a `image` field had already been added, but the `organization_image`field had not been removed
- removes unused logging module from `settings.py`
- Renames `Posts` model to `Post` for consistency
- adds `role_in_project` and `time_per_week` columns to `ProjectMembers` model
- makes the `url_slug` field in the `Organizations` and `Projects` model a required field - each organization/project needs a url slug or their page cannot be viewed
- Allows a project to be created without specifying skills that they are looking for - especially for projects who don't want to collaborate it doesn't make sense to make the `skills` field required
- Adds `ordering` to projects to make pagingation work
- Allows `ProjectParents` to not have a `parent_organization` - this makes sense for personal projects
- removes the `order` field from `ProjectParents` as there was no use for it
- allows organizations to only have one `organization_tag` - same for projects and `project_tag`
- doesn't force `OrganizationTag` to submit `additional_info` anymore
- Creates a `pagination.py` file with custom pagination classes for projects and members
- Creates Serializers for `Posts`, `PostComments` and `ProjectComments`, `ProjectTags`, `ProjectTaggings`, `ProjectParents`
- Creates stub serializers for organizations to show mini previews
- Gets `Tags`, `Posts`, `Comments` and `ProjectParents` when getting a project
- Uses `url_slug` instead of `id` when making requests - users who just have the url won't know the id

### Frontend changes
- Actually catch project data from API in `index.js`
- Keep track of the page ourselves instead of letting `react-infinite-scroller` handle it
- Introduce `isLoading` variable to make sure `react-infinite-scroller` doesn't cause 2 pages to be caught at once on scroll
- Actually catch project and projectMember data from API on `[projectId].js`
- Set default step back to 0 on `share.js`. I forgot to reset it after doing debugging on page 3.
- Add a `LoginNudge` component to remind users that they are missing out on content by not being logged in
- Adapt various components to layout of API response
- Add some ToolTips with explanations to the project page
- Fix duplicate key issue in `ProjectPreviews.js`
- remove `project_status_metadata` - catch it from the API instead

Closes #156, Closes #157


backend/organization/models/members.py
```diff
@@ -19,10 +19,53 @@ class ProjectMember(models.Model):
...
+        (OVER_FOURTY, "More than 40")
+    )
+
+    time_per_week = models.CharField(
+        help_text="Shows how many hours per week the user is putting into this specific project.",
+        verbose_name="Time per week",
+        max_length=64,
+        choices=TIME_PER_WEEK_OPTIONS,
+        default=NOT_SPECIFIED
+    )
```
ddhanesha, line 67: I think you should use availability table here. Its already created.

backend/organization/models/organization.py
```diff
@@ -24,8 +24,6 @@ class Organization(models.Model):
         help_text="Points to organization url slug",
         verbose_name="URL Slug",
         max_length=1024,
-        null=True,
-        blank=True,
         unique=True
```
ddhanesha, line 27: So the problem here removing null=True and blank=True is. You are required to insert text to this column at the time of row creation. Currently we insert name + ID as a url_slug. So I think we should keep this as it was before. Otherwise there will conflict when you create a row.

backend/organization/models/project.py
```diff
@@ -20,8 +20,6 @@ class Project(models.Model):
         help_text="URL slug for project",
         verbose_name="URL slug",
         unique=True,
-        null=True,
-        blank=True,
         max_length=1024
```
ddhanesha, line 23: Same goes for project url slug too. We should keep null=True, blank=True.

backend/organization/serializers/content.py
```diff
@@ -0,0 +1,73 @@
...
+    serializer = UserProfileStubSerializer(obj.author_user.user_profile)
+    return serializer.data
+
+  def get_replies(self, obj):
+    top_level_comments = obj.post_comment.exclude(parent_comment_id__isnull=False)
+    serializer = PostCommentSerializer(top_level_comments, many=True)
+    return serializer.data
+
+class PostCommentSerializer(serializers.ModelSerializer):
+  #TODO: add pagination
```
ddhanesha, line 33: Just a note: Pagination can be added at API level

backend/organization/serializers/content.py
```diff
@@ -0,0 +1,73 @@
...
+  #TODO: add additional functionality such as 'is_abusive', 'deleted_at', 'updated_at'
+  #TODO: get likes
+  author_user = serializers.SerializerMethodField()
+
```
ddhanesha, line 37: You can do this: auther_user = UserProfileStubSerializer()

backend/organization/serializers/content.py
```diff
@@ -0,0 +1,73 @@
...
+  
+  def get_author_user(self, obj):
+    serializer = UserProfileStubSerializer(obj.author_user.user_profile)
+    return serializer.data
+
+class ProjectCommentSerializer(serializers.ModelSerializer):
+  #TODO: add pagination
+  #TODO: add additional functionality such as 'is_abusive', 'deleted_at', 'updated_at'
+  #TODO: get likes
+  author_user = serializers.SerializerMethodField()
```
ddhanesha, line 55: same here for author_user = UserProfileStubSerializer()

backend/organization/serializers/project.py
```diff
@@ -1,40 +1,117 @@
...
         serializer = SkillSerializer(obj.skills, many=True)
         return serializer.data
+    
+    def get_project_parents(self, obj):
+        serializer = ProjectParentsSerializer(obj.project_parent, many=True)
+        return serializer.data
+    
+    def get_project_posts(self, obj):
+        serializer = PostSerializer(obj.post_project, many=True)
+        return serializer.data
```
ddhanesha, line 44: I would get project posts and comments in  separate API, hence I would not add comments and posts as a serializer fields here. It will get pretty big pretty quickly and handling this big a of response can impact rendering and server performance.

backend/organization/views/project_views.py
```diff
@@ -208,15 +209,14 @@ def delete(self, request, project_id, member_id):
...
-    lookup_field = 'pk'
+    lookup_field = 'url_slug'
     serializer_class = ProjectMemberSerializer
-    pagination_class = PageNumberPagination
+    pagination_class = MembersPagination
     permission_classes = [IsAuthenticated]
 
     def get_queryset(self):
-        project = Project.objects.get(id=int(self.kwargs['pk']))
+        project = Project.objects.get(url_slug=self.kwargs['pk'])
```
ddhanesha, line 219: Change self.kwargs from `pk` to `url_slug`

backend/organization/views/project_views.py
```diff
@@ -99,15 +98,17 @@ class ProjectAPIView(APIView):
 
     def get(self, request, pk, format=None):
```
ddhanesha, line 99: Change parameter from `pk` to `url_slug`

backend/organization/views/project_views.py
```diff
@@ -99,15 +98,17 @@ class ProjectAPIView(APIView):
...
         try:
-            project = Project.objects.get(id=int(pk))
+            project = Project.objects.get(url_slug=pk)
```
ddhanesha, line 101: `url_slug = str(url_slug)`

backend/organization/views/project_views.py
ddhanesha: Also you will need to change the API URL from 
```
    path('projects/<int:pk>/', ProjectAPIView.as_view(), name='project-api-view'),
    path('projects/<int:pk>/members/', ListProjectMembersView.as_view(), name='project-members-api'),
```
to this 
```
path('projects/<slug:url_slug>/', ProjectAPIView.as_view(), name='project-api-view'),
path('projects/<slug:url_slug>/members/', ListProjectMembersView.as_view(), name='project-members-api'),
````

ddhanesha:
One other thing: Make sure you merger latest develop branch before and check if there are any migration conflicts.

backend/organization/models/organization.py
positiveimpact: makes sense!

backend/organization/models/members.py
positiveimpact: You're right. I changed it.

positiveimpact:
Addressed all comments.

PR #313, opened 2020-08-04
added team and comment count to tab labels

susanchesnut:


positiveimpact:
Looks good to me! One change request: 
The file `backend/organization/pagination.py` handles how many comments/members are returned on request - so if there is more than 12 members or more than 10 comments, on initial request the frontend would only know that there are at least that number of members/comments. Therefore the tabs should say "Members (12+)" and "Comments (10+)" if there is 12 members or 10 comments.

positiveimpact:
LGTM!

PR #315, opened 2020-08-08
added description to sign up page, matching design of share project

susanchesnut:
Added helper text: "Here you can create your personal account. You will have an opportunity to create/add an organization once signed up." Placed it on the first step at the same place as the explanatory text on https://climateconnect.earth/share


positiveimpact:
LGTM!

PR #272, opened 2020-07-26
Implement API route(s) for private messaging between users

ddhanesha:
Resolve #17

PR #354, opened 2020-09-26
Private messaging

positiveimpact:
This is building on #272. I wasn't able to push directly to that PR.


positiveimpact:
merged into develop and made PR from develop

Issue #196, opened 2020-05-29
Make footer show when you scroll up on InfiniteScroll pages


positiveimpact:
On the main page new projects/organizations/members load as you scroll down. This results in the footer never being shown. Therefore it should automatically be shown when you scroll up and be hideen again when you scroll down again. Otherwise people will never get to the footer if they are on the main page.

PR #388, opened 2020-11-18
WIP: Make footer show when scrolling up (#196)

vliaskov:
Use a separate style with "fixed" position when an upwards scroll is triggered,
in order to make the footer visible.

Comments or other suggestions welcome. 

I have only done this in ```frontend/src/components/general/Footer.js``` currently, let me know if ```frontend/src/components/footer/LargeFooter.js``` would require the same.

I am new to react and django, but interested in contributing to this project.

### DIFF
--- a/frontend/src/components/general/Footer.js
+++ b/frontend/src/components/general/Footer.js
@@ -20,6 +20,11 @@ const useStyles = makeStyles(theme => ({
     position: "absolute",
     bottom: 0
   },
+  relativePosition: {
+    position: "fixed",
+    bottom: 0,
+    backgroundColor: '#FFFFFF',
+  },
   spacingTop: {
     marginTop: theme.spacing(2)
   },
@@ -66,19 +71,20 @@ const useStyles = makeStyles(theme => ({
 }));
 
 //TODO: make footer stay on bottom on normal layout again
-export default function Footer({ className, noSpacingTop, noAbsolutePosition, large }) {
+export default function Footer({ className, noSpacingTop, noAbsolutePosition, showOnScrollUp, large }) {
   if (!large)
     return (
       <SmallFooter
         className={className}
         noSpacingTop={noSpacingTop}
         noAbsolutePosition={noAbsolutePosition}
+        showOnScrollUp={showOnScrollUp}
       />
     );
   else return <LargeFooter className={className} />;
 }
 
-const SmallFooter = ({ className, noSpacingTop, noAbsolutePosition }) => {
+const SmallFooter = ({ className, noSpacingTop, noAbsolutePosition, showOnScrollUp }) => {
   const classes = useStyles();
   const isNarrowScreen = useMediaQuery(theme => theme.breakpoints.down("xs"));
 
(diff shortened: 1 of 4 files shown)


frontend/src/components/general/Footer.js
```diff
@@ -82,11 +88,17 @@ const SmallFooter = ({ className, noSpacingTop, noAbsolutePosition }) => {
   const classes = useStyles();
   const isNarrowScreen = useMediaQuery(theme => theme.breakpoints.down("xs"));
 
+  const trigger = useScrollTrigger({
```
positiveimpact, line 91: This seems to be initialized as `false` which causes the Footer to show by default if you haven't scrolled yet.
However, the footer shouldn't be shown if you're scrolled all the way to the top, because otherwise, you might not be able to see a lot of projects/organizations/members without scrolling down when you're on the `/browse` page.

frontend/src/components/general/Footer.js
```diff
@@ -82,11 +88,17 @@ const SmallFooter = ({ className, noSpacingTop, noAbsolutePosition }) => {
...
+    threshold: 100
+  });
+
   return (
     <Box
       component="footer"
       className={`${className} ${classes.root} ${!noSpacingTop &&
-        classes.spacingTop} ${!noAbsolutePosition && classes.absolutePosition}`}
+        classes.spacingTop} ${!noAbsolutePosition &&
+        (!trigger ? classes.relativePosition : classes.absolutePosition) }`}
```
positiveimpact, line 101: This should only be used on "Infinite Scroll" pages. Specifically on the `/browse` page.
A good solution would be to add a variable `showOnScrollUp` which can be passed by the browse page to `WideLayout` and then to the `Footer` component.

positiveimpact (changes requested):
Great work, thanks for submitting this PR! Using material UI's `useScrollTrigger` is a very elegant way to do this!
Some details in the documentation of the issue were been unclear, specifically that the footer shouldn't be shown by default when you haven't scrolled yet. 
To address your question: This is only required for `frontend/src/components/general/Footer.js`, not for `frontend/src/components/footer/LargeFooter.js`.
Would be great if you address my comments and submit this for review again!

vliaskov:
Thanks for the review! I have added the suggested ```showOnScrollUp``` property to ```WideLayout```.
in 2 different implementations:

- The [first](https://github.com/vliaskov/climateconnect/commit/330cb3f538397f9f226ba5c584ea3504c4e2818e) uses ```useScrollTrigger``` and a custom hook ```TopOfPage``` to detect being at the start of the page.
  This hook is probably an overkill, I assume there must be a simpler way to get ```window.PageYoffset``` :).

- The [second](https://github.com/vliaskov/climateconnect/commit/a4189f3ffbfe94ae256fe112e481939e26a5bd04) uses a custom hook returning the scroll direction (```up```, ```down```, or ```none``` if at top of
   the page). This does not use ```useScrollTrigger```.

In both cases the functions are in their own files and exported, though I don't know if they would be useful for other features.

Let me know if one is more desirable than the other, and I can resubmit.

positiveimpact:
Thanks for resubmitting and sorry for the late review! 
I like both solutions, but slightly prefer the first solution, because it seems a bit easier to understand without having a look at the function. 

One problem that I found in there: 
In `browse.js` you call the TopOfPage hook like this: 

`const atTopOfPage = topOfPage(true);`

That doesn't work for two reasons: 
- `TopOfPage` is imported with a capital `T` and you're calling with a lowercase `t`
- You are passing `true`, but you have to specify which parameter you're passing: the `initTopOfPage` parameter 

Therefore the line should be changed to:
`const atTopOfPage = TopOfPage({initTopOfPage: true});`

Other than that it looks good to me and ready to resubmit!

vliaskov:
Thanks for the review, I fixed those 2 obvious issues in tmy PR branch :) Let me know if anything else is needed.

positiveimpact (approved):
LGTM!
Great job on the PR! In case you are interested in committing to this project regularly, I'd like to offer you to join on volunteer team and the corresponding Slack channel. If you're interested in that you can contact me via email at contact@climateconnect.earth.

PR #397, opened 2020-11-24
Loading spinner for browse pages, shows up while fetching further content

bulve:
Interesting is that the InfiniteScroll component takes the loader component as a prop, but it doesn't show it immediately or sometimes at all, so I am not using it. CircularProgress will show up when **isLoading** is set to true.

### DIFF
--- a/frontend/src/components/organization/OrganizationPreviews.js
+++ b/frontend/src/components/organization/OrganizationPreviews.js
@@ -3,6 +3,7 @@ import OrganizationPreview from "./OrganizationPreview";
 import { Grid } from "@material-ui/core";
 import { makeStyles } from "@material-ui/core/styles";
 import InfiniteScroll from "react-infinite-scroller";
+import CircularProgress from "@material-ui/core/CircularProgress";
 
 const useStyles = makeStyles({
   reset: {
@@ -11,8 +12,8 @@ const useStyles = makeStyles({
     listStyleType: "none",
     width: "100%"
   },
-  loader: {
-    clear: "both"
+  spinner: {
+    marginTop: '48px'
   }
 });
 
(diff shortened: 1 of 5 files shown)


frontend/src/components/organization/OrganizationPreviews.js
```diff
@@ -11,62 +12,57 @@ const useStyles = makeStyles({
...
+  spinner: {
+    marginTop: '48px'
   }
 });
 
 export default function OrganizationPreviews({
   organizations,
   loadFunc,
   hasMore,
-  showMembers,
```
positiveimpact, line 23: removing this property seems to make sense because it's not being used by any component at the moment. However if you remove it here, it would make sense to also remove it in `frontend/src/components/organization/OrganizationPreview.js` and `frontend/src/components/organization/OrganizationMetaData.js`.

positiveimpact (approved):
Great job on the PR! 
I made a comment about one minor thing that would be good to change. Once you change that, you can merge the PR without submitting it for review again. Good work!

frontend/src/components/organization/OrganizationPreviews.js
bulve: I think I missed the OrgnaziationPreview. I have pushed the update.
OrganizationMetaData didn't have this prop.

frontend/src/components/organization/OrganizationPreviews.js
```diff
@@ -11,62 +12,57 @@ const useStyles = makeStyles({
...
   showOrganizationType,
   parentHandlesGridItems
 }) {
-  const [isLoading, setIsLoading] = React.useState(false);
-  const classes = useStyles();
-  const [gridItems, setGridItems] = React.useState(
-    organizations.map((o, index) => (
+  const toOrganizationPreviews = (organizations) => organizations.map(o => (
```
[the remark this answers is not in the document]
bulve: Thanks, I can see even more duplication that could be removed, but maybe in other PR's

PR #399, opened 2020-11-26
Upgrade Pillow lib to support python 3.9

ddhanesha:


piperchester:
The major breaking change seems mostly harmless: https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst#800-2020-10-15

Are there areas we should test for regressions?

ddhanesha:
> The major breaking change seems mostly harmless: https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst#800-2020-10-15
> 
> Are there areas we should test for regressions?

Yeah I think we do. We use pillow image formatting so testing this would help. `climateconnect_main/climateconnect/backend/climateconnect_main/utility/general.py`

I just upgraded to 3.9.0 locally so I am testing it locally. I might add unittest if that's helpful too.

positiveimpact (approved):
The only function affected by this should be `get_image_from_data_url` in `/backend/climateconnect_main/utility/general.py`.
I tested this function locally and everything works as expected. Approved!

PR #401, opened 2020-12-01
Create test data for local development

ddhanesha:
This is a basic test data script that I have wrote to create test data locally for developers. 

All you need to do is run this command: `python manage.py create_test_data --number_of_rows 4`


backend/climateconnect_api/management/commands/create_test_data.py
```diff
@@ -0,0 +1,106 @@
...
+    print("Creating project roles...")
+
+    if not Role.objects.filter(name='Creator').exists():
+        Role.objects.create(
+            name='Creator', role_type=Role.ALL_TYPE
+        )
+        print("Creator role is successfully created.")
+    else:
+        print("Creator role already exists.")
+
```
positiveimpact, line 34: I would also add The administrator role in here:
`
if not Role.objects.filter(name='Administrator').exists():
        Role.objects.create(
            name='Administrator', role_type=Role.READ_WRITE_TYPE
        )
        print("Administrator role is successfully created.")
    else:
        print("Administrator role already exists.")
`

I think this makes sense because there is only 3 roles on the live server anyways and this isn't very likely to change in the near future.

backend/climateconnect_api/management/commands/create_test_data.py
```diff
@@ -0,0 +1,106 @@
...
+    for i in range(number_of_rows):
+        name = "Test project {}".format(i)
+        url_slug = name.replace(" ", "")
+        if not Project.objects.filter(name=name).exists():
+            Project.objects.create(
+                name=name, url_slug=url_slug,
+                status=ProjectStatus.objects.get(name="In Progress"),
+                start_date=timezone.now(),
+                short_description="This is a test project.",
+                country="Germany", collaborators_welcome=True
```
positiveimpact, line 86: I would add `city="Test {}".format(i)`. Otherwise, project cards will look a bit weird if the city is always empty.

positiveimpact (approved):
LGTM other than my comments and the fact that currently no users are being generated. 
If you address my comments and add users being generated, you can merge!

ddhanesha:
@positiveimpact added user test data and addressed your comments.

PR #404, opened 2020-12-03
Ordering of user profiles to show first profiles with images and late…

bulve:
Ordering of userProfiles, first by is_image_null in descending order and then by id in descending order too. 
The new annotation field is_image_null in queryset will have a value of 1 for profiles with images and 0 for profiles without. 

Removed ordering from Meta of UserProfile, because now ordering happens later in ListMemberProfilesView#get_queryset


backend/climateconnect_api/views/user_views.py
```diff
@@ -140,7 +141,10 @@ def get_serializer_class(self):
         return UserProfileStubSerializer
 
     def get_queryset(self):
-        user_profiles = UserProfile.objects.filter(is_profile_verified=True)
+        user_profiles = UserProfile.objects\
+            .filter(is_profile_verified=True)\
+            .annotate(is_image_null=Count("image"))\
```
positiveimpact, line 146: It seems like an empty image field defaults to `""`(empty string) rather than `null`.
In order for this value to be true when the user doesn't have an image, you'll have to use an additional filter:
`.annotate(is_image_null=Count('image', filter=Q(image="")))`.

positiveimpact (changes requested):
Great job on the PR. However as you'll see in my comment you'll have to change one thing because the image field acts a bit weird in Django. 

I recommend pulling Test Data with the function from @ddhanesha's newest PR so you can test these types of things more easily locally. For testing you can create testdata with said function, then navigate to `localhost:8000`to your django admin panel and add profile pictures to some of the test profiles via the user interface.

backend/climateconnect_api/views/user_views.py
bulve: Yeah, you are right, thank you. I have update PR.

positiveimpact (approved):
LGTM. Great job on the PR!

Issue #324, opened 2020-08-21
Loading feedback when you filter projects/organizations/members


positiveimpact:
# Current behavior
When you select filters in any of the 3 tabs on the browse page (projects, organizations, members) and then click the "apply" button there is nothing indicating that your results are being loaded.

# Expected future behavior
When you click the "apply" button and the request for new projects/organizations/members has been sent out, the content that is currently shown should be replaced with a circular progress material UI spinner with a message "Loading filtered [projects/organizations/members]" or similar below it. 
When the request returns, the following should happen:
- If the request throws an error, the old projects/organizations/members (before the filter was applied) should be shown again and an error message should be shown.
- If the request returns successfully the spinner should disappear and the filtered projects should be shown

positiveimpact:
Splitting this into multiple issues and unassigning this from myself for now. I'll be working on implementing various static pages and won't have time for this for now.

positiveimpact:
One use-case that I totally forgot:
This should also work for using full text search filter

positiveimpact:
This has been resolves by #456

PR #410, opened 2020-12-06
Add loading spinner and invoke when a user is searching or filtering options

piperchester:
## Description 

Implements the loading spinner feature from #324. Should render when the user is 
1. Searching for projects, orgs, members and
1. Applying new filters for projects, orgs, memebers

Initially I considered a `LoadingContext` for the spinner, but opted to not use it as we'd need to update state of the parent element when a user searches for something. This led me to looking into how we were managing state in `FilterSearchBar` -- I opted to hoist state management into the top-level `browse.js`, as opposed to passing `setState` and state through children components. Generally I think it's a best practice to keep `state` managed cohesively within a component, and to pass `props` through. Definitely would like the team's thoughts on this.

Edit: after #427 landed with the state management refactor, I revisited `LoadingContext` as a tradeoff towards performance and complexity management. Now the loading spinner uses that context to see if it needs to render, and we compose most of the browse container within a Provider.

## Future considerations for the team
Behaviorally I noticed that we search `onChange` for FAQ, but in the main search we search `onSubmit` -- I think we should normalize these from a UX perspective, and probably search for project types on change, but we should test the perf implications.

I generalized spinner into its own component; we could likely use that in other places but didn't refactor any more as I wanted the team's thoughts on this. 

Stylistically I tend to alphabetize props, imports, etc. where I can. I also create whitespace in between 1st and 3rd party imports and tend to favor a newline for conditionals. Any codebase style prefs like these I'm happy to change if undesired. 

Wanted to gauge the level of refactor and the diff size here -- is there a preference in keeping these more atomic? Since I'm new to the codebase I'm seeing some low hanging fruit and opportunity to clean things up so I'll generally tack it onto the PR if it doesn't balloon the diff significantly. But wanted to source discussion. 

## More changes
- Fixed `ReferenceError` in `possibleFilters` with optional chaining
- Remove unused prop references within `FilterSection`
- Includes some minor Prettier and lint fixes
- Minor readability and usability improvements to `Makefile`
- Drop extraneous `console.logs`


frontend/src/components/indexPage/FilterSection.js
```diff
@@ -49,29 +50,8 @@ const searchBarLabels = {
...
-export default function FilterSection({
-  filtersExpanded,
-  setFiltersExpanded,
-  typesByTabValue,
-  tabValue,
-  getProjects,
-  getOrganizations,
-  getMembers,
-  membersWithAdditionalInfo,
-  token,
```
piperchester, line 65: Many of these props weren't actually being used and just plucked out of the object. Removed them entirely and removed them from their call site within `browse`.

frontend/src/components/indexPage/FilterSection.js
```diff
@@ -82,49 +62,18 @@ export default function FilterSection({
...
-          items: { ...state.items, [type]: filteredItemsObject[type] },
-          hasMore: { ...state.hasMore, [type]: filteredItemsObject.hasMore },
-          urlEnding: { ...state.urlEnding, [type]: newUrlEnding },
-          nextPages: { ...state.nextPages, [type]: 2 },
-        });
-      } catch (e) {
-        console.log(e);
-      }
-    }
-  };
```
piperchester, line 114: Hoist all search submit handling to `browse`, and dedupe state management there. This should simplify how we're handling state management with search, and loading.

frontend/pages/browse.js
```diff
@@ -123,46 +127,75 @@ export default function Index({
...
     members: {},
     organizations: {},
   });
 
+  /**
+   * When applying new filters, update the state of the
+   * loading spinner until the async call finishes.
+   */
   const applyNewFilters = async (type, newFilters, closeFilters) => {
-    if (filters !== newFilters) {
```
piperchester, line 135: Instead of grouping everything within this conditional, I refactored it outside and added an early short circuit conditional if the objects are equal.

@ reviewers, wdyt? We could consider [isEqual](https://www.ecosia.org/search?q=lodash+is+equal&addon=chrome&addonversion=3.3.0&method=topbar) for object comparison if desired.

frontend/src/components/organization/OrganizationPreviews.js
```diff
@@ -67,7 +67,7 @@ export default function OrganizationPreviews({
       {parentHandlesGridItems
         ? organizations && organizations.length > 0
           ? toOrganizationPreviews(organizations)
-          : "No Results"
+          : "No organizations found."
```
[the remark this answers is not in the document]
positiveimpact: Makes a lot of sense. Maybe it would be even better to put "No organizations found. Try changing or removing your filter or search query.". However, I'm not sure if this message makes sense for all use-cases because I don't know if a filter/search without results is the only case in which this is shown.

frontend/src/components/project/ProjectPreviews.js
```diff
@@ -64,7 +68,7 @@ export default function ProjectPreviews({ projects, loadFunc, hasMore, parentHan
       {parentHandlesGridItems
         ? projects && projects.length > 0
           ? toProjectPreviews(projects)
-          : "No Results"
+          : "No projects found."
         : gridItems}
       {loadingSpinner()}
```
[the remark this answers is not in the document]
positiveimpact: I think generalizing this makes a lot of sense!

positiveimpact (commented):
Great PR! I added a few minor comments. I love the fact that you don't just make changes to flawed code but refactor it to make the code follow best practices! 
We should definitely always handle state in the highest-level component that has to access it. Most of the places where this is not the case are my fault and are (hopefully) from the very early stages of this project when I just tried to get everything done quick and dirty and had very limited knowledge of React.

I totally agree that it would make a lot of sense to search directly when typing in the search bar as this is much easier to understand and use for the user (generally the user-friendliness of our search and filters is horrible at the moment). Performance shouldn't be impacted too drastically if we don't called a new search query before an old one has returned. I was even thinking whether it makes sense to search projects, members and organizations at once when filtering and search as it might be confusing that all the filters change when you change a tab. But that would actually definitely have some performance implications.

I agree with generalizing spinner in its own component and it would be great if you refactor the places where it's being used.

As long as you clearly mark places where it might not be 100% clear whether you're just doing a refactor or also adding/removing/changing something I think it's great to refactor things as you go. In fact it would probably make sense to create a code style guide for the project. As long as it doesn't become too confusing and you ensure that the reactors don't break anything I'm totally fine with having them in the same PR!

Again great work and it's super helpful that you help make the codebase itself better on the side while implementing new features!

frontend/pages/browse.js
```diff
@@ -123,46 +127,71 @@ export default function Index({
...
-            items: { ...state.items, [type]: filteredItemsObject[type] },
-            hasMore: { ...state.hasMore, [type]: filteredItemsObject.hasMore },
-            urlEnding: { ...state.urlEnding, [type]: newUrlEnding },
-            nextPages: { ...state.nextPages, [type]: 2 },
-          });
-        } catch (e) {
-          console.log(e);
-        }
+    // Short circuit if filters are equal
+    if (filters == newFilters) {
```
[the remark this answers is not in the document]
piperchester: Yeah completely agreed 👍 leaving a suggestion:

```suggestion
    if (filters === newFilters) {
```

frontend/pages/faq.js
```diff
@@ -1,14 +1,15 @@
...
 import Cookies from "next-cookies";
 import axios from "axios";
+
+import FilteredFaqContent from "../src/components/faq/FilteredFaqContent";
+import FilterSearchBar from "../src/components/filter/FilterSearchBar";
+import HeaderImage from "../src/components/staticpages/HeaderImage";
 import tokenConfig from "../public/config/tokenConfig";
 import UnfilteredFaqContent from "../src/components/faq/UnfilteredFaqContent";
-import FilteredFaqContent from "../src/components/faq/FilteredFaqContent";
+import WideLayout from "../src/components/layouts/WideLayout";
```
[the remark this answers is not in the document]
piperchester: For sure. Will keep that in mind 🙏

piperchester:
> I totally agree that it would make a lot of sense to search directly when typing in the search bar as this is much easier to understand and use for the user (generally the user-friendliness of our search and filters is horrible at the moment). Performance shouldn't be impacted too drastically if we don't called a new search query before an old one has returned. I was even thinking whether it makes sense to search projects, members and organizations at once when filtering and search as it might be confusing that all the filters change when you change a tab. But that would actually definitely have some performance implications.

👍 captured the polish here: https://github.com/climateconnect/climateconnect/issues/422

> I agree with generalizing spinner in its own component and it would be great if you refactor the places where it's being used.

I'll take this on.

> As long as you clearly mark places where it might not be 100% clear whether you're just doing a refactor or also adding/removing/changing something I think it's great to refactor things as you go. In fact it would probably make sense to create a code style guide for the project. As long as it doesn't become too confusing and you ensure that the reactors don't break anything I'm totally fine with having them in the same PR!

🙏 will keep that in mind.

frontend/pages/browse.js
```diff
@@ -123,46 +127,71 @@ export default function Index({
...
+      return;
+    }
+
+    setFilters({ ...filters, [type]: newFilters });
+
+    const newUrlEnding = buildUrlEndingFromFilters(newFilters);
+    if (state.urlEnding[type] == newUrlEnding) {
```
positiveimpact, line 153: Additional suggestion:
```suggestion
    if (state.urlEnding[type] === newUrlEnding) {
```

frontend/src/components/communication/chat/ChatPreviews.js
```diff
@@ -111,7 +102,7 @@ export default function ChatPreviews({ chats, loadFunc, hasMore }) {
           chat={chat}
         />
       ))}
-      {loadingSpinner()}
+      <LoadingSpinner />
```
positiveimpact, line 105: We should pass the `isLoading` prop here I think
```suggestion
      <LoadingSpinner isLoading={isLoading}/>
```

frontend/src/components/organization/OrganizationPreviews.js
```diff
@@ -73,7 +63,7 @@ export default function OrganizationPreviews({
           ? toOrganizationPreviews(organizations)
           : "No organizations found."
         : gridItems}
-      {loadingSpinner()}
+      <LoadingSpinner />
```
positiveimpact, line 66:
```suggestion
      <LoadingSpinner isLoading={isLoading}/>
```

frontend/src/components/profile/ProfilePreviews.js
```diff
@@ -69,7 +59,7 @@ export default function ProfilePreviews({
           ? toProfilePreviews(profiles)
           : "No members found."
         : gridItems}
-      {loadingSpinner()}
+      <LoadingSpinner />
```
positiveimpact, line 62:
```suggestion
      <LoadingSpinner isLoading={isLoading}/>
```

frontend/src/components/project/ProjectPreviews.js
```diff
@@ -70,7 +58,7 @@ export default function ProjectPreviews({ projects, loadFunc, hasMore, parentHan
           ? toProjectPreviews(projects)
           : "No projects found."
         : gridItems}
-      {loadingSpinner()}
+      <LoadingSpinner />
```
positiveimpact, line 61:
```suggestion
      <LoadingSpinner isLoading={isLoading}/>
```

positiveimpact (commented):
I think you forgot to pass the `isLoading` prop to `<LoadingSpinner />` when using it unless I'm missing something and we don't actually need it. 

I was just about to approve, but I pulled the PR locally and tried it out and my filters always get cleared when I click apply. Will look into this further.

frontend/pages/browse.js
```diff
@@ -117,7 +165,7 @@ export default function Browse({
   return (
     <>
       <WideLayout
-        title="Climate Connect - global platform form climate change solutions"
+        title="Climate Connect - Global platform for climate change solutions"
```
piperchester, line 123: Fixed this title ☝️

frontend/pages/projects/[projectId].js
```diff
@@ -91,7 +91,7 @@ export default function ProjectPage({ project, members, posts, comments, token,
       message={message.message}
       messageType={message.messageType}
       title={project ? project.name : "Project not found"}
-      description={project.shortdescription}
+      description={project?.shortdescription}
```
Comment (piperchester) [inline frontend/pages/projects/[projectId].js:94]: @positiveimpact optional chaining to fix `ReferenceError`s is great, but WDYT about preferring throwing an error, or updating the UI when values are null / undefined instead?

frontend/src/components/browse/NoItemsFound.js
```diff
@@ -14,8 +14,8 @@ export default function NoItemsFound({ type }) {
   const classes = useStyles();
   return (
     <Typography component="h4" variant="h5" className={classes.infoMessage}>
-      Error trying to retrieve {type} data from the server. <br />
-      Please contact contat@climateconnect.earth if this happens repeatedly.
+      Error trying to retrieve <strong>{type}</strong> data from the server. <br />
+      Please contact contact@climateconnect.earth if this happens repeatedly.
```
piperchester, line 18: Fixed this ☝️

frontend/src/components/browse/BrowseContent.js
```diff
@@ -179,77 +203,101 @@ export default function BrowseContent({
...
+              unexpandFilters={unexpandFilters}
+              possibleFilters={possibleFilters(TYPES_BY_TAB_VALUE[0], filterChoices)}
+            />
+          )}
+
+          {/*
+            Render the spinner until results are returned.
+            Render the not found page if the object came back empty.
+          */}
+          {isLoading ? (
```
positiveimpact, line 223: We might need 2 different types of `isLoading` here: one for filters and one for just loading more projects. When applying a filter I would expect the spinner to **replace** the projects and when loading new projects I would expect the spinner to **appear under** the projects.

positiveimpact (changes requested):
There seems to be one bug with integrating the generic spinner component. Other than that I left a few minor comments, other than my comments it looks good to me now!

backend/organization/views/project_views.py
```diff
@@ -79,9 +81,13 @@ def get_queryset(self):
         if 'category' in self.request.query_params:
             project_category = self.request.query_params.get('category').split(',')
             project_tags = ProjectTags.objects.filter(name__in=project_category)
+            # Use .distinct to dedupe selected rows. We
+            # must use order_by in conjunction with distinct:
```
positiveimpact, line 85: When not passing a database column to distinct, we also don't need to use `order_by`

frontend/src/components/browse/NoItemsFound.js
```diff
@@ -10,11 +10,12 @@ const useStyles = makeStyles((theme) => {
   };
 });
 
-export default function NoItemsFound(/*{ type }*/) {
+export default function NoItemsFound({ type }) {
   const classes = useStyles();
   return (
     <Typography component="h4" variant="h5" className={classes.infoMessage}>
-      No results.
+      Error trying to retrieve <strong>{type}</strong> data from the server. <br />
```
positiveimpact, line 17: I changed this to "no results." because this message also appears if your filter simply returns no results. In that case it would be misleading to say that there has been an error retrieving data from the server.

frontend/src/components/general/LoadingSpinner.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Generalized loading spinner that's centered and to be used
+ * for search and filtering use cases. Uses a global loading context
+ * to determine if the spinnner should be rendered.
+ */
+const LoadingSpinner = ({ isLoading = false }) => {
+  const classes = useStyles();
+
+  // A short-circuit isLoading prop will bypass the loading context.
```
positiveimpact, line 23: I like this! :+1:

positiveimpact (approved):
It works great now, good job! I made 4 smaller comments. You can consider implementing the changes from these comments and then merge this PR without further review!

frontend/src/components/browse/NoItemsFound.js
piperchester: Makes sense 👍 will restore!

PR #421, opened 2020-12-12
Minor polish 

piperchester:
## Description 
This wasn't ticketed, but just some opinionated and misc. polish I wanted to float by the team. Happy to abandon all of this or tweak as desired.

https://www.loom.com/share/5a573d37e0ef466ab4c38c6deb23e3be

Notable:
- ~Alphabetize skills selection~
- Normalize casing and spelling of skills
- ~Drop divider under Tabs~
- ~Drop border bottom from header~
- Tweak gray color of border on footer
- Use a "close" icon to close the dialog over the keyboard back icon
- Fix the hover shift of project preview cards
- Add hover color to footer icons


positiveimpact:
Quick feedback on the suggested changes without reviewing the code in detail yet: 
> Alphabetize skills selection
Makes sense
> Normalize casing and spelling of skills
Makes sense
> Drop divider under Tabs
Would discuss in #design on Slack
> Drop border bottom from header
Would discuss in #design on Slack
> Tweak gray color of border on footer
Would discuss in #design on Slack
> Use a "close" icon to close the dialog over the keyboard back icon
On the one hand I like it because it's more clear, on the other hand most users would expect this button on the top right. I'm slightly in favor but would discuss in #design on Slack
> Add hover color to footer icons
Makes sense!

piperchester:
> On the one hand I like it because it's more clear, on the other hand most users would expect this button on the top right. I'm slightly in favor but would discuss in #design on Slack

Great call about it being upper right 👍

frontend/src/components/project/ProjectMetaData.js
```diff
@@ -99,15 +102,15 @@ const WithDescription = ({ className, project_parent, hovering, project, main_pr
...
+          <Collapse in={hovering} timeout="auto">
             <Typography className={classes.shortDescription}>
               {project.short_description}
             </Typography>
           </Collapse>
-          {!hovering && <Categories main_project_tag={main_project_tag} />}
         </Box>
       </Container>
-      {hovering && <Categories main_project_tag={main_project_tag} hovering={hovering} />}
+      <Categories main_project_tag={main_project_tag} hovering={true} />
```
positiveimpact, line 113: Why is hovering set to always true here? One problem this causes is that this always adds a background to the categories which looks a bit weird if the cards have different height (see screenshots)
Screenshot from 2020-12-21 18-39-49

positiveimpact (approved):
LGTM other than the one comment I left! One you change it so that the background isn't shown on categories when not hovering, you can merge to master without additional review!

frontend/src/components/project/ProjectMetaData.js
piperchester: Good call! Backed this out. The intent was to minimize a minor layout shift when hovering, like here:


https://user-images.githubusercontent.com/1530684/104094569-89eb0000-5246-11eb-8f0b-154b022d58cf.mov

PR #425, opened 2020-12-19
Add thin wrapper on top of react-timeago's dateFormatter to support desired dates

### LINKED ISSUE #305: Timeago currently shows "2 years" if the project is 1 year and one day old
Always round down on years in timeago. For example on the project page.

piperchester:
## Description 
This fixes #305. I opted to go a step further, and include the number of days in the formatting. Happy to discuss the end result and iterate on whatever's desired! For example,

Screen Shot 2021-01-10 at 3 59 53 PM

Screen Shot 2021-01-10 at 3 59 39 PM

Please let me know if I missed any edge cases in the date calculation logic. I think #437 will help with more test data generation for local envs. 

## Also
- Fixed linting
- Updated test data script to use start dates beyond a year


README.md
```diff
@@ -76,6 +76,17 @@ python manage.py test <file_path> or <file_path + class_name>
...
+
+We use Jest as our testing framework write to tests for the FE code. Write
+test files with `.test.js` and execute them directly with
+
+```sh
+yarn jest path/to/testfile.test.js
+```
+
+See npm scripts in `package.json`.
+
```
piperchester, line 89: Added this snippet to start onboarding developers to more tests.

backend/climateconnect_api/management/commands/create_test_data.py
```diff
@@ -135,21 +136,30 @@ def create_project_test_data(number_of_rows: int):
         url_slug = name.replace(" ", "")
 
         if not Project.objects.filter(name=name).exists():
+            # Buffer some start date to create more
+            # sense of older projects. This originated out of
+            # fixing a bug around the "timeago" utility, where
+            # timeago would round up to 2 years, when a project
+            # is only 1 year and 1 day old:
+            over_a_year_ago = timezone.now() - timezone.timedelta(days = 366)
```
piperchester, line 144: @ddhanesha I added this as the default to create more variance in our test data. WDYT?

frontend/.eslintrc.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
     ecmaFeatures: {
       jsx: true,
     },
-    ecmaVersion: 2018,
+    ecmaVersion: 2020,
```
piperchester, line 16: @positiveimpact I bumped this to support optional chaining. I think we should probably update our lint config in the future, but seems low priority.

frontend/package.json
```diff
@@ -7,11 +7,13 @@
     "deploy": "yarn build && yarn export",
     "dev": "next dev",
     "export": "next export",
+    "fix-lint": "eslint . --fix",
```
piperchester, line 10: My preference is to actually include this directly into lint, so that any time a contributor runs "lint" they automatically fix. I know that some people don't like that though so added this. Wdyt @positiveimpact?

frontend/pages/_app.js
```diff
@@ -44,7 +44,7 @@ export default function MyApp({
         anonymizeIp: true,
       },
     });
-    ReactGA.pageview(!!pathName ? pathName : "/");
+    ReactGA.pageview(pathName ? pathName : "/");
```
piperchester, line 47: Lint fix.

frontend/src/components/browse/NoItemsFound.js
```diff
@@ -10,7 +10,7 @@ const useStyles = makeStyles((theme) => {
   };
 });
 
-export default function NoItemsFound({ type }) {
+export default function NoItemsFound() {
```
piperchester, line 13: Remove unused for fixing lint.

frontend/src/utils/formatting.js
```diff
@@ -0,0 +1,40 @@
...
+  if (unit === "year") {
+    // The days calculation comes directly from react-timeago:
+    const days = Math.round(Math.abs(epochMiliseconds - Date.now()) / (1000 * 60 * 60 * 24));
+
+    // With the number of days, we can just append the remaining
+    // days after taking into account the year overflow
+    const dayAfterYears = days % (365 * value);
+    const pluralizeDays = (days) => {
+      return days !== 1 ? "days" : "day";
+    };
```
piperchester, line 26: This is the core fix.

piperchester:
@ reviewers: please review https://github.com/climateconnect/climateconnect/pull/438 first as it has some of these lint fixes.

backend/climateconnect_api/management/commands/create_test_data.py
positiveimpact: Sounds good! The more different we make our test data, the easier it will be to find missing edge cases while developing locally

frontend/package.json
positiveimpact: I have never tried this feature so I'm not sure how reliable it is. After all, we can still reverse a change it makes before committing. Still, I'd leave it as the `fix-lint` command for now and possibly move it to `lint` later.

frontend/src/utils/formatting.js
```diff
@@ -0,0 +1,40 @@
...
+ * the number of days is since the input date, and
+ * then append a suffix to pluralize number of days.
+ *
+ * See more on the formatting options, here:
+ * https://github.com/nmn/react-timeago#formatter-optional
+ */
+const yearAndDayFormatter = (value, unit, suffix, epochMiliseconds) => {
+  if (unit === "year") {
+    // The days calculation comes directly from react-timeago:
+    const days = Math.round(Math.abs(epochMiliseconds - Date.now()) / (1000 * 60 * 60 * 24));
```
positiveimpact, line 19: I'm a bit confused about the `epochMiliseconds - Date.now()` part - Shouldn't this result in `0`?
I just checked the react-timeago docs and they describe the fourth argument of the formatter like this:
`epochSeconds: The result of Date.now() or the result of a custom now prop.`

positiveimpact (approved):
LGTM overall. One very small general comment would be that I think it will now return `1 year and 0 days` rather than `1 year` if the time is exactly 1 year ago. 
Also, one line that I pointed out confused me a little bit but since it's taken directly from their react-timeago it must be correct.
Overall I like the idea of showing the date more exactly!
Approved! If the `1 year and 0 days` thing is correct please implement a change for it to say `exactly 1 year`or something like that. 
Afterwards you can merge without further review.

frontend/src/utils/formatting.js
piperchester: Yeah good question! It's non-obvious (I believe the docs for `react-timeago` are incorrect as `Date.now()` returns milliseconds) and took a bit of double checking to work through. I opened a PR into the project to fix it: https://github.com/nmn/react-timeago/pull/151

I'll handle the 0 case, and add a couple more comments outlining it.

piperchester:
Do you mind sanity checking this @positiveimpact locally? I think it's fine I'd love a final pair of eyes before landing it.

frontend/src/utils/formatting.js
```diff
@@ -13,19 +13,37 @@
...
-    const days = Math.round(Math.abs(epochMiliseconds - Date.now()) / (1000 * 60 * 60 * 24));
+    // https://github.com/nmn/react-timeago/blob/master/src/formatters/buildFormatter.js#L84-L86
+
+    // Number of ms in a second: 1000
+    // Number of seconds in a minute: 60
+    // Number of minutes in an hour: 60
+    // Number of hours in a day: 24
+    // Should be 86,400,000
+    const msInDays = 1000 * 60 * 60 * 24;
+    const elapsedDaysSinceEpoch = Math.round(elapsedMilliseconds / msInDays);
```
positiveimpact, line 28: `elapsedMilliseconds` isn't the diff between now and the date we're formatting. It's the total unix time elapsed until the date that we're formatting. I think you want the time elapsed between the date you're trying to format and now here.

positiveimpact (changes requested):
There is one logic error in the new implementation, see my comment

frontend/pages/projects/[projectId].js
```diff
@@ -9,7 +9,6 @@ import ProjectOverview from "../../src/components/project/ProjectOverview";
 import ProjectContent from "../../src/components/project/ProjectContent";
 import ProjectTeamContent from "../../src/components/project/ProjectTeamContent";
 import ProjectCommentsContent from "../../src/components/project/ProjectCommentsContent";
-import { getParams } from "./../../public/lib/generalOperations";
```
Comment (piperchester) [inline frontend/pages/projects/[projectId].js:12]: This was unused.

piperchester:
Oof, this was a mind bender. The internal implementation of timeago was really confusing me with its inconsistent naming. 

Definitely was more work than I thought, but I think everything's working now:
Screen Shot 2021-02-03 at 6 17 29 PM

README.md
```diff
@@ -107,6 +107,17 @@ python manage.py test <file_path> or <file_path + class_name>
...
```
ddhanesha, line 120: Could rebase to master. We have made some changes to the README.md file

backend/climateconnect_api/management/commands/create_test_data.py
```diff
@@ -135,21 +136,31 @@ def create_project_test_data(number_of_rows: int):
...
+            one_year_and_one_day_ago = timezone.now() - timezone.timedelta(days = 366)
+
             project = Project.objects.create(
                 name=name,
                 city="Test {}".format(i),
                 collaborators_welcome=True,
                 country="Germany",
                 short_description="This is a test project.",
-                start_date=timezone.now(),
+                start_date=one_year_and_one_day_ago,
```
ddhanesha, line 153: Hmmmm... Makes sense. Should we do random start_date maybe? Don't think its necessary just a thought.

frontend/src/utils/formatting.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * This is thin wrapper on top of the formatter that's passed to
+ * reat-timeago. The desired date is a format that's something like
```
ddhanesha, line 3: nit: react-timeago

frontend/src/utils/formatting.js
```diff
@@ -0,0 +1,71 @@
...
+) => {
+  // Only apply custom logic for the year case
+  if (unit === "year") {
+    // The days calculation comes directly from react-timeago:
+    // https://github.com/nmn/react-timeago/blob/master/src/formatters/buildFormatter.js#L84-L86
+
+    // Number of ms in a second: 1000
+    // Number of seconds in a minute: 60
+    // Number of minutes in an hour: 60
+    // Number of hours in a day: 24
```
ddhanesha, line 33: Nit: I like comments, but some code is self explanatory. This could be concise comment.

positiveimpact (approved):
LGTM now!

backend/climateconnect_api/management/commands/create_test_data.py
piperchester: I like that idea for generating more entropy and randomness! We could revisit in a future PR.

frontend/src/utils/formatting.js
piperchester: Great call out 🙏 agree this is overly verbose. I'll condense - thanks for highlighting!

frontend/src/utils/formatting.js
piperchester, line 3:
```suggestion
 * react-timeago. The desired date is a format that's something like
```

piperchester: Good catch @ddhanesha! Here's an example of the GitHub "suggestion" feature for spelling suggestions if you hadn't seen it before. It's super useful IMO.

piperchester:
Thanks Dip and Chris - made some comment clarifications and updated `README.md`.

PR #433, opened 2021-01-08
fix(*): undefined project with optional chaining

piperchester:
Slack context here: https://climateconnect.slack.com/archives/CPZPNBVKR/p1610065403001800

### DIFF
--- a/backend/climateconnect_api/management/commands/create_test_data.py
+++ b/backend/climateconnect_api/management/commands/create_test_data.py
@@ -1,3 +1,5 @@
+from organization.models.tags import ProjectTagging, ProjectTags
+from organization.models.project import ProjectParents
 from django.core.management.base import BaseCommand
 from django.utils import timezone
 from django.contrib.auth.models import User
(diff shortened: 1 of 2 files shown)


positiveimpact:
I don't believe this actually fixes it, because the error says `Cannot read property 'shortdescription' **of null**`. This means that `project` is actually `null`. 
I just debugged this and turns out the problem is that there is no `ProjectParents` entry for the auto-generated projects. 

`parseProject(project)` tries to get the creator like this:
`creator: project.project_parents[0].parent_organization
      ? project.project_parents[0].parent_organization
      : project.project_parents[0].parent_user`

This seems to result in the whole functions returning null. I think the best fix for this would be to simply also create a project parent within the script for generating projects. I will get on this and commit to this PR. We could leave the optional chaining in, but I don't think it makes a lot of sense to accept unexpected inputs.

positiveimpact:
So I just realized my comment here was a bit dumb, because optional chaining is exactly for this use-case and it does actually fix the error. However, your project will still be empty. So I'd still push a fix to the `create_test_data` script, but obviously this does provide an immediate fix.

positiveimpact (approved):

piperchester:
> So I just realized my comment here was a bit dumb, because optional chaining is exactly for this use-case and it does actually fix the error. However, your project will still be empty. So I'd still push a fix to the `create_test_data` script, but obviously this does provide an immediate fix.

Woot, 👏 thanks Chris!

PR #436, opened 2021-01-10
Optimize Page Titles

### LINKED ISSUE #435: Standardize and optimize page titles for SEO

Dristian:
## Description

Standardize page titles for better SEO results. 
- Add "| Climate Connect" to every page title as default.
- Capitalize all headlines of page titles (I hope it's correct)
- Change "project" of page titles to "climate solution" (e.g. Create your Climate Solution vs. Create a Project") for better results

Closes #435 

## Todo

- Go through all pages and check page titles for a default.

### DIFF
--- a/frontend/pages/manageProjectMembers/[projectUrl].js
+++ b/frontend/pages/manageProjectMembers/[projectUrl].js
@@ -61,7 +61,7 @@ export default function manageProjectMembers({
     );
   else {
     return (
-      <Layout title="Manage Solution Members" hideHeadline>
+      <Layout title="Manage Solution's Members" hideHeadline>
         <ManageProjectMembers
           user={user}
           members={members}
--- a/frontend/pages/manageProjectMembers/[projectUrl].js
+++ b/frontend/pages/manageProjectMembers/[projectUrl].js
@@ -35,13 +35,13 @@ export default function manageProjectMembers({
   );
   if (!user)
     return (
-      <WideLayout title="Please log in to manage the members of this project" hideHeadline={true}>
+      <WideLayout title="Please Log In to Manage the Members of this Climate Solution" hideHeadline={true}>
         <LoginNudge fullPage whatToDo="manage the members of this project" />
       </WideLayout>
     );
   else if (!members.find((m) => m.id === user.id))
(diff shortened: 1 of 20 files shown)


frontend/pages/manageProjectMembers/[projectUrl].js
```diff
@@ -53,15 +53,15 @@ export default function manageProjectMembers({
...
+      <WideLayout title="No Permission to Manage Members of this Climate Solution" hideHeadline={true}>
         <Typography variant="h4" color="primary" className={classes.headline}>
           You need to be an administrator of the project to manage project members.
         </Typography>
       </WideLayout>
     );
   else {
     return (
-      <Layout title="Manage project's members" hideHeadline>
+      <Layout title="Manage Solution Members" hideHeadline>
```
Comment (positiveimpact) [inline frontend/pages/manageProjectMembers/[projectUrl].js:64]: I think we still have to keep the 's here.
`Solution's Members`

frontend/pages/share.js
```diff
@@ -20,13 +20,13 @@ export default function Share({
   const { user } = useContext(UserContext);
   if (!user)
     return (
-      <WideLayout title="Please log in to share a project" hideHeadline={true}>
+      <WideLayout title="Please Log In to Share your Climate Solution" hideHeadline={true}>
```
positiveimpact, line 23: I think if we use capital letters for `Climate Solution` we should also use capital letters for `To` and `Your`. Otherwise it's inconsistent

positiveimpact (changes requested):
Looks good overall! I found 2 small things that should be changed. 
I think we need to be consistent with capitalization of letters in all places. I'd be slightly in favor of using lowercase letters but we could also consistently capitalize each word.

frontend/pages/share.js
Dristian: We have talked about it. Either we have to write all words in lower case or we have to follow the english headline capitalization rules. In this case nearly all words are written in upper case except articles, coordinating conjunctions and short prepositions. But I think there are more than one main rule and possible written styles :)

positiveimpact (approved):
LGTM! Approved.

PR #440, opened 2021-01-12
Fix extraneous network fetches in Previews components 

piperchester:
## Description

Re-introducing local state to guard against duplicate network calls. This mitigates a known limitation with [`react-infinite-scroller`](https://github.com/danbovey/react-infinite-scroller#props)'s `hasMore` prop. This should throttle the number of requests. In the future, we might consider debouncing using something like [`_.debounce`](https://lodash.com/docs/4.17.15#debounce).

Also added a TODO comment to fix the React warning that InfiniteScroll is throwing.

## Todo

### DIFF
--- a/frontend/src/components/communication/chat/ChatPreviews.js
+++ b/frontend/src/components/communication/chat/ChatPreviews.js
@@ -7,7 +7,7 @@ import {
   Divider,
   Badge,
   Typography,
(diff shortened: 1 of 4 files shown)


positiveimpact (approved):
LGTM! approved.

PR #443, opened 2021-01-13
Drop fallback theme; improve infinite scroll performance

piperchester:
## Description

Shout out to @positiveimpact for discovering this. We're seeing a significant perf increase locally from dropping this fallback. We can follow up with a different fix for the console warning.

## Todo


positiveimpact (approved):
LGTM!

PR #446, opened 2021-01-14
Implement a temporary location service solution

### LINKED ISSUE #283: Save geolocation instead of just the location string in database
We should use our local Pelias API from issue #125 to convert text input to location coordinates.

### LINKED ISSUE #284: Create location Autocomplete frontend component
# General
Create a component where users get suggestions for locations that exist in our location service while typing. Our `AutocompleteSearchbar` component can be used as a base for this. 
Also our custom `Form` component should be changed to support `LocationAutocomplete` Fields.
Calls API endpoint(s) from #330.

### LINKED ISSUE #285: Allow filtering by radius around geolocation in Filters
# How it currently works
Before having a location service like google maps or Pelias, location is saved as a `country` and a `city` string in our database. So far people could filter projects by putting the exact strings of country or city using `ListProjectsView`.
# How it should work
Once a location service is implemented, we should allow filtering by a radius around a location. 
## How to call the new API endpoint for filtering by location
The url parameters for filtering by location for this could look something like this: `https://api.climateconnect.earth/projects/?location=[place_id]&radius=[radius in km or miles]`
The place id is an id provided by the location service provider, e.G. Google maps.
## What the new API endpoint should do
The API should then make a request to our location service to get the latitude and longitude of the place. It should then make a database query to find all project which lie within the provided radius of the specified place. They should be sorted by nearest to furthest. The default value for radius should be 20km or the rough equivalent in miles.

### LINKED ISSUE #320: Implement Google maps as a temporary location service solution
# Problem description
In the long run we want to use a self-hosted version of the open-source alternative [Pelias](https://github.com/pelias/pelias) as described in #125 .
However this may take a bit to set up and Google Maps platform offers users 200$ of free credits per month. 
A location service is vital for issues #283, #284 and #285 which allow users to have their location autocompleted, actually save geocoordinates of a place and therefore allow for searching projects that are near to a certain location.
An API Key has already been set up and can be found at https://console.cloud.google.com/apis/credentials.
The endpoints of #283, #284 and #285 should be set up in a way that they can easily be migrated from Google Maps to Pelias.

### LINKED ISSUE #492: Implement location search legacy system
# General
When #446 goes live, we will use an open API for location search. In case this API experiences any downtime or we have any errors in our code, we should have a fallback method for saving and displaying locations. Otherwise the website will be mostly unusable if we experience problems with location search.

# The idea
Allow falling back to a system where users just specify "city" and "country" as strings when filling a location input anywhere on the website. Switching to the legacy system should be done by changing the envrionment variable `ENABLE_LEGACY_LOCATION_FORMAT`.

# Details
#### In each place in the code where you can input a location, the following changes will need to be made:
**Conversion:** Replace "location" input field with inputs for "city" and "country". Convert them to an object similar to what `LocationSearchBar` returns (just without geocoordinates and place ids) when passing them.

#### Inputs using the `frontend/src/components/account/EditAccountPage.js` component: 
- `frontend/pages/editprofile.js`
- `frontend/pages/editOrganization/[organizationUrl].js`
- `frontend/src/components/organization/EnterDetailledOrganizationInfo.js`
- [x] Do conversion in the `EditAccountPage` component

#### Inputs using the `frontend/src/components/general/Form.js` component:
- `frontend/src/components/signup/AddInfo.js`
- `frontend/src/components/shareProject/ShareProject.js`
- [x] Do conversion in the component itself

#### Inputs using `frontend/src/components/organization/EnterDetailledOrganizationInfo.js` independently (without Form component):
- `frontend/src/components/editProject/EditProjectOverview.js`
- [x] Do conversion in the component itself

positiveimpact:
## Description

This PR will do the following:
- create geospatial fields for all location in the database (in addition to the human-readable fields)
- Convert all fields where you can enter a location in the frontend to autocomplete fields using Open Streetmaps
- Save all locations as coordinates and as a string redundantly

## Todo

- [x] Convert all places where location is edited or displayed to LocationSearchBar's and save coordinates in db - Finished Jan 29
- [x] Allow searching for locations in new system - Deadline Jan 30
- [x] Add radius option when searching for locations Deadline Feb 1st
- [x] Make system backwards compatible in case we experience issues - allow setting env variable `ENABLE_LEGACY_LOCATION_SYSTEM` which falls back to just using city and country fields of the location table. New locations will still be saved in the location table but without the polygon/point field filled out. - Deadline Feb 11th
- [x] Create script for converting all current location data to the new format - Deadline Feb 12th
- [ ] Run script and merge PR (Will cause downtime) - Deadline Feb 13th
Closes #285 Closes #284 Closes #283  Closes #492

- [x]  PR has a meaningful title?


backend/climateconnect_api/views/user_views.py
```diff
@@ -233,14 +233,16 @@ def post(self, request):
...
+            user_profile.background_image = get_image_from_data_url(
+                request.data['background_image'], True, 1280
+            )[0]
+
+        if 'location' in request.data:
+            geo_location = get_geo_location(request.data['location'])
+            user_profile.location = geo_location['location']
+            user_profile.latitude = geo_location['latitude']
+            user_profile.longitude = geo_location['longitude']
+
```
ddhanesha, line 245: I think because of this you are getting wrong location? Don't we have to change the add a new column?

backend/location/migrations/0004_auto_20210121_1452.py
```diff
@@ -0,0 +1,29 @@
...
+            model_name='location',
+            name='main_polygon',
+            field=django.contrib.gis.db.models.fields.PolygonField(blank=True, geography=True, null=True, srid=4326),
+        ),
+        migrations.AlterField(
+            model_name='location',
+            name='state',
+            field=models.CharField(blank=True, help_text="Points to location's state name", max_length=1024, null=True, verbose_name='Name'),
+        ),
+    ]
```
ddhanesha, line 29: Since you are making changes to a same model and its not deployed to production, you can remove 4 migrations and then create new one so it more organized.

backend/location/models.py
```diff
@@ -0,0 +1,49 @@
...
+
+    osm_id = models.PositiveIntegerField(
+        help_text="The internal id of this location openstreetmaps",
+        verbose_name="OSM ID"
+    )
+
+    place_id = models.PositiveIntegerField(
+        help_text="Nominatim's place id of this location",
+        verbose_name="Place ID"
+    )
```
ddhanesha, line 49: I'd recommend to add `Meta` class. And a return string
```
class Meta:
    verbose_name = "Location"
    verbose_name_plural = "Location"

def __str__(self):
    return "%s" % (self.name)
```

backend/organization/management/commands/convert_organization_location.py
```diff
@@ -0,0 +1,39 @@
...
+                    organization.save()
+                    org_counter += 1
+                else:
+                    print(
+                        "Geo location not found for organization {}".format(organization.id)
+                    )
+            else:
+                print("Location not found for organization {}".format(organization.id))
+        
+        print("Total organization updated: {}".format(org_counter))
```
ddhanesha, line 39: I know still in WIP but just adding this comment so we remember: Same here we won't need this anymore or we could update it to use `Location object`

backend/organization/models/organization.py
```diff
@@ -91,6 +91,31 @@ class Organization(models.Model):
...
+
+    longitude = models.CharField(
+        help_text="Longitude of organization's location",
+        verbose_name="Longitude",
+        max_length=512,
+        null=True,
+        blank=True
+    )
+
     short_description = models.TextField(
```
ddhanesha, line 119: The migration has `loc` as a foreign key, but the Organization model is missing.

ddhanesha: backend/organization/migrations/0060_auto_20210121_1040.py

backend/organization/views/project_views.py
```diff
@@ -274,22 +284,13 @@ def patch(self, request, url_slug, format=None):
...
-        if 'description' in request.data:
-            project.description = request.data['description']
-        if 'country' in request.data:
-            project.country = request.data['country']
-        if 'city' in request.data:
-            project.city = request.data['city']
+        if 'location' in request.data:
+            project.location = request.data['location']
+        if 'location' in request.data:
+            location = get_location(request.data['loc'])
```
ddhanesha, line 290: This would need to change to use `Location` table.

backend/climateconnect_api/views/user_views.py
positiveimpact: You're right, we need to change this. But I'm only testing with edit project atm, so this can't be the problem

positiveimpact: But you're right, we have to remove this

backend/location/migrations/0004_auto_20210121_1452.py
positiveimpact: Good point. done (locally)!

backend/location/models.py
positiveimpact: Good point!

backend/climateconnect_api/serializers/user.py
```diff
@@ -60,19 +67,25 @@ def get_first_name(self, obj):
 
     def get_last_name(self, obj):
         return obj.user.last_name
+    
+    def get_location(self, obj):
+        if obj.location == None:
+            return None
+        return obj.location.name
```
ddhanesha, line 74: Nit: You can simply this to `return obj.location.name if obj.location else None`

backend/location/utility.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+    elif isinstance(input_polygon, MultiPolygon):
+        polygons = list(input_polygon)
+        switched_multipolygon = []
+        for polygon in polygons:
+            switched_polygon = get_polygon_with_switched_coordinates(polygon)
+            switched_multipolygon.append(switched_polygon)        
+        return MultiPolygon(switched_multipolygon)
+    else:
+        raise Exception("Wrong input")
```
ddhanesha, line 64: I am going to change the exception message here. This is too generic. It would be harder to debug if it happens in production

backend/organization/management/commands/convert_organization_location.py
ddhanesha: I think we will have to change these script to use `get_location` you wrote instead of google maps.

backend/organization/models/project.py
```diff
@@ -110,6 +113,25 @@ class Project(models.Model):
...
     )
 
+    # Field not in use. Keeping temporarily for backwards compatibility
+    location = models.CharField(
+        help_text="Project location",
+        verbose_name="Location",
+        max_length=2048,
+        null=True,
+        blank=True
+    )
```
ddhanesha, line 123: We can remove this field.

ddhanesha: or replace this with `loc` column.

backend/organization/views/project_views.py
```diff
@@ -219,6 +221,13 @@ def patch(self, request, url_slug, format=None):
             project = Project.objects.get(url_slug=url_slug)
         except Project.DoesNotExist:
             return Response({'message': 'Project not found: {}'.format(url_slug)}, status=status.HTTP_404_NOT_FOUND)
+        
+        # Author: Dip
+        # Code formatting here. So fields are just pass through so combing them and using setattr method insted.
+        pass_through_params = ['collaborators_welcome', 'description', 'helpful_connections','short_description', 'website']
+        for param in pass_through_params:
+            if param in request.data:
+                setattr(project, param, request.data[param])
```
ddhanesha, line 230: Didn't you report that there was an issue with this logic? Is it still happening?

positiveimpact: I think I fixed that. As far as I remember the problem was that it was trying access `request.data.param` rather than `request.data[param]`

backend/organization/models/project.py
positiveimpact: Yep, would remove this and rename loc to location

backend/climateconnect_api/utility/notification.py
```diff
@@ -34,7 +36,7 @@ def create_email_notification(receiver, chat, message_content, sender, notificat
         number_of_participants = Participant.objects.filter(chat=chat, is_active=True).count()
         is_group_chat = number_of_participants > 2
         if is_group_chat:
-            if email_settings['email_on_group_chat_message'] == True:
```
positiveimpact, line 37: Maybe we should rather create an env variable disabling notifications all together. 
I think it would make sense to do that in a seperate PR

ddhanesha: Okay, I will remove this then.

backend/location/utility.py
```diff
@@ -0,0 +1,209 @@
...
+    if location['type'] != "administrative":
+        return False
+    # short circuit if the address contains any information other than country and country code
+    for key in location['address'].keys():
+        if key not in ["country", "country_code"]:
+            return False       
+    return True
+
+def get_first_part(address, order):
+    for el in order:
```
ddhanesha, line 167: Can you have object name a bit more descriptive? `el` does not make sense what it means.

ddhanesha (approved):
Approved with some small suggestions. I tested both new and backwards compatible code. Works good.

backend/climateconnect_api/serializers/user.py
```diff
@@ -60,19 +68,40 @@ def get_first_name(self, obj):
...
```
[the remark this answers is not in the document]
positiveimpact: It is. We still set the location.name property in legacy mode. It's simply `profile.city + ", " + profile.country`

PR #451, opened 2021-01-19
Feature/leave group chat

### LINKED ISSUE #448: Allow leaving a group chat that you're part of
# Current behavior
Once you're added to a group chat you cannot leave it again. Only the creator can remove you from it

# Expected future behavior
When you are part of a group chat you should have the option to leave the chat in the chat interface

- [x] In the group chat interface there should be a red "Leave" button in the top right corner. On mobile, this might have to be placed in an expanding menu
- [x] When clicking this button there should be a dialog asking you to confirm whether you really want to leave. You can use `/frontend/src/components/dialogs/ConfirmDialog.js` for this.
- [x] There should be an API endpoint 'chat/<str:chat_uuid>/leave' which allows you to leave a chat. You can use the `is_active` column of the `MessageParticipants` table for this. When a user is not part of the chat anymore they shouldn't receive any more messages or notifications from the chat
- [x] When the user confirms that they want to leave the frontend should make a request to the 'chat/<str:chat_uuid>/leave' . On error it should show an error message to the user, on success it should redirect the user to the `/inbox` page

positiveimpact:
## Description
- Add button to allow users to leave a group chat
- Allow showing success/error messages in FixedHeight Layout - This layout is being used by inbox and chat and couldn't show any alerts before.
- Create an endpoint for leaving a chat. This is done by setting the `is_active` flag to `False`. If you were the creator of the chat this role is passed on to a random user
- Only show a chat, its messages, and notifications about this chat to users who have the `is_active` flag set to `True`
- Only let users who have the `is_active` flag set to `True` interact with a chat

Closes #448

## Todo


frontend/src/components/layouts/FixedHeightLayout.js
```diff
@@ -19,13 +22,44 @@ const useStyles = makeStyles({
...
+  messageType
+}) {
   const classes = useStyles();
-
+  const [initialMessageType, setInitialMessageType] = React.useState(null);
+  const [alertOpen, setAlertOpen] = React.useState(true);
+  const [initialMessage, setInitialMessage] = React.useState("");
+  useEffect(() => {
+    const params = getParams(window.location.href);
+    if (params.message) setInitialMessage(decodeURI(params.message));
```
piperchester, line 37: Nit: do you mind adding braces for one-liners? I find it slightly easier to debug. This is approaching bikeshedding IMO so definitely welcome a push back here 🙏 
```suggestion
    if (params.message) {
        setInitialMessage(decodeURI(params.message));
    }
```

frontend/src/components/layouts/FixedHeightLayout.js
```diff
@@ -19,13 +22,44 @@ const useStyles = makeStyles({
...
+    if (params.errorMessage) {
```
piperchester, line 38: Can we guarantee that `params` will never be `undefined`? 

```suggestion
    if (params?.errorMessage) {
```

frontend/src/components/layouts/FixedHeightLayout.js
```diff
@@ -19,13 +22,44 @@ const useStyles = makeStyles({
...
+            className={classes.alert}
+            severity={
+              messageType ? messageType : initialMessageType ? initialMessageType : "success"
+            }
+            onClose={() => {
+              if(message)
+                setAlertOpen(false);
+              else
+                setInitialMessage(null)
+            }}
```
piperchester, line 58:
```suggestion
              if (message) {
                setAlertOpen(false);
              } else {
                setInitialMessage(null);
              }
            }}
```

frontend/pages/chat/[chatUUID].js
```diff
@@ -196,6 +239,10 @@ Chat.getInitialProps = async (ctx) => {
     getChatMessagesByUUID(ctx.query.chatUUID, token, 1),
     getRolesOptions(),
   ]);
+  if(!chat)
+    return {
+      chat_id: null
+    }
```
Comment (piperchester) [inline frontend/pages/chat/[chatUUID].js:245]:
```suggestion
  if (!chat) {
    return {
      chat_id: null
    }
  }
```

backend/chat_messages/consumer.py
```diff
@@ -51,7 +51,7 @@ async def new_message(self, chat_uuid, user, message_content):
             )
         except MessageParticipants.DoesNotExist:
             chat = None
-        receiver_user_ids = Participant.objects.filter(chat=chat).values_list('user', flat=True)
+        receiver_user_ids = Participant.objects.filter(chat=chat, is_active=True).values_list('user', flat=True)
```
piperchester, line 54:
```suggestion
        # Only select active participant IDs 
        receiver_user_ids = Participant.objects.filter(chat=chat, is_active=True).values_list('user', flat=True)
```

piperchester (approved):
Nice! I'm still getting context on the chat subsystem and wasn't able to sanity check this locally as I don't have chat set up, but code looks good to me.

frontend/src/components/layouts/FixedHeightLayout.js
```diff
@@ -19,13 +22,44 @@ const useStyles = makeStyles({
   },
 });
 
-export default function FixedHeightLayout({ title, children }) {
+export default function FixedHeightLayout({ 
+  title, 
+  children,
+  message,
+  messageType
```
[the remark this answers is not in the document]
positiveimpact: Ah, I didn't know about that! Thanks for the hint!

frontend/src/components/layouts/FixedHeightLayout.js
```diff
@@ -19,13 +22,44 @@ const useStyles = makeStyles({
...
+}) {
   const classes = useStyles();
-
+  const [initialMessageType, setInitialMessageType] = React.useState(null);
+  const [alertOpen, setAlertOpen] = React.useState(true);
+  const [initialMessage, setInitialMessage] = React.useState("");
```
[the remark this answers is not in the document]
positiveimpact: Makes sense! I need to get used to this because I used to always leave the `React.` in there but I'll import the hook in the future and call it without `React.`

frontend/src/components/layouts/FixedHeightLayout.js
positiveimpact: I agree that this makes it easier to read so I'm fine with doing this in the future. As long as it offers real benefits such as readability I always welcome suggestions

frontend/src/components/layouts/FixedHeightLayout.js
positiveimpact: Yeah, getParams always returns an object or an empty array. Therefore it's never undefined.

PR #455, opened 2021-01-27
Set up python celery for asynchronous tasks

### LINKED ISSUE #419: Set up python celery for asynchronous tasks
This could be super useful to asynchronously handle tasks - such as adding a user to the newsletter subscription list on sign up - without having to wait for the task to finish. 
https://docs.celeryproject.org/en/stable/django/first-steps-with-django.html

ddhanesha:
## Description

This is setup for python celery that we will be using for asynchronous task.


backend/climateconnect_main/settings.py
```diff
@@ -212,3 +210,7 @@
         }
     }
 }
+
+# For celery we are using redis as a broker URL
```
piperchester, line 214: Neat. What was the rationale on choosing Redis (over another option like RabbitMQ) as our broker? Just more familiarity and we're already using it in the project?

```suggestion
# For Celery we use Redis as a broker URL
```

backend/climateconnect_main/celery.py
```diff
@@ -0,0 +1,14 @@
...
+from celery import Celery
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'climateconnect_main.settings')
+
+app = Celery('climateconnect_main')
+
+# Using a string here means the worker doesn't have to serialize
+# the configuration object to child processes.
+# - namespace='CELERY' means all celery-related configuration keys
+#   should have a `CELERY_` prefix.
```
piperchester, line 12: Good call, thanks for adding this.

piperchester, line 12: Good call, thanks for adding this. Is the string for 'climateconnect_main' or the 'django.conf:settings' bit?

piperchester (approved):
Flyby comments with some nits. From my limited BE context on the project 😅 things look good! Purely out of curiosity I'd love to learn more about how we're thinking about the design and usages for the task queue.

backend/climateconnect_main/celery.py
ddhanesha: Its is for django.conf:settings -> climateconnect_main/settings.py

backend/climateconnect_main/settings.py
ddhanesha: Two reasons....

1) We are already using Redis for chat messaging. i.e.: For our integration of django-channels (https://channels.readthedocs.io/en/stable/). 

2) The usage we are thinking here we do not require a dedicated message broker like RabbitMQ. Also we have more familiarity with Redis than RabbitMQ.

I would like your input though. I have found that Celery with Redis works pretty well for most use cases.

ddhanesha:
> Flyby comments with some nits. From my limited BE context on the project 😅 things look good! Purely out of curiosity I'd love to learn more about how we're thinking about the design and usages for the task queue.

I have couple use cases where we can use Celery
1) Spawning async task on Signup to send Email verification to our users. Right now we are calling all methods synchronous in the API.
2) Sending web push notifications to offline users. (This is debatable) 
3) We can also leverage Celery for cron jobs (i.e. Periodic tasks)
4) We can use Celery for queuing mechanism (Like: AWS Simple Queue Service) 

Happy to chat more about it offline. :)

backend/climateconnect_main/settings.py
piperchester: Awesome, thanks for the context. I've limited experience with both (I'm not familiar with how one or the other performs under scale for our use cases) so I'd err on the side of more familiarity.

piperchester:
> Happy to chat more about it offline. :)

Nice! Ok updated the wiki with these to build context: https://github.com/climateconnect/climateconnect/wiki/Celery

backend/climateconnect_main/settings.py
ddhanesha: Yeah I think redis can handle more complex and performs well under scale. I have used redis heavily before. But if time comes it easy to change the from redis to rabbtimq.

ddhanesha:
Merging this PR... I will make another PR with deployment script.

Issue #288, opened 2020-08-01
Make filters change the url so you can send somebody a link of the filtered page


positiveimpact:
For example when filtering by location="San Francisco", the url should automatically change to `urlEncode(climateconnect.earth/?location="San Francisco")`.
You could then send that link to a friend to directly link them to all the projects there are in San Francisco.

positiveimpact:
@vliaskov Are you still actively working on this issue or should I re-assign it to somebody else?

positiveimpact:
I will unassign this for now. You can of course assign it to yourself again if you're interested on working on it again in the future @vliaskov

piperchester:
I'll take a look at this @vliaskov 👍

PR #456, opened 2021-01-27
Implement persisted query params and various filtering improvements

piperchester:
## Description

This feature includes various filtering UX improvements, including implementing #288. I ended up genericizing the little util that already existed into a shared location that could be reused. 

Now when a user is filtering, we immediately filter (there's no longer an "Apply" button). Also if a user enters a persisted URL with specific query params, we force those UI elements to reflect that selection on window load. 

https://user-images.githubusercontent.com/1530684/113189205-c871d480-920f-11eb-8fcb-e6d7fcda9718.mp4

## Todo

- [x] Manually test more use cases
- [ ] Fix couple remaining bugs (see linked issues)
- [x] Handle location (@positiveimpact is taking a stab at this 💯 🏆)


frontend/src/components/browse/BrowseContent.js
```diff
@@ -141,12 +142,40 @@ export default function BrowseContent({
...
+   * automatically change to active filters. This enables filtered searches
+   * to persist, so that they can be easily shareable to other users.
+   *
+   * Builds a URL and updates window state. E.g. something like:
+   * http://localhost:3000/browse?&country=Austria&city=vienna&
+   */
+  const persistFiltersInURL = (activeFilters) => {
+    // Build query params, include the `?` but ignore
+    // the standalone 'ampersand' case when no active filters are present.
+    const filteredParams = buildUrlEndingFromFilters(activeFilters);
```
piperchester, line 156: We could likely refactor this function to not return an empty ampersand, but wasn't sure of its other use cases. So handling this directly.

frontend/src/components/browse/BrowseContent.js
```diff
@@ -141,12 +142,40 @@ export default function BrowseContent({
...
+
+    // Build a URL with properties. E.g., /browse?...
+    const origin = window?.location?.origin;
+    const pathname = window?.location?.pathname;
+    const newUrl = `${origin}${pathname}${filteredQueryParams}`;
+
+    // Only push state if there's a URL change
+    if (newUrl !== window?.location?.href) {
+      window.history.pushState({}, "", newUrl);
+    }
```
piperchester, line 167: Ended up pushing window state. Not necessary, but this seemed like a best practice - WDYT?

piperchester:
@ reviewers, I'd appreciate a sanity check on deduping the function into that shared util 🙏

frontend/src/utils/index.js
```diff
@@ -0,0 +1,17 @@
...
+    if (filters[filterKey] && filters[filterKey].length > 0) {
+      if (Array.isArray(filters[filterKey]))
+        url += encodeURI(filterKey + "=" + filters[filterKey].join()) + "&";
+      else url += encodeURI(filterKey + "=" + filters[filterKey] + "&");
+    }
+  });
+  return url;
+};
+
+export { buildUrlEndingFromFilters };
```
positiveimpact, line 17: We have used the folder `/public/lib` for that so far. I think this function would fit there better.

frontend/src/components/browse/BrowseContent.js
positiveimpact: I don't believe we really need the empty ampersand anywhere but to be safe we'd have to test it with the other places where its used.

frontend/src/components/browse/BrowseContent.js
positiveimpact: That makes a lot of sense! This way people can go back to their last filtered query.

positiveimpact (changes requested):
Great work so far! I haven't documented the issue super well but I actually meant one more feature to be included in this:
When you go to a filtered link such as `http://localhost:3000/browse?status=In%20Progress&` this link should directly have the filter applied. So if you go to that link you should have the filter applied and only see projects that are in progress.
The current features look good other than one comment that I left!

frontend/src/utils/index.js
piperchester: Nice, TIL. I'll move there 👍 Why is it named /public`?

piperchester:
> When you go to a filtered link such as `http://localhost:3000/browse?status=In%20Progress&` this link should directly have the filter applied. So if you go to that link you should have the filter applied and only see projects that are in progress.

Love that idea! Ok can do 👍 for the sake of smaller PRs, what do you think about decoupling this feature from that feature?

frontend/src/utils/index.js
positiveimpact: public is usually used for images and other static data because I wanted to keep src only for components. But thinking about it putting it in src actually makes much more sense. Maybe we can move it in the future (in a seperate PR)

positiveimpact (changes requested):
LGTM on the browse page now!
One thing I just saw is that this doesn't work on the hubs pages yet and I think it should work equally in every place where you can filter. Maybe it would make sense to have the logic in `BrowseContent` as that is used by `browse` as well as all topic hubs and for example potential future city hubs.

piperchester:
> One thing I just saw is that this doesn't work on the hubs pages yet and I think it should work equally in every place where you can filter. Maybe it would make sense to have the logic in `BrowseContent` as that is used by `browse` as well as all topic hubs and for example potential future city hubs.

Sounds good! Ok potentially dumb question: I think we'll need to add some hubs test data to see hubs on `/hubs`? Or should I be looking at a different URL for browsing? I could work on contributing that to the `create_test_data` command.
Screen Shot 2021-02-03 at 6 45 09 PM

positiveimpact:
> > One thing I just saw is that this doesn't work on the hubs pages yet and I think it should work equally in every place where you can filter. Maybe it would make sense to have the logic in `BrowseContent` as that is used by `browse` as well as all topic hubs and for example potential future city hubs.
> 
> Sounds good! Ok potentially dumb question: I think we'll need to add some hubs test data to see `/hubs`? Or should I be looking at a different URL for browsing?
> 
> Screen Shot 2021-02-03 at 6 45 09 PM
> 
> I could work on contributing that to the `create_test_data` command -- I'd be in favor of landing the work as is, and then iterating on hubs parity in a future PR along with the search feature - wdyt @positiveimpact ?

Oops, yep you'll need to add a hub in the database. Maybe it'd be worth it to add that to the test data script.

frontend/pages/browse.js
```diff
@@ -129,31 +132,19 @@ export default function Browse({
...
         <BrowseContent
-          initialProjects={projectsObject}
-          initialOrganizations={organizationsObject}
-          initialMembers={membersObject}
           applyNewFilters={applyNewFilters}
+          applySearch={applySearch}
           filterChoices={filterChoices}
+          initialMembers={membersObject}
+          initialOrganizations={organizationsObject}
+          initialProjects={projectsObject}
```
piperchester, line 140: Just alphabetizing props to improve readability.

frontend/pages/hubs.js
```diff
@@ -1,9 +1,10 @@
 import React from "react";
+import axios from "axios";
+import { Typography, Container, makeStyles, useMediaQuery } from "@material-ui/core";
+
```
piperchester, line 4: Nit: I like to group 3rd party deps at the top of the file, and then local files below them. I find this makes it easier to grok a file. Maybe a good addition to a "codebase health and style" guide? Let me know if this doesn't align with with anyone's preferences and happy to not do it 🙏

frontend/pages/hubs/[hubUrl].js
```diff
@@ -43,20 +45,19 @@ const useStyles = makeStyles((theme) => ({
...
+  filterChoices,
   headline,
+  hubUrl,
+  image_attribution,
   image,
+  initialOrganizations,
+  initialProjects,
+  name,
   quickInfo,
+  statBoxTitle,
```
Comment (piperchester) [inline frontend/pages/hubs/[hubUrl].js:57]: Just alphabetizing props to improve readability.

frontend/pages/hubs/[hubUrl].js
```diff
@@ -110,12 +140,15 @@ export default function Hub({
         page: 1,
         token: token,
         urlEnding: newUrlEnding,
+        // TODO: I think this is the primary difference between the applyNewFilters logic
+        // here locally in [hubUrl] and within browse.js -- we should deduplicate
```
Comment (piperchester) [inline frontend/pages/hubs/[hubUrl].js:144]: @positiveimpact added this TODO - I _think_ we can dedupe this `applyNewFilters` logic with what's in `browse` but didn't want to take that on in this PR, so I duplicated the query param persistence here for now.

frontend/pages/hubs/[hubUrl].js
```diff
@@ -110,12 +140,15 @@ export default function Hub({
...
         hubUrl: hubUrl,
       });
+
       if (type === "members") {
         filteredItemsObject.members = membersWithAdditionalInfo(filteredItemsObject.members);
       }
-      console.log(filteredItemsObject);
```
Comment (piperchester) [inline frontend/pages/hubs/[hubUrl].js:118]: Nit: assumed this was unneeded anymore.

frontend/pages/hubs/[hubUrl].js
```diff
@@ -199,18 +232,6 @@ const HubDescription = ({ hub }) => {
...
-  Object.keys(filters).map((filterKey) => {
-    if (filters[filterKey] && filters[filterKey].length > 0) {
-      if (Array.isArray(filters[filterKey]))
-        url += encodeURI(filterKey + "=" + filters[filterKey].join()) + "&";
-      else url += encodeURI(filterKey + "=" + filters[filterKey] + "&");
-    }
-  });
-  return url;
-};
-
```
Comment (piperchester) [inline frontend/pages/hubs/[hubUrl].js:213]: Hoisted this into the `urlParsing` lib.

frontend/pages/hubs.js
positiveimpact: I think this makes a lot of sense :+1:

frontend/pages/hubs/[hubUrl].js
```diff
@@ -95,10 +96,39 @@ export default function Hub({
...
 
+  /**
+   * For example when filtering by location="San Francisco", the url should
+   * automatically change to active filters. This enables filtered searches
+   * to persist, so that they can be easily shareable to other users.
+   *
+   * Builds a URL and updates window state. E.g. something like:
+   * http://localhost:3000/browse?&country=Austria&city=vienna&
+   */
+  const persistFiltersInURL = (activeFilters) => {
```
Comment (positiveimpact) [inline frontend/pages/hubs/[hubUrl].js:107]: This is being executed twice now if you're on a hub page.
1. The first time when `handleApplyNewFilters` is being called in `BrowseContent`
2. `handleApplyNewFilters` then calls the `applyNewFilters` function in `[hubUrl].js` which executes this function again

I don't think we need this here at all since `[hubUrl].js` and `browse.js` both implement the `BrowseContent` component and `browse.js` also doesn't use this function additionally.

frontend/pages/hubs/[hubUrl].js
```diff
@@ -95,10 +96,39 @@ export default function Hub({
...
+    }
+  };
+
   const applyNewFilters = async (type, newFilters, closeFilters, oldUrlEnding) => {
     if (filters === newFilters) {
       return;
     }
+
+    // Save these filters as query params to the URL
+    persistFiltersInURL(newFilters);
```
Comment (positiveimpact) [inline frontend/pages/hubs/[hubUrl].js:130]: This should also be removed (explanation in my above comment)

frontend/pages/hubs/[hubUrl].js
Comment (positiveimpact) [reply frontend/pages/hubs/[hubUrl].js:144]: Yep, definitely possible to dedupe atm. I did it like this on purpose incase they deviate further from each other in the future, but it's probably better to just make once function in `BrowseContent`.

positiveimpact (changes requested):
Added a few minor comments and found a potential problem:
We'll have to escape some characters when building the url. For example locally I filtered by the category "Gastronomy/Catering" and this resulted in the following url: 
`http://localhost:3000/hubs/food?&category=Gastronomy/Catering&`
The `/` could potentially cause problems, so I think it'd be worth it to escape it and think about what other characters could cause problems.

piperchester:
Refactored to use `encodeURIComponent` to handle reserved characters. @positiveimpact from what I could tell this should be fine where it's used to call the server, like here https://github.com/climateconnect/climateconnect/blob/master/frontend/pages/browse.js#L43 - but it'd be great to sanity check that with you.

Also deleted the 2nd persist func, and added tests to handle cases.

piperchester:
Refactored to use `encodeURIComponent` to handle reserved characters. @positiveimpact from what I could tell this should be fine where it's used to call the server, like here https://github.com/climateconnect/climateconnect/blob/master/frontend/pages/browse.js#L43 - but it'd be great to sanity check that with you.

Also deleted the 2nd persist func, and added tests to handle cases.

positiveimpact (changes requested):
Filtering doesn't work for me anymore with this change. Seems like the `=` component is being escaped, so the request to the server looks like this: 
`/api/projects/?page=1&hub=food&category%3DLowering%20food%20waste%26`
I think we should only encode the actual values that we're filtering for. And even there we'll have to test whether all of these are detected by the backend.

piperchester:
@positiveimpact I just pushed the latest bulk of changes, but keeping this in Draft until I do a final clean up and resolve the open TODOs. Almost there 😰

frontend/pages/browse.js
```diff
@@ -1,21 +1,25 @@
...
   membersWithAdditionalInfo,
 } from "../public/lib/getOptions";
+import { encodeQueryParamsFromFilters } from "../public/lib/urlOperations";
 import { parseData } from "../public/lib/parsingOperations";
 import BrowseContent from "../src/components/browse/BrowseContent";
-import TopOfPage from "../src/components/hooks/TopOfPage";
 import HubsSubHeader from "../src/components/indexPage/HubsSubHeader";
 import MainHeadingContainerMobile from "../src/components/indexPage/MainHeadingContainerMobile";
+import tokenConfig from "../public/config/tokenConfig";
+import TopOfPage from "../src/components/hooks/TopOfPage";
```
piperchester, line 22: Alphabetized imports, and grouped into 3rd party / built-ins and relative imports for maintainability. I took this approach in other files too...happy to back this out if undesired! Otherwise we should adopt as one of our codebase style guidelines IMO.

frontend/pages/browse.js
```diff
@@ -26,34 +30,52 @@ export default function Browse({
...
+  /**
+   * Fetches data from the server based on the newly provided
+   * filters. Returns an object with the new filter data, as well
+   * as other options.
+   *
+   * @param {string} type one of "projects", "members", "organizations"
+   * @param {Object} newFilters something like {"location": "", status: [], etc... }
+   * @param {boolean} closeFilters
+   * @param {string} oldUrlEnding previous end of URL through the query param
+   */
```
piperchester, line 50: Added some misc. comments throughout to help with debugging and what not. Happy to back / tweak any of them. 

This feature work made me really want to start incorporating TypeScript into the project after we build out more functionality and UX improvements.

frontend/pages/browse.js
```diff
@@ -213,7 +238,12 @@ async function getMembers(page, token, urlEnding) {
 
 async function getDataFromServer({ type, page, token, urlEnding }) {
   let url = `${process.env.API_URL}/api/${type}/?page=${page}`;
-  if (urlEnding) url += urlEnding;
+
+  // Handle query params as well
+  if (urlEnding) {
+    // &category=Lowering%20food%20waste&
+    url += urlEnding;
+  }
```
piperchester, line 246: For single line `if` I generally just add two braces for readability. Happy to back out if undesired.

frontend/pages/hubs/[hubUrl].js
```diff
@@ -44,20 +45,16 @@ const useStyles = makeStyles((theme) => ({
...
-  hubUrl,
-  name,
-  statBoxTitle,
   headline,
+  hubUrl,
+  image_attribution,
   image,
+  name,
   quickInfo,
+  statBoxTitle,
```
Comment (piperchester) [inline frontend/pages/hubs/[hubUrl].js:54]: Alphabetizing props.

frontend/pages/hubs/[hubUrl].js
```diff
@@ -189,13 +190,8 @@ export default function Hub({
           <div ref={contentRef} className={classes.contentRef} />
           <BrowseExplainer />
           <BrowseContent
-            initialProjects={initialProjects}
-            initialOrganizations={initialOrganizations}
-            filterChoices={filterChoices}
-            loadMoreData={loadMoreData}
```
Comment (piperchester) [inline frontend/pages/hubs/[hubUrl].js:195]: I _believe_ these are unused now. Need to confirm w/ Chris.

frontend/src/components/browse/BrowseContent.js
```diff
@@ -31,17 +34,18 @@ const useStyles = makeStyles((theme) => {
...
-  handleSetErrorMessage,
   errorMessage,
+  filterChoices,
+  handleSetErrorMessage,
+  hideMembers,
+  initialFiltersExpanded,
+  initialMembers,
+  initialOrganizations,
+  initialProjects,
+  loadMoreData,
```
piperchester, line 48: Alphabetizing.

frontend/src/components/browse/BrowseContent.js
```diff
@@ -113,6 +126,46 @@ export default function BrowseContent({
...
+   * query params from a filter in it. In this use
+   * case, we should automatically set the filters dynamically. Ensure
+   * that this isn't invoked on extraneous renders.
+   */
+  let hasFilteredFromQueryParams = false;
+  useEffect(() => {
+    if (hasQueryParams && !hasFilteredFromQueryParams) {
+      // Update the state of the visual filters, like Select, Dialog, etc
+      // Then actually fetch the data. We need a way to map what's
+      // in the query param, to what's UI control is present on the screen. For
```
piperchester, line 144:
```suggestion
      // in the query param, to what UI element is present on the screen. For
```

frontend/src/components/browse/BrowseContent.js
```diff
@@ -113,6 +126,46 @@ export default function BrowseContent({
...
+        if (value.indexOf(",") > 0) {
+          queryObject[key] = value.split(",");
+        }
+      }
+
+      let newFilters = { ...queryObject };
+
+      // TODO(piper): fix hardcoded projects here...
+      // Apply new filters with the query object immediately:
+      handleApplyNewFilters("projects", newFilters, false, state.urlEnding["projects"]);
```
piperchester, line 162: @piperchester -- need to remove the hardcoded "projects" bit here: 
 
```suggestion
      // Apply new filters with the query object immediately:
      handleApplyNewFilters("projects", newFilters, false, state.urlEnding["projects"]);
```

frontend/src/components/filter/FilterContent.js
```diff
@@ -26,20 +28,81 @@ export default function FilterContent({
...
+      // Handle string values too. If there's something set, and it's strings concat'd -
+      // we need to make an array out of them
+      if (value && value.indexOf(",") > 0) {
+        const splitItems = value.split(",");
+        reducedPossibleFilters[key] = [...splitItems];
+      } else {
+        reducedPossibleFilters[key] = value;
+      }
+    }
+  });
```
piperchester, line 65: Would appreciate close review and testing here ☝️

frontend/src/components/filter/FilterContent.js
```diff
@@ -26,20 +28,81 @@ export default function FilterContent({
...
 
   const [open, setOpen] = React.useState({});
 
+  const [currentFilters, setCurrentFilters] = React.useState(reducedPossibleFilters);
+
+  // TODO(Piper): I believe the selected items is only being set by the multilevel?
+  // Could merge the initialProjeccategories here and then pass that down as well!
```
piperchester, line 72:
```suggestion
```

frontend/src/components/filter/FilterContent.js
```diff
@@ -26,20 +28,81 @@ export default function FilterContent({
...
+            // Only one item in the array,
+            accumulator[currentValue.key] = [];
+
+            // Ensure we've an array for multiple items
+            const splitItems = currentFilters[currentValue.key].split(",");
+
+            // Have to transform to a name property, so that the items can render the name
+            const selectedCategory = {
+              // Old just a string
+              // name: currentFilters[currentValue.key],
```
piperchester, line 94:
```suggestion
```

frontend/src/components/filter/FilterContent.js
```diff
@@ -26,20 +28,81 @@ export default function FilterContent({
...
+            accumulator[currentValue.key] = currentFilters[currentValue.key];
+          } else {
+            // Not an array, need to handle differently.
+            // Only one item in the array,
+            accumulator[currentValue.key] = [];
+
+            // Ensure we've an array for multiple items
+            const splitItems = currentFilters[currentValue.key].split(",");
+
+            // Have to transform to a name property, so that the items can render the name
```
piperchester, line 91:
```suggestion
            // Have to transform to key on a "name" property, so that ListItem text can render the text correctly
```

frontend/src/components/filter/FilterContent.js
```diff
@@ -51,107 +114,120 @@ export default function FilterContent({
...
       [filterKey]: currentFilters[filterKey].filter((f) => f !== filterName),
-    });
-    if (selectedItems[filterKey])
+    };
+
+    // TODO(piper): verify this:
+    // When dismissing a selected filter chip, we also want to update the
+    // window state to reflect the currently active filters, and fetch
+    // the updated data from the server
+    applyFilters(type, updatedFilters, isSmallScreen);
```
piperchester, line 145: @piperchester to verify this:

```suggestion
    // When dismissing a selected filter chip, we also want to update the
    // window state to reflect the currently active filters, and fetch
    // the updated data from the server
    applyFilters(type, updatedFilters, isSmallScreen);
```

frontend/src/components/general/MultiLevelSelector.js
```diff
@@ -164,13 +166,13 @@ const useStyles = makeStyles((theme) => {
...
+  dragAble,
+  isInPopup,
+  itemNamePlural,
   itemsToSelectFrom,
   maxSelections,
-  itemNamePlural,
-  isInPopup,
-  dragAble,
+  selected,
+  setSelected,
```
piperchester, line 175: Alphabetizing.

frontend/src/components/general/MultiLevelSelector.js
```diff
@@ -365,7 +371,7 @@ function SelectedList({
                           onClick={() => onClickUnselect(item)}
                           disableRipple
                         >
-                          <ListItemText>{item.name}</ListItemText>
+                          <ListItemText>test {item.name}</ListItemText>
```
piperchester, line 374:
```suggestion
                          <ListItemText>{item.name}</ListItemText>
```

frontend/src/components/project/ProjectPreview.js
```diff
@@ -10,13 +11,15 @@ const useStyles = makeStyles((theme) => {
...
+        "box-shadow": "2px 2px 1px #EEE",
       },
       "-webkit-user-select": "none",
       "-moz-user-select": "none",
       "-ms-user-select": "none",
       userSelect: "none",
       backgroundColor: theme.palette.background.default,
-      borderRadius: 0,
+      borderRadius: 3,
+      border: `1px solid #EEE`,
```
piperchester, line 22: I tweaked the card style slightly here.

frontend/src/components/browse/BrowseContent.js
piperchester: @piperchester -- need to remove the hardcoded "projects" bit here:

frontend/src/components/filter/Filters.js
```diff
@@ -74,37 +76,43 @@ const useStyles = makeStyles((theme) => {
...
       )}
+
+      {/* TODO(piper): add this functionality to all elements where we apply new filters */}
       <div className={`${classes.flexContainer} ${isInOverlay && classes.verticalFlexContainer}`}>
         {possibleFilters.map((filter) => {
+          // Get the current values for each potential filter
+          // from what could already be previously selected
+          const currentFilterValue = currentFilters[filter.key];
+
+          // TODO(Piper): verify this is working post merge
```
piperchester, line 115:
```suggestion
```

frontend/src/components/filter/Filters.js
```diff
@@ -74,37 +76,43 @@ const useStyles = makeStyles((theme) => {
...
+
   return (
     <>
       {errorMessage && (
         <div className={classes.errorMessageWrapper}>
           <Typography color="error">{errorMessage}</Typography>
         </div>
       )}
+
+      {/* TODO(piper): add this functionality to all elements where we apply new filters */}
```
piperchester, line 108:
```suggestion
```

frontend/src/components/browse/BrowseContent.js
```diff
@@ -113,6 +126,45 @@ export default function BrowseContent({
...
+      for (const [key, value] of Object.entries(queryObject)) {
+        if (value.indexOf(",") > 0) {
+          queryObject[key] = value.split(",");
+        }
+      }
+
+      let newFilters = { ...queryObject };
+
+      // Apply new filters with the query object immediately:
+      handleApplyNewFilters("projects", newFilters, false, state.urlEnding["projects"]);
```
piperchester, line 161: @positiveimpact I might need your help in genericizing this during testing 🙏

backend/requirements.txt
```diff
@@ -25,6 +25,7 @@ cryptography==3.3.2
 daphne==2.5.0
 Django==2.2.13
 django-cors-headers==3.2.1
+django-debug-toolbar==3.2
```
ddhanesha, line 28: Can you rebase to master once? Don't think this is necessary.

positiveimpact (changes requested):
Great progress overall, I really like how smooth everything is working now. 
There's only a few problems left on this huge PR:
- when you have no filters selected, the URL changes to `<URL>/browse?undefined` Would be better to have no query string in this case
- Location filtering doesn't really work with the URL. It records what you're typing when searching for a location but as soon as you select one (and therefore actually filter by a location), the string is gone from the URL. It would be best to only add a location to the URL when you select one and then save the place_id in the url
- When applying a filter in one tab, it will also show up in the other 2 tabs. However, it will only be applied when refreshing the page. I like this feature, but it should already apply when switching to the tab (or just not be applied to multiple tabs at once).
- There are some problems with MultiLevelSelect filters, I documented them in the comments

Again, great progress overall!

piperchester:
> Great progress overall, I really like how smooth everything is working now.
> There's only a few problems left on this huge PR:
> 
> * when you have no filters selected, the URL changes to `<URL>/browse?undefined` Would be better to have no query string in this case
> * Location filtering doesn't really work with the URL. It records what you're typing when searching for a location but as soon as you select one (and therefore actually filter by a location), the string is gone from the URL. It would be best to only add a location to the URL when you select one and then save the place_id in the url
> * When applying a filter in one tab, it will also show up in the other 2 tabs. However, it will only be applied when refreshing the page. I like this feature, but it should already apply when switching to the tab (or just not be applied to multiple tabs at once).
> * There are some problems with MultiLevelSelect filters, I documented them in the comments

Awesome; thanks @positiveimpact! Will start to work on these. Drafting again to signal WIP 👍

frontend/src/components/browse/BrowseContent.js
```diff
@@ -113,6 +126,45 @@ export default function BrowseContent({
...
+      // Update the state of the visual filters, like Select, Dialog, etc
+      // Then actually fetch the data. We need a way to map what's
+      // in the query param, to what UI element is present on the screen. For
+      // example, if we have a MultiLevelSelect dialog representing categories
+      // and we have a ?&category=Food waste, then we need to update the
+      // the MultiLevelSelect dialog's selection to that value somehow.
+
+      // For each query param option, ensure that it's
+      // split into array before spreading onto the new filters object
+      const queryObject = _.cloneDeep(router.query);
```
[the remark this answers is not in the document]
piperchester: Fixed in cc64863

piperchester:
@positiveimpact - still cleaning up some things. I'll take this out of Draft and re-request a review soon :+1:

positiveimpact:
@piperchester Added the code to make locations also persistent in URL. Would appreciate a review :)

frontend/pages/hubs/[hubUrl].js
```diff
@@ -264,21 +266,23 @@ export default function Hub({
...
             applyNewFilters={applyNewFilters}
             applySearch={applySearch}
-            hideMembers={!isLocationHub}
             customSearchBarLabels={customSearchBarLabels}
-            handleSetErrorMessage={handleSetErrorMessage}
             errorMessage={errorMessage}
-            hubQuickInfoRef={hubQuickInfoRef}
+            // TODO: is this still needed?
+            // filterChoices={filterChoices}
```
Comment (positiveimpact) [inline frontend/pages/hubs/[hubUrl].js:274]: This is still needed. If removed, you can't catch possibleFilters from `public/data/filterChoices.js` and will get this error: `Error: No filter choices supplied`

positiveimpact:
Everything should be working on the browse page now @piperchester 
Haven't started on the hubs page yet

piperchester:
Looking great so far @positiveimpact! Have started reading through. IMO once code review is over, we could land this into develop ahead of implementing it in Hubs. That'd make testing and review much easier I think. 

One thing I noticed in the console is we might have an infinite loop somewhere:

Screen Shot 2021-07-20 at 5 39 39 PM

frontend/src/components/filter/Filters.js
```diff
@@ -173,10 +184,22 @@ export default function Filters({
...
+             */
             const handleSetSelectedItems = (newSelectedItems) => {
-              setSelectedItems({ ...selectedItems, [filter.key]: newSelectedItems });
+              setSelectedItems({
+                ...selectedItems,
+                [filter.key]: newSelectedItems,
+              });
             };
+
+            // TODO: what is the showIf property used for?
```
piperchester, line 202: Just used as a flag?
```suggestion
```

frontend/public/lib/selectOptionsOperations.js
```diff
@@ -1,6 +1,11 @@
 export function parseOptions(options, parentPropertyName) {
+  //Return an array of parent properties with their subcategories. If the parent property of an item is missing: assume it's a top level property
```
piperchester, line 2: Thanks for adding this!
```suggestion
  // Returns an array of parent properties with their subcategories. If the parent property of an item is missing, assume it's a top-level property
```

frontend/public/lib/parsingOperations.js
```diff
@@ -36,3 +41,31 @@ export function getMessageFromUrl(message) {
   if (typeof message === "object") return message;
   else return decodeURIComponent(message).replaceAll("+", " ");
 }
+
+export function getInfoMetadataByType(type, locale) {
+  if (type === "organizations") return getOrganizationInfoMetadata(locale);
+  if (type === "profiles" || type === "members") return profile_info_metadata(locale);
+  if (type === "projects") return getProjectInfoMetadata();
```
piperchester, line 48: Nit: Nice. Why is `profile_info_metadata` snake cased, and the others are camel? We should normalize for consistency IMO.

frontend/public/lib/locationOperations.js
```diff
@@ -214,3 +217,34 @@ export function getLocationValue(values, locationKey) {
   }
   return values[locationKey];
 }
+
+//When filtering by location, the url only holds the place_id, osm_id and loc_type, but not the name
+//This function is used to retrieve the whole location object
```
piperchester, line 222: Nit: just comment formatting on func declaration vs. `//` 
```suggestion
/** 
 * When filtering by location, the url only holds the place_id, osm_id and loc_type, but not the name. This is used to retrieve the whole location object
*/
```

frontend/public/data/possibleFilters.js
```diff
@@ -3,23 +3,34 @@ import ExploreOutlinedIcon from "@material-ui/icons/ExploreOutlined";
...
     return getMembersFilters(filterChoices, texts);
+  } else if (key === "ideas") {
+    return getIdeasFilters(filterChoices, texts);
+  } else if (key === "all") {
+    const projectsFilters = getProjectsFilters(filterChoices, texts, english_texts);
+    const organizationsFilters = getOrganizationsFilters(filterChoices, texts);
+    const membersFilters = getMembersFilters(filterChoices, texts);
+    return [
+      ...projectsFilters,
+      ...organizationsFilters.filter((o) => !projectsFilters.map((p) => p.key).includes(o.key)),
```
piperchester, line 29: Why do we need an additional `.filter` and `.map` here? Do you mind adding a one-line comment for future us?

frontend/pages/hubs/[hubUrl].js
```diff
@@ -291,22 +222,22 @@ export default function Hub({
...
+            filterChoices={filterChoices}
+            handleSetErrorMessage={handleSetErrorMessage}
+            hideMembers={!isLocationHub}
+            hubName={name}
             hubProjectsButtonRef={hubProjectsButtonRef}
+            hubQuickInfoRef={hubQuickInfoRef}
+            initialLocationFilter={initialLocationFilter}
+            // TODO: is this still needed?
+            // initialOrganizations={initialOrganizations}
+            // initialProjects={initialProjects}
```
Comment (piperchester) [inline frontend/pages/hubs/[hubUrl].js:239]: ☝️ are these still needed?

frontend/pages/hubs/[hubUrl].js
```diff
@@ -167,100 +146,52 @@ export default function Hub({
...
   };
 
+  const handleUpdateFilterValues = (valuesToUpdate) => {
+    setFilters({
+      ...filters,
+      ...valuesToUpdate,
+    });
+  };
+
+  console.log(filters);
```
Comment (piperchester) [inline frontend/pages/hubs/[hubUrl].js:193]:
```suggestion
```

piperchester (commented):
Starting to read through more and leaving a first pass of comments. Looking really good! I think we need to fix some handling on the location filter -- I can't really type anything after checking this morning:


https://user-images.githubusercontent.com/1530684/126663288-6778c5dd-a9ee-4175-b7bb-3def0a7ce463.mov

backend/climateconnect_api/utility/translation.py
```diff
@@ -19,7 +20,7 @@ def get_locale(language_code):
         "DE": "de"
     }
     if(language_code not in LANGUAGE_CODE_MAP):
-       print('Unsupported language: ' + language_code)
+       logging.error('Unsupported language: ' + language_code)
```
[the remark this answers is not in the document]
positiveimpact: logging.error specifically shows in the 'error' logs while print shows in the 'info' logs. The logging stream in Azure only shows the error logs by default afaik.

frontend/test/components/filter/FilterContent.test.js
```diff
@@ -0,0 +1,60 @@
+import { findAllItems } from "../../../src/components/filter/FilterContent";
+
+describe("FilterContent", () => {
```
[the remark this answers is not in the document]
positiveimpact: Nope, I'll work on making them work again :)

piperchester:
Seeing that we finally landed this absolute gargantuan beast of a PR @positiveimpact ❤️ 💯 -- amazing work!

Issue #453, opened 2021-01-25
My Projects button doesn't scroll down to projects


tobiasrehm:
"My Projects" button in personal header sub-menu just redirects to profile page without scrolling down to projects

mdPial:
Hello, I would like to work on that issue 😃️

positiveimpact:
@mdPial Great! I assigned it to you. Let me know if you have any questions or need any help setting up the project locally.

positiveimpact:
You will need it to test locally. If you believe you can fix this without trying it out locally you won't need it. But I'd highly recommend it so you can make sure it actually works.
Since this is quite a simple change it could be possible without setting up the backend though.

mdPial:
When I am running the front-end I am getting this error. I couldn't figure out.
![image](https://user-images.githubusercontent.com/43095489/105852137-4518e480-600e-11eb-81a2-9cf64ba33da4.png)

positiveimpact:
The frontend can not work on its own with the current setup. 
You'll need to set up the backend to resolve this error and to be able to test with the frontend

mdPial:
Okay. Thank you. Highly appreciated your quick response.

mdPial:
Could you provide guideline in setting those environment variable? I am confused about the value.
API_HOST
API_URL
BASE_URL_HOST
DONATION_CAMPAIGN_RUNNING
ENVIRONMENT
GOOGLE_ANALYTICS_CODE
LATEST_NEWSLETTER_LINK
SOCKET_URL
LETS_ENCRYPT_FILE_CONTENT

ddhanesha:
> Could you provide guideline in setting those environment variable? I am confused about the value.
> API_HOST
> API_URL
> BASE_URL_HOST
> DONATION_CAMPAIGN_RUNNING
> ENVIRONMENT
> GOOGLE_ANALYTICS_CODE
> LATEST_NEWSLETTER_LINK
> SOCKET_URL
> LETS_ENCRYPT_FILE_CONTENT

You **DO NOT** need these env variables...
1) LETS_ENCRYPT_FILE_CONTENT
2) DONATION_CAMPAIGN_RUNNING
3)GOOGLE_ANALYTICS_CODE
4) LATEST_NEWSLETTER_LINK

mdPial:
> > Could you provide guideline in setting those environment variable? I am confused about the value.
> > API_HOST
> > API_URL
> > BASE_URL_HOST
[...]

PR #457, opened 2021-01-28
Fix Scroll Issue Of My Projects Button

mdPial:
## Description

This solved the issue #453.
✨️**Reason**: The challenge is `components are different`. In the Previously implemented way, scroll doesn't work if calls from a different component.

I  was looking for the previous scroll implementation. I have found useRef is used in the codebase for scrolling. To maintain the previous architecture and simple implementation I have implemented useRef for scrolling directly in the [profileUrl].js file. Now the problem is we need to call the method for scrolling. So, I have used useEffect to call the method and check, Is the current URL for scroll down to My Projects or not. I have removed the id because it is no longer required.

## Todo


positiveimpact (approved):
LGTM! Great work.

mdPial:
Thank You😃

Issue #461, opened 2021-01-30
Smaller titles on project page in mobile


tobiasrehm:
Project Title is very big on mobile and should be smaller to make room for more content
![image](https://user-images.githubusercontent.com/64414429/106355072-19d81300-62f6-11eb-8a93-0c4d2fa59567.png)

PR #462, opened 2021-01-30
Fix title size on project page #461

marinazzz:
## Description
This solved the issue #461

Project title fixed by using viewport units and calc()


positiveimpact (changes requested):
Thanks for creating this PR!
The title on mobile is handled in the `SmallScreenOverview` component in `frontend/src/components/project/ProjectOverview.js`. 
It looks like this:
```
<Typography component="h1" variant="h3" className={classes.smallScreenHeader}>
  {project.name}
</Typography>
```
In material ui the `component` prop just decides which html component this element will have, but the `variant` prop decides which typography styling from the theme it will use. However we don't really use the typography stylings from our themes in many places. We usually overwrite them with custom css. The problem seems to be that the class `smallScreenHeader` isn't defined. I think the best way to fix this without impacting other components is adding this class with the expected fontSize to the ProjectOverview component.

As for package-lock.json: We use yarn instead of npm. The information from this file is in `yarn.lock` in the repo. Therefore it would be great if you remove this from the diff.

frontend/src/components/project/ProjectOverview.js
```diff
@@ -49,6 +49,9 @@ const useStyles = makeStyles((theme) => ({
     display: "flex",
     marginTop: theme.spacing(3),
     justifyContent: "space-between"
+  },
+  smallScreenHeader {
+    fontSize: "calc(1.6rem + 6 * ((100vw - 320px) / 680))"
```
positiveimpact, line 54: You'll need a ":" for this to work:
```suggestion
  smallScreenHeader: {
    fontSize: "calc(1.6rem + 6 * ((100vw - 320px) / 680))"
```

frontend/src/themes/theme.js
```diff
@@ -19,7 +19,7 @@ const coreTheme = createMuiTheme({
   typography: {
     fontFamily: "Open Sans",
     h1: {
-      fontSize: "3rem",
+      fontSize: "calc(1.6rem + 6 * ((100vw - 320px) / 680))",
```
positiveimpact, line 22: Please revert this change as it might have unintended side-effects on other parts of the page (i.e. h1s being displayed differently than expected)

positiveimpact (changes requested):
Good job! I suggested 2 more small changes, after that this PR will be ready to merge.

positiveimpact (approved):
LGTM!

PR #547, opened 2021-03-19
Merge python celery server setup

ddhanesha:
## Description

Added workflows for test and production environment. 

## Todo

- [ ] Lint passes?


positiveimpact (commented):
Seems like our celery App Services are for some reason failing during startup.
Since this doesn't include any breaking changes for the main codebase we can still merge but we'll need to investigate that.

ddhanesha:
> Seems like our celery App Services are for some reason failing during startup.
> Since this doesn't include any breaking changes for the main codebase we can still merge but we'll need to investigate that.

@positiveimpact that is because the startup script is not merged to master yet. That is why its failing in production. If you check our test servers it should be running celery without any issues.

ddhanesha:
Going to close this PR for now. I will create a new one once issue is resolved.

PR #569, opened 2021-04-14
WIP: Implement "translate project" step for share project, start working on backend

### LINKED ISSUE #560: Allow translating project in "share project" workflow
Sub-issue of #527 

All projects should be available in English and German.

**Frontend**
An additional step for translating the project should be added at the end of the "Share project" process according to [this design](https://xd.adobe.com/view/263c33d4-4ee4-441b-7c68-f4533c94f5ee-500d/screen/87bf4cec-433c-4528-97b0-8984d50b31f7/)
- Clicking on "Automatically translate text" should fetch an automated translation from the DeepL API and display it in the corresponding fields. Users should then still be able to edit those translations. If you have already filled out one or more of the translation fields, you should be able to select in a popup "overwrite my translations", "Just translated empty fields automatically" or "cancel"
- When clicking "Skip" (could rename this to "finish"), all empty fields will be marked to automatically be translated and then be translated on the server side.

**Backend**
- Create an endpoint for just translating the contents of a project. This endpoint should be called when clicking "automatically translate text"
- Add logic for automatically translating text fields for which the user didn't pass a translation
- Insert translated version of the project into the corresponding translation table

### LINKED ISSUE #561: Create a generic page for editing translations of profile, organization, and project texts
Sub-issue of #527

### LINKED ISSUE #565: Allow translating organizations in "create organization" workflow
Sub-issue of #527
Detailed concept still being created...

### LINKED ISSUE #566: Allow translating profile in "create profile" workflow
Sub-issue of #527
Detailed concept still being created...

positiveimpact:
Closes #560 Closes #566 Closes #565 Closes #561
Adds a "translate project" step to "share project".

## Todo

- [ ] Lint passes?


backend/organization/utility/project.py
```diff
@@ -39,6 +40,7 @@ def create_new_project(data: Dict) -> Project:
         project.is_draft = data['is_draft']
     if 'website' in data:
         project.website = data['website']
+    project.language = source_language
```
ddhanesha, line 43: Hey so language is a foreign key in project table. So I think you need to first get the language from `source_language` string and then set `project.language = language`

backend/organization/utility/project.py
```diff
@@ -83,3 +85,18 @@ def get_project_description(project: Project, language_code: str) -> str:
         ).description_translation
     
     return project.description
+
+def get_project_translations(request):
```
ddhanesha, line 89: I would recommend that you don't pass the request object here but a data object. Its best practice to not pass request object but pass references/values you want to use.

backend/organization/views/project_views.py
```diff
@@ -173,9 +175,31 @@ def post(self, request):
...
+            }, status=status.HTTP_400_BAD_REQUEST)  
+
+        project = create_new_project(request.data, translations['source_language'])
+
+        for language in translations:
+            if not language == translations['source_language']:
+                texts = translations[language]
+                translation = ProjectTranslation.objects.create(
+                    project=project, 
+                    language=language,
```
ddhanesha, line 194: Same here, language is a foreign key so you have first get the language object before you create project translation.

backend/organization/views/project_views.py
```diff
@@ -173,9 +175,31 @@ def post(self, request):
...
-            })
+            })  
 
-        project = create_new_project(request.data)
+        try:
+            translations = get_project_translations(request)            
+        except ValueError:
+            return Response({
+                'message': "Please use the same language for all texts."
+            }, status=status.HTTP_400_BAD_REQUEST)
```
ddhanesha, line 185: I wouldn't return 400 error here if translation fails here. But log the error maybe somehow notify us to see what went wrong. The goal here is to allow users to create projects. Translation should not be a blocker on creating projects.

backend/organization/utility/project.py
positiveimpact: Good point!

backend/organization/views/project_views.py
positiveimpact: Good point!

positiveimpact (commented):
@ddhanesha agreed to all your points and integrated them!

PR #573, opened 2021-04-19
bug fix : search users by skills

FnayouSeif:
## Return users matching the filtered skills instead of all users

the endpoint /api/members returns 500 when filtered by skills. This is due to the fact that django does not support filter + distinct(field_name). we can just use distinct()

## Todo

- [ ] Lint passes?


ddhanesha (approved):
Good catch! Approved

positiveimpact (approved):
LGTM! Great find! :)
Feel free to merge!

Issue #584, opened 2021-04-27
Create idea board to ease people into sharing ideas and projects


positiveimpact:
Ideaboard: https://xd.adobe.com/view/55e3ed3e-15a2-414e-911a-61b97a44bae9-00b0/screen/54bdc188-086c-435c-ad2b-6c4d11b2236b

https://xd.adobe.com/view/263c33d4-4ee4-441b-7c68-f4533c94f5ee-500d/screen/8b4c183e-c3b5-4764-aed6-526773ad8f5a

ddhanesha:
I will be creating more tickets to split this work into multiple smaller tasks.

ddhanesha:
@tobiasrehm @positiveimpact I am working on database design right now. I had a question about the badges on these board. Are they fixed badges for each idea or will it change for different ideas? 

<img width="146" alt="Screen Shot 2021-04-29 at 9 47 11 PM" src="https://user-images.githubusercontent.com/15343711/116649578-774f2080-a934-11eb-8306-3a8dae325cc8.png">

tobiasrehm:
The badge symbolises the Hub or category that the idea fits in. So it can
be changed by the user when submitting or editing an idea.

Am Fr., 30. Apr. 2021 um 06:48 Uhr schrieb Dip Dhanesha <
***@***.***>:

PR #599, opened 2021-05-04
Implementation of ideasboard

ddhanesha:
## Description

Resloves  #584 

## Todo

- [x] Lint passes?
- [x] Ran `yarn format` in frontend?


positiveimpact:
I think I would just make one PR for the whole ideas board. So I'l also work on this branch if you're fine with it :)

ddhanesha:
> I think I would just make one PR for the whole ideas board. So I'l also work on this branch if you're fine with it :)

Yeah that’s fine.

backend/ideas/models/ideas.py
```diff
@@ -0,0 +1,105 @@
...
+        max_length=256
+    )
+
+    summary = models.CharField(
+        help_text="Summary of an idea",
+        verbose_name="Summary",
+        max_length=1024
+    )
+
+    description = models.TextField(
```
positiveimpact, line 25: I think we'll just have a short description? but not sure if that is 100% decided yet

backend/ideas/models/translation.py
```diff
@@ -0,0 +1,67 @@
...
+
+    language = models.ForeignKey(
+        Language,
+        help_text="Points to the language idea is translated for",
+        verbose_name="Language",
+        related_name="idea_language",
+        on_delete=models.CASCADE
+    )
+
+    name = models.CharField(
```
positiveimpact, line 24: I would keep the convention of having `[fieldName]_translation` here, e.g. `name_translation` in this case

positiveimpact (commented):
LGTM, added 2 minor comments. Will also start committing to this PR

backend/ideas/models/translation.py
ddhanesha: Done

backend/ideas/models/ideas.py
ddhanesha: Removed

PR #602, opened 2021-05-07
fixed dev setup bugs

FnayouSeif:
## Description

- fixed the project status error when projects have a `default` status 
- fixed the migration script to new languages when local db does not have the languages in language table


positiveimpact (approved):
LGTM!

PR #606, opened 2021-05-11
Bugfix/locationsearch

FnayouSeif:
Bugs: 

1 . When search input comes from browsing page, the former is not stored in the backend due to JSON deserialization error. 
2. When search input is a point instead of a relation, the polygon coordinates are not available. Search should use centre point instead


positiveimpact (approved):
Great work! LGTM and works smoothly locally! :ok_hand:

PR #627, opened 2021-06-10
Feature/joinprojectrequest

FnayouSeif:
## Todo


backend/organization/permissions.py
```diff
@@ -224,4 +224,15 @@ def has_permission(self, request, view):
...
+        return False
+
+
+class ApproveDenyProjectMemberRequest(BasePermission):
+    def has_permission(self, request, view):
+        if request.method in SAFE_METHODS:
+            return True
+        else:
+            project = Project.objects.filter(url_slug=str(view.kwargs.get('project_slug'))).first()
+            return ProjectMember.objects.filter(project=project,user=request.user, role__role_type__in=[Role.ALL_TYPE]).exists()
```
positiveimpact, line 236: Does this throw an error if the project doesn't exist?

backend/organization/utility/project.py
```diff
@@ -1,3 +1,12 @@
+<<<<<<< HEAD
```
positiveimpact, line 1: This seems to be a leftover from a merge (Just like the ======= below)

backend/organization/utility/project.py
```diff
@@ -9,6 +18,7 @@
 
 from organization.models import Project
 from organization.models.tags import ProjectTags
+>>>>>>> 8787e69f7524e93096eb6aa9f74f3881403998b0
```
positiveimpact, line 21: Leftover from a merge just like the ======= below

backend/organization/utility/project.py
```diff
@@ -60,6 +70,45 @@ def create_new_project(data: Dict, source_language: Language) -> Project:
     project.save()
     return project
 
+<<<<<<< HEAD
```
positiveimpact, line 73: Leftover from a merge just like the ======= below

backend/organization/utility/project.py
```diff
@@ -127,3 +176,4 @@ def get_project_translations(data: Dict):
         )
     except ValueError:
         raise ValueError
+>>>>>>> 8787e69f7524e93096eb6aa9f74f3881403998b0
```
positiveimpact, line 179: leftover from a merge

backend/organization/utility/project.py
```diff
@@ -60,6 +70,45 @@ def create_new_project(data: Dict, source_language: Language) -> Project:
...
+
+def get_project_admin_creators(project,limit_to_admins=False):
+    """
+    Returns a given project UserProfiles of Creators or Administrators. if limit_to_admins is set to True, only admins will be returned.
+    :param project: target project 
+    :type project: Project 
+    :param limit_to_admins: limit output to admins only 
+    :ype limit_to_amins: bool
+    """
+    targets_roles = Role.objects.filter(Q(name="Creator") | Q(name="Administrator")).all()
```
positiveimpact, line 99: I wouldn't go by name here. You can check whether `role_type__in=[Role.READ_WRITE_TYPE, Role.ALL_TYPE]`. 
Relying on name for roles can lead to some unexpected results :D

backend/organization/utility/project.py
```diff
@@ -60,6 +70,45 @@ def create_new_project(data: Dict, source_language: Language) -> Project:
...
+    if targets_roles.count() < 1: raise Exception(f"Project does not have any Admins! {targets_roles}")  
+    admin_role, creator_role = targets_roles.filter(name="Administrator").first(), targets_roles.filter(name="Creator").first()
```
positiveimpact, line 101: Same thing here, wouldn't go by creator

backend/organization/utility/project.py
```diff
@@ -60,6 +70,45 @@ def create_new_project(data: Dict, source_language: Language) -> Project:
...
+def get_project_admin_creators(project,limit_to_admins=False):
+    """
+    Returns a given project UserProfiles of Creators or Administrators. if limit_to_admins is set to True, only admins will be returned.
+    :param project: target project 
+    :type project: Project 
+    :param limit_to_admins: limit output to admins only 
+    :ype limit_to_amins: bool
+    """
+    targets_roles = Role.objects.filter(Q(name="Creator") | Q(name="Administrator")).all()
+    if targets_roles.count() < 1: raise Exception(f"Project does not have any Admins! {targets_roles}")
```
positiveimpact, line 100: This actually gets thrown if there is no creator or administrator role in the database, not if the project has no admins (If I interpret it correctly)

backend/organization/utility/requests.py
```diff
@@ -0,0 +1,156 @@
...
+    B. To manage an existing request, provide the request id only as a kwargs. If project or organization are added, then the class check for the consistency of the supplied id 
+    and the corresponding project/ organization. If validation fails, property validation_failed will be True and errors property will contain a list of errors.
+    """
+    def __init__(self,**kwargs):
+        
+        self.validation_failed = False
+        self.duplicate_request = False 
+        
+        self.errors = list()
+        if 'membership_request_id' not in kwargs.keys(): #new request
```
positiveimpact, line 37: Since the code executed in the if and the else statement is completely different I think it would be better to just make 2 different functions for the 2 different use cases. But not super sure about this.

backend/organization/utility/requests.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+    def approve_request(self,**kwargs):
+
+
+        self.membership_request.request_status = RequestStatus.APPROVED.value 
+        self.membership_request.approved_at = timezone.now()
+
+        #create a record in ProjectMember
+        user_role = Role.objects.filter(name="Member").first()
```
positiveimpact, line 123: Here I'd also filter by role_type=Role.READ_ONLY_TYPE instead of filtering by name.

backend/organization/views/project_views.py
```diff
@@ -60,6 +60,40 @@
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
+<<<<<<< HEAD
```
positiveimpact, line 63: merge leftover

backend/organization/views/project_views.py
```diff
@@ -60,6 +60,40 @@
...
+)
+from rest_framework.exceptions import ValidationError, NotFound
+from climateconnect_main.utility.general import get_image_from_data_url
+from climateconnect_api.models import Role, Skill, Availability, UserProfile
+from ..utility.requests import MembershipRequestsManager
+from ..utility import MembershipTarget
+from django.db.models import Q
+import logging
+=======
+>>>>>>> 8787e69f7524e93096eb6aa9f74f3881403998b0
```
positiveimpact, line 96:
```suggestion

```

backend/organization/views/project_views.py
```diff
@@ -792,3 +826,96 @@ def post(self, request, url_slug):
...
+            elif request_manager.validation_failed:
+                #request_manager.errors
+                return Response({'message': 'Operation Failed.'}, status=status.HTTP_400_BAD_REQUEST)
+
+            if request_action == 'approve':
+                request_manager.approve_request()
+                create_project_join_request_approval_notification(requester=request.user,project=project)
+
+            elif request_action == 'reject':
+                request_manager.reject_request()
```
positiveimpact, line 912: I think it could also be nice to notify the user when they have been rejected

positiveimpact (commented):
Great work overall! I had a few questions and nits but overall the code looks really good!
One general question: How does TableName.objects.filter(<query>).first() behave if there is no results? So far we always used TableName.objects.get(<query) with a try/catch wrapped around it
Also one TODO: When we send notifications we should also send an email notification to the user.

PR #645, opened 2021-06-21
Add ability to @mention other users.

samschaevitz:
## Description

This feature so far:
- queries climateconnect users while you type in the comment box when you start a word with '@'
- saves the mentioned users as linkable html (unsafe!)

It will:
- notify users when they are tagged on comment save

## Todo

DO_NOT_SUBMIT: this implementation is unsafe -- users can inject whatever they want into the comment link contents. Transform the markdown server-side instead.


samschaevitz:
I'm looking for a first pass on this implementation. 

I still need to a) wire up the notification call and b) not-dangerously create links in the client to be retrieved & executable as HTML for future clients :grimacing:

backend/climateconnect_api/utility/common.py
```diff
@@ -1,14 +1,16 @@
...
 
 def get_role_name(role: Role, language_code: str) -> str:
-    if language_code == "en":
+    a = getattr(role, "name_{}_translation".format(language_code))
+    if language_code == "en" or a == None:
         return role.name
-    else:
-        return getattr(role, "name_{}_translation".format(language_code))
\ No newline at end of file
+    return a
```
positiveimpact: This change makes a lot of sense! However I think for unrelated changes it could make sense to quickly create an additional PR which we can then instantly merge. This way small fixes aren't blocked by larger features

positiveimpact: Nevermind. This doesn't work for english because there is no name_en_translation attribute. Therefore we'll need to go back to the old code or find a different solution

positiveimpact (commented):
Great job so far, it's working for me locally :) 
There's a few things that still need to be polished obviously but I didn't add any comments about those because it's still a draft.
I have a few questions and points:
- We're always using develop as the base branch for PRs now. This way we can test new features on the testserver
- I'm a bit confused: Didn't we already merge the "getting the token in the frontend" stuff?
Generally there are some weird diffs that I thought wouldn't be diffs. Maybe that's because you didn't merge the newest version of develop/master into your branch?
- Which library do you use for formatting?

samschaevitz:
Lint is done! Ready for review. 

Be wary of bad merge outcomes..I did my best and it runs, but I may have missed something weird.

PR #658, opened 2021-07-08
fix: misc polish of wording; typo fixes

piperchester:
Couple very small pieces I noticed just navigating around.

### DIFF
--- a/frontend/public/texts/profile_texts.js
+++ b/frontend/public/texts/profile_texts.js
@@ -108,16 +108,16 @@ export default function getProfileTexts({ profile, locale }) {
       de: ", um die vollständigen Informationen dieses/dieser Benutzer*in zu sehen",
     },
     your_projects: {
-      en: "Your projects",
-      de: "Deine Projekte",
+      en: "Your climate projects",
+      de: "Deine Klimaschutzprojekte",
     },
     this_users_projects: {
       en: `${profile?.first_name}'s Climate Projects`,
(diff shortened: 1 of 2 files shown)


frontend/public/texts/profile_texts.js
```diff
@@ -108,16 +108,16 @@ export default function getProfileTexts({ profile, locale }) {
...
     },
     this_users_projects: {
       en: `${profile?.first_name}'s Climate Projects`,
       de: `Klimaschutzprojekte von ${profile?.first_name}`,
     },
     your_ideas: {
-      en: "Your Climate Ideas",
-      de: "Deine Klimaschutzideen",
+      en: "Your ideas",
+      de: "Deine Organisationen",
```
piperchester, line 120: Made this consistent wording across projects, orgs, and ideas:

Screen Shot 2021-07-07 at 7 27 16 PM

Not sure if I got the 🇩🇪 right!

positiveimpact: Actually I changed that to "climate projects" and "climate ideas" on purpose so that people who find a profile through Google can see that the platform is about climate projects and climate ideas. The reason why I didn't change it for organizations was because not every organization that has a climate projects would be considered a climate organization, especially for-profit companies.

piperchester: Ah gotcha - ok will back that out 👍

ddhanesha (approved):

PR #669, opened 2021-07-24
Joining a project

FnayouSeif:
## Description

This PR enables the workflow of requesting to join a project and managing these requests. 

**Concept**
It is important to know the workflow: 

1. A user asks to join a project 
2. The request will be persisted in the corresponding membershiprequests table 
3. The project admins should be able to see(1) the requests to join their projects on an interface 
4. The project admins can either approve or deny the request(2) through a button

(1) This will be defined with @piperchester. I will probably need to add a new endpoint that returns the set of user requests. The latter can be seen on the project members tab with a different coloring/icons or on a whole different form/page. 
(2): This is also to be defined with our dear @piperchester and it depends on (1)  

API definition: 

**1. Requesting to join a project** 
endpoint: `projects/<str:project_slug>/request_membership/<str:user_slug>/`
Auth required: Yes 
A. url params: 
1. `project_slug`: The project slug that the user wants to become of member of
2. `user_slug`: The requesting user slug 

B. body params:
1. `message`: The message the user can add with his membership request
2. `user_availability`: the user availability id  

If successful, a 200 response will be returned along with a body containing request id (not useful for now). 

**2. Approving/ Rejecting a membership request** 
endpoint: `projects/<str:project_slug>/request_membership/<str:request_action>/<str:request_id>/`
Auth required: Yes (Special permission in class : `ApproveDenyProjectMemberRequest`) 
url params: 
1. `project_slug`: same as above 
2. `request_action`: indicates the action to do for the membership request (`approve` or `reject`)
3. `request_id`: the request id value

if successful, a 200 response will be returned. Otherwise, an error will be returned along with a message


backend/organization/models/members.py
```diff
@@ -114,3 +114,68 @@ class Meta:
...
+
+    user = models.ForeignKey(
+        User,
+        verbose_name="feedback_user",
+        help_text="Points to the user who sent the request",
+        on_delete=models.CASCADE,
+        null=False
+    )
+
+    target_membership_type = models.SmallIntegerField(choices=[(x.value,x.name) for x in MembershipTarget]
```
ddhanesha, line 133: Nit: I think you are applying SQL style here. In python I think this is a recommended format. 
```
target_membership_type = models.SmallIntegerField(
    choices=[(x.value, x.name) for x in MembershipTarget],
    null=False,
    verbose_name="Membership Request Target Type",
    help_text="The type of entity a user requested to join (organization, project...)"
)
```

backend/organization/models/members.py
```diff
@@ -114,3 +114,68 @@ class Meta:
...
+                                                     ,null=False
+                                                     ,verbose_name="Membership Request Target Type",
+                                                      help_text="The type of entity a user requested to join (organization, project...)")
+
+    target_project =  models.ForeignKey(Project
+                                            ,null=True
+                                            ,on_delete=models.CASCADE
+                                            ,help_text="Points to the requested project")
+
```
ddhanesha, line 142: You are missing `verbose_name` here. Django doesn't require it but to be consistent with the rest of the code I would recommend adding them. 

Same for `related_name` on `ForeginKey` relationships. If you don't define related_name django automatically handles it but for consistency purposes I guess doesn't hurt to add a related_name too. You can look examples for Idea or Project tables to define custom `related_names`

piperchester:
Merged into #674 🎉

PR #670, opened 2021-07-24
Python Linter

FnayouSeif:
## Description
Added Backend python linter and applied few linting fixes (mainly unused assigned variables)
Errors in scope: ([Source I](https://flake8.pycqa.org/en/latest/user/error-codes.html); [Source II](https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes))

<html>
<body>

F811 | redefinition of unused name from line N
-- | --
F821 | undefined name name
F822 | undefined name name in __all__
F823 | local variable name … referenced before assignment
F831 | duplicate argument name in function definition
F841 | local variable name is assigned to but never used

</body>
</html>

<html>
<body>

E9 | Runtime
-- | --
E901 | SyntaxError or IndentationError
E902 | IOError

</body>
</html>

<html>
<body>

F601 | dictionary key name repeated with different values
-- | --
F602 | dictionary key variable name repeated with different values
F621 | too many expressions in an assignment with star-unpacking
F622 | two or more starred expressions in an assignment (a, *b, *c = d)
F631 | assertion test is a tuple, which is always True
F632 | use ==/!= to compare str, bytes, and int literals
F633 | use of >> is invalid with print function
F634 | if test is a tuple, which is always True

</body>
</html>

<html>
<body>

F701 | a break statement outside of a while or for loop
-- | --
F702 | a continue statement outside of a while or for loop
F703 | a continue statement in a finally block in a loop
F704 | a yield or yield from statement outside of a function
F705 | a return statement with arguments inside a generator
F706 | a return statement outside of a function/method
F707 | an except: block as not the last exception handler
F721 | syntax error in doctest
F722 | syntax error in forward annotation
F723 | syntax error in type comment

</body>
</html>

<html>
<body>

F811 | redefinition of unused name from line N
-- | --
F821 | undefined name name
F822 | undefined name name in __all__
F823 | local variable name … referenced before assignment
F831 | duplicate argument name in function definition
F841 | local variable name is assigned to but never used

</body>
</html>

## Todo


backend/chat_messages/models/message.py
```diff
@@ -58,7 +58,7 @@ class Meta:
         verbose_name_plural = "Chats"
         ordering = ["-last_message_at", "-created_at"]
 
-    def __str__(self):
+    def __DEPRECEATED_str__(self):
```
FnayouSeif, line 61: @positiveimpact 
Please check your changes here. A new __str__ was added while leaving the other one. The class had 2 __str__ methods.

backend/ideas/utility/idea.py
```diff
@@ -2,7 +2,7 @@
 import urllib.parse
 from typing import Dict, Optional
 
-from climateconnect_api.models import language
+#from climateconnect_api.models import language
```
ddhanesha, line 5: Why not just remove this?

backend/climateconnect_api/views/user_views.py
```diff
@@ -1,6 +1,7 @@
-import datetime
+#import datetime
 from ideas.serializers.idea import IdeaFromIdeaSupporterSerializer
 from ideas.models.support import IdeaSupporter
+#import datetime
```
ddhanesha, line 4: Remove unused imports.

FnayouSeif: Done

backend/ideas/utility/idea.py
FnayouSeif: Done

ddhanesha (approved):
Comments addressed. Good job @FnayouSeif adding python linter. 
Approved!

piperchester:
If no objections, will land this given @ddhanesha's approval so we can start taking advantage of it. Thanks for adding this Seif!

backend/chat_messages/models/message.py
positiveimpact: Thanks! Removed the old str method.

Issue #596, opened 2021-05-01
Implement lint GitHub action to run in CI


piperchester:
We should be linting (and also running unit tests) on every pull request IMO. 

This was started in #456.

This should likely leverage https://github.com/marketplace/actions/lint-action#javascript-example-eslint-and-prettier.

piperchester:
See #628!

Issue #672, opened 2021-07-24
Implement flow for project admins to approve / deny project requests


piperchester:
This relates to #669.

Issue #671, opened 2021-07-24
Implement UI for project admins to see project requests


piperchester:
Relates to https://github.com/climateconnect/climateconnect/pull/669

PR #674, opened 2021-07-24
feat: implement project requests, approve, and deny flow 

piperchester:
This implements #671, #672 and includes the PR from #669.

Also 
- Fixes a couple bugs and adds more conditions
- Fixes the DEEPL_API_KEY assignment
- Misc. readability improvements
- Fixes https://github.com/climateconnect/climateconnect/issues/596


frontend/src/components/project/ProjectOverview.js
```diff
@@ -250,17 +254,68 @@ function SmallScreenOverview({
...
+          // TODO: fix payload
+          message: "Hi!",
+          // user_availability: "3",
+          user_availability: "4",
+        },
+
+        headers: {
+          // TODO: is this right?
+          Authorization: "Token 5b25c244d9afe56b939eb07049f0eee641ea70bd81273518a1e83f77504149f4",
+        },
```
piperchester, line 296: Definitely need insight into how we should correctly provide the token here 👆

frontend/TESTING.md
```diff
@@ -10,6 +10,17 @@ I'm aiming to keep the ROI high with this, and just outlining some basic functio
...
+
+<!-- Note: this documentation might move elsewhere as the project evolves. -->
+
+### API calls
+
+### Tips
+
+1. Don't forget the trailing slash!
+1. We resuse Postman collections and snippets -- see ...
+
```
piperchester, line 23: @piperchester to finish this / move elsewhere.

frontend/src/components/communication/CommentInput.js
```diff
@@ -85,7 +85,9 @@ export default function CommentInput({
               className={classes.commentButton}
               onClick={(event) => handleSendComment(event)}
             >
-              {texts.send}
+              {/* TODO(piper): should update to "post" comment instead of send, which implies chat */}
+              {/* {texts.send}  */}
+              Post
```
piperchester, line 90: @ reviewers -- make sense?

frontend/src/components/dialogs/ProjectRequestersDialog.js
```diff
@@ -0,0 +1,178 @@
...
+  };
+
+  return (
+    <GenericDialog
+      onClose={handleClose}
+      open={open}
+      // TODO(Piper): iterate on naming here
+      // title={texts.requesters_of + " " + project.name}
+      title={"Requested to join project"}
+      //   title={"Users who have requested to follow"}
```
piperchester, line 76: What title makes the most sense here?

frontend/src/components/dialogs/ProjectRequestersDialog.js
```diff
@@ -0,0 +1,178 @@
...
+  // TODO(piper): make appropriate API calls here to approve/reject
+  const handleApproveRequester = () => {
+    console.log("Approved!");
+    onClose();
+  };
+
+  const handleDenyRequester = () => {
+    console.log("Denied!");
+    onClose();
+  };
```
piperchester, line 125: @piperchester entry points for approving / rejecting requesters.

frontend/src/components/project/ProjectOverview.js
```diff
@@ -250,17 +254,58 @@ function SmallScreenOverview({
...
+  /**
+   * Calls backend, sending a request to join this project
+   */
+  async function handleSendProjectJoinRequest() {
+    // TODO(piper): fix user name
+    // TODO(piper): fix project name
+    const projectName = "Anotherproject6";
+    // const userName = "testchester";
+    // const userName = "pip.erchester@gmail.com";
+    const userName = "testchester9";
```
piperchester, line 280: @piperchester all of this needs to be cleaned up!

piperchester:
Going to open this up for an initial review! I still have some TODOs, but welcoming any feedback on the approach so far.

backend/organization/urls.py
```diff
@@ -144,4 +150,15 @@
         project_views.LeaveProject.as_view(),
         name='leave-project'
     )
+    ,
+    path(
+        'projects/<str:project_slug>/request_membership/<str:user_slug>/',
```
positiveimpact, line 155: This url is a bit weird to me. The User object is passed in an authenticated request anyways, so I don't see the need to pass the user_slug.

positiveimpact (commented):
Started a review here but didn't get it done. I'll just submit my unfinished review with some comments already.

Great progress overall but still quite some polishing to do. One important thing is to only show the buttons to review requests or to request to join when your permissions match those. I got shown the button to join even though I was already part of the project.
Also one thing that's still to do is to make sure that the design works on all screensizes, especially mobile

backend/organization/models/members.py
```diff
@@ -114,3 +114,76 @@ class Meta:
...
+                                        ,default=None 
+                                        ,null=True                                       
+                                        )
+    rejected_at = models.DateTimeField(help_text="Time of request rejection"
+                                        ,verbose_name="Rejected at"
+                                        ,default=None 
+                                        ,null=True  
+                                        )
+
+    message = models.CharField(help_text="Message associated with the request",
```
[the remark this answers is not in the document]
piperchester: Neat; this was useful for me: https://stackoverflow.com/questions/7354588/django-charfield-vs-textfield -- apparently there isn't a performance difference or difference in practice in Postgres, but it seems like a better pattern regardless based on some of the other attributes.

piperchester:
Drafting for WIP.

backend/organization/urls.py
piperchester: Ok good to know. Will update and update call to the endpoint!

.github/workflows/lint-frontend.yml
```diff
@@ -0,0 +1,28 @@
...
+
+        # ESLint and Prettier must be in `package.json`
+        - name: Install Node.js dependencies
+          run: yarn
+
+        - name: Lint frontend
+          uses: yarn lint
+          with:
+            eslint: true
+            prettier: true
```
ddhanesha, line 28: 👍🏽 for this

backend/climateconnect_api/models/notification.py
```diff
@@ -31,9 +33,11 @@ class Notification(models.Model):
         (POST_COMMENT, "post_comment"),
         (REPLY_TO_POST_COMMENT, "reply_to_post_comment"),
         (GROUP_MESSAGE, "group_message"),
+        (JOIN_PROJECT_REQUEST,"join_project_request"),
+        (PROJECT_JOIN_REQUEST_APPROVED,"project_join_request_approved"),
         (IDEA_COMMENT, "idea_comment"),
         (REPLY_TO_IDEA_COMMENT, "reply_to_idea_comment"),
-        (PERSON_JOINED_IDEA, "person_joined_idea")
+        (PERSON_JOINED_IDEA, "person_joined_idea"),
     )
```
ddhanesha, line 41: Non-Blocker and Note to engineers: I think its time we move notification_type to its on database table. I think this will be expanded to too many different types so better to have them in a database table then hard coded in the codebase.

backend/organization/models/members.py
```diff
@@ -114,3 +114,76 @@ class Meta:
 
     def __str__(self):
         return "User %d member for organization %s" % (self.user.id, self.organization.name)
+
+class MembershipRequests(models.Model):
+
+    id = models.AutoField(help_text='identifier of a join request'
+                                , primary_key=True
+                                ,verbose_name="join request Id")
```
ddhanesha, line 122: Non-blocker: Format this like `user` column on line 125

ddhanesha: You don't have to identify `id` since django automatically identify this.

backend/organization/models/members.py
```diff
@@ -114,3 +114,76 @@ class Meta:
...
+
+
+    user = models.ForeignKey(
+        User,
+        verbose_name="feedback_user",
+        help_text="Points to the user who sent the request",
+        on_delete=models.CASCADE,
+        null=False,
+        related_name='users'
```
ddhanesha, line 131: `related_name` is too broad here. You can either not set a related_name (let django handle it) or set something like `user_membership_request`. A bit more self explanatory related  name.

backend/organization/models/members.py
```diff
@@ -114,3 +114,76 @@ class Meta:
...
+                                                     ,null=False
+                                                     ,verbose_name="Membership Request Target Type",
+                                                      help_text="The type of entity a user requested to join (organization, project...)")
+
+    target_project =  models.ForeignKey(Project
+                                            ,null=True
+                                            ,on_delete=models.CASCADE
+                                            ,help_text="Points to the requested project"
+                                            ,verbose_name="Target Project of Membership Request"
+                                            ,related_name='projects')
```
ddhanesha, line 144: Same here about `related_names`

backend/organization/models/members.py
```diff
@@ -114,3 +114,76 @@ class Meta:
...
+
+    target_organization =  models.ForeignKey(Organization
+                                            ,null=True
+                                            ,on_delete=models.CASCADE
+                                            ,help_text="Points to the requested organization"
+                                            ,related_name='organizations'
```
ddhanesha, line 150: Same for organization.

backend/organization/models/members.py
```diff
@@ -114,3 +114,76 @@ class Meta:
...
+                                            ,verbose_name='Target Ogranization of Membership request')
+
+    availability = models.ForeignKey(Availability
+                                     ,null=False
+                                     ,on_delete=models.CASCADE
+                                     ,help_text="Points to the Availability offered by the user"
+                                     ,verbose_name='requested membership availability'
+                                     ,related_name='availabilities'
```
ddhanesha, line 158: same for availability.

backend/organization/permissions.py
```diff
@@ -224,4 +224,12 @@ def has_permission(self, request, view):
...
\ No newline at end of file
+        return False
+
+class ApproveDenyProjectMemberRequest(BasePermission):
+    def has_permission(self, request, view):
+        if request.method in SAFE_METHODS:
+            return True
+        else:
+            project = Project.objects.filter(url_slug=str(view.kwargs.get('project_slug'))).first()
+            return ProjectMember.objects.filter(project=project,user=request.user, role__role_type__in=[Role.ALL_TYPE]).exists()
```
ddhanesha, line 235: Shouldn't the role type only admins to approve or deny project memebership?

backend/organization/urls.py
ddhanesha: I don't a problem with this URL. This specifically identifies that `user_id 1` has requested a member ship for `project_id 100`. More clearer. Its not about authentication but how the URL should be formed.

backend/organization/utility/__init__.py
```diff
@@ -0,0 +1,10 @@
+from enum import Enum 
+
+class RequestStatus(Enum): 
+   PENDING =  1
+   APPROVED = 2
+   REJECTED = 3
```
ddhanesha, line 6: Non-blocker: This is a good approach, but I think the better approach would be using a state machine. So in future if we have different user behavior around status we could easily manage it. https://github.com/viewflow/django-fsm

Of course we can work on this as a separate ticket later. I will leave up to authors of this PR.

backend/organization/utility/notification.py
```diff
@@ -43,4 +43,41 @@ def create_project_follower_notification(project_follower):
...
+    Creates a notification about a joining request from a requester to a project admin. 
+    :param requester: UserProfile object of the user who's sent the request 
+    :type requester: UserProfile
+    :param project_admin: Iterable UserProfile object of the project administrators
+    :type project_admin: List(UserProfile)
+    
+    """
+    requester_name = UserProfile.objects.filter(user=requester).first().name
+    notification = Notification.objects.create(
+        notification_type=9,
```
ddhanesha, line 60: Rather than number I would prefer using `Notification.JOIN_PROJECT_REQUEST` here. Much better and cleaner to understand.

backend/organization/utility/notification.py
```diff
@@ -43,4 +43,41 @@ def create_project_follower_notification(project_follower):
...
+    Creates a notification about an approved request to join a project to the requester.
+    :param requester: UserProfile object of the user who's sent the request 
+    :type requester: UserProfile
+    :param project: Iterable UserProfile object of the project administrators
+    :type project: List(UserProfile)
+    
+    """
+
+    notification = Notification.objects.create(
+        notification_type=10,
```
ddhanesha, line 78: Same for this line.

backend/organization/utility/organization.py
```diff
@@ -78,3 +78,10 @@ def is_valid_organization_size(org_size) -> bool:
         if size_option[0] == org_size:
             return True
     return False
+
```
ddhanesha, line 81: Non-blocker and Linting: Two spaces between methods.

backend/organization/utility/project.py
```diff
@@ -127,3 +130,52 @@ def get_project_translations(data: Dict):
...
+    Adds a user to a project. Assumes valid data at input.
+    """
+
+    ProjectMember.objects.create(
+                        project=project
+                        ,user=user
+                        ,role=user_role
+                        ,role_in_project=role_in_project
+                        ,availability=availability
+                    )
```
ddhanesha, line 145: NB: This is a pythonic way of writing code. 
```
    ProjectMember.objects.create(
        project=project,
        user=user,
        role=user_role,
        role_in_project=role_in_project,
        availability=availability
    )
```

backend/organization/utility/project.py
```diff
@@ -127,3 +130,52 @@ def get_project_translations(data: Dict):
...
+def is_part_of_project(user,project):
+    """
+    Returns True if user belongs to a project
+    :param user: user to be checked
+    :type user: User
+    :param project: project to be checked
+    :type project: ProjectMember
+    """
+    n = ProjectMember.objects.filter(project=project, user=user).count()
+    if n ==0:
```
ddhanesha, line 178: NB: Can just do `if ProjectMember.objects.filter(project=project, user=user).count() == 0:`

backend/organization/utility/requests.py
```diff
@@ -0,0 +1,149 @@
...
+
+from climateconnect_api.models import Role
+from organization.utility.project import add_project_member,is_part_of_project
+from organization.utility.organization import add_organization_member
+from organization.models.members import MembershipRequests
+from organization.utility import MembershipTarget, RequestStatus
+import logging
+
+logger = logging.getLogger(__name__)
+
```
ddhanesha, line 16: NB: Two spaces here.

backend/organization/utility/requests.py
```diff
@@ -0,0 +1,149 @@
...
+                                          ,target_organization_id=self.organization).count()
+
+
+
+            if n > 0:
+                self.validation_failed = True
+                self.errors.append("Request Already Exists")
+                self.duplicate_request = True
+
+
```
ddhanesha, line 64: NB: Remove extraneous spaces

backend/organization/utility/requests.py
```diff
@@ -0,0 +1,149 @@
...
+            self.message = kwargs['message']
+            self.membership_request = None
+
+            # check if the request exists already
+            n = MembershipRequests.objects.filter(user=self.user
+                                          ,target_membership_type=self.membership_target.value
+                                          ,target_project_id=self.project
+                                          ,target_organization_id=self.organization).count()
+
+
```
ddhanesha, line 57: NB: Remove extraneous spaces

backend/organization/models/members.py
```diff
@@ -114,3 +114,76 @@ class Meta:
...
+        max_length=4096,
+        null=True,
+        blank=True)
+
+    request_status = models.SmallIntegerField(
+        choices=[(x.value,x.name) for x in RequestStatus],
+        default=RequestStatus.PENDING.value
+
+    )
+
```
ddhanesha, line 189: Based on the logic it looks like a member can only request a project once. Why don't we add index here to set `user` and `project` or `user` and `organization` are unique together. I see a lot of code checking for record == 0 or record != 1. 
Resource: https://docs.djangoproject.com/en/3.2/ref/models/options/#unique-together

ddhanesha: You can unique constraint under `class Meta`. Look other models in the code for reference.

frontend/src/components/dialogs/ProjectRequestersDialog.js
ddhanesha: `Members requested to join project` makes sense to me

ddhanesha (changes requested):
Half PR review comments.

piperchester:
Going to address some of @ddhanesha's feedback; drafting again.

backend/organization/models/members.py
ddhanesha: Yes to formatting. But `id` is auto generated field in django so you don't really need to identify it in the code.

piperchester: Handled in 5963944

backend/organization/utility/__init__.py
piperchester: Neat. I didn't even know about this feature in Django! Do you think the future refactor cost would outweigh the benefit? I don't have a strong POV on whether we should adopt this now or later. I'm inclined to not address, and consider FSM more going forward.

backend/organization/permissions.py
piperchester: That makes sense to me! I'll update to handle this change 👍

backend/organization/utility/organization.py
piperchester: 🙌 Does GitHub suggestion's feature allow you to insert newlines for things like this?

backend/organization/models/members.py
piperchester: For my own understanding, can I just update this line to `user_membership_request`? Or would I have to propagate that change elsewhere too?

backend/organization/permissions.py
piperchester: Ok coming back here - this might be misunderstanding the code, but I interpret this as that's actually the behavior, looking at: https://github.com/climateconnect/climateconnect/blob/master/backend/climateconnect_api/models/role.py#L41

backend/organization/urls.py
ddhanesha: We can fix this later since we have been working on this PR for a while.

backend/organization/utility/__init__.py
ddhanesha: On a second thought, I think enums are fine for this PR. We can talk about state machine but not necessary for this work

backend/organization/utility/notification.py
ddhanesha: This is also fine for now. No need to make code changes

backend/organization/utility/notification.py
ddhanesha: Same don't need to change right now.

backend/organization/utility/organization.py
ddhanesha: I will take a look at mypy for this automate.

positiveimpact (changes requested):
Great job handling this monster PR!
I made quite a lot of comments but most of them are really minor. 

A few things additional to my comments:
- The join button makes the ProjectOverview component look too cluttered now, I think it will be hard for users to choose what to click on. I'd move the join button below to the ProjectContent component in the same place where the "edit project" button is for logged in users.
- As a next step we'll definitely make the option to add a message 
- If I accept a user's join request and then remove that user from the team, the user won't be able to make another request.
- When approving a user I'm getting a 500 error but the user is still added to the team afterwards
- If my request was rejected and I'm making another request, I won't get any indication that my request hasn't been sent. Not sure if this is on purpose to prevent people from making spam requests on additional accounts?

frontend/src/components/project/ProjectCommentsContent.js
```diff
@@ -82,7 +84,7 @@ export default function CommentsContent({ user, project, token, setCurComments }
   return (
     <div>
       <CommentInput user={user} onSendComment={onSendComment} hasComments={comments.length > 0} />
-      <Typography>{comments.length + " " + texts.comments}</Typography>
+      <Typography>{comments.length + " " + texts.comments.toLowerCase()}</Typography>
```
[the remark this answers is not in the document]
piperchester: Neat! TIL, will restore 👍

backend/organization/views/project_views.py
```diff
@@ -885,6 +916,32 @@ def get_queryset(self):
...
+    serializer_class = ProjectRequesterSerializer
+
+    def get_queryset(self):
+        try:
+            project = Project.objects.get(url_slug=self.kwargs['url_slug'])
+        except Project.DoesNotExist:
+            return Response(data={'message': f'Project does not exist'}, status=status.HTTP_400_BAD_REQUEST)
+
+        all_requests = MembershipRequests.objects.all()
+        print(f"Total number of membership requests: {len(all_requests)}")
```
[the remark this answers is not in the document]
piperchester: Nice catch!

frontend/src/components/project/Buttons/JoinButton.js
```diff
@@ -0,0 +1,110 @@
...
+        className={classes.largeJoinButton}
+        color={"primary"}
+        disabled={requestedToJoin === true}
+        onClick={handleSendProjectJoinRequest}
+        startIcon={!requestedToJoin && <AddIcon />}
+        variant="contained"
+      >
+        <div className={classes.buttonLabel}>
+          {/* TODO: update german texts here */}
+          <div className={classes.buttonText}>{requestedToJoin ? "Requested" : "Join"}</div>
```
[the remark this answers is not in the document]
piperchester: Added in 4a641f4b.

frontend/src/components/project/ProjectContent.js
```diff
@@ -177,32 +242,54 @@ export default function ProjectContent({
...
+              {user_permission &&
+                [ROLE_TYPES.all_type, ROLE_TYPES.read_write_type].includes(user_permission) && (
+                  <>
+                    {/* Badge is dynamic based on the number of membership requesters */}
+                    <Badge badgeContent={requesters.length} color="primary">
+                      <Button
+                        className={classes.editProjectButton}
+                        onClick={viewOpenProjectRequests}
+                      >
+                        Review join requests
```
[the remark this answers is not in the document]
piperchester: Added in ae2c3a19.

frontend/src/components/communication/notifications/Notification.js
```diff
@@ -69,31 +69,29 @@ const NOTIFICATION_TYPES = [
...
+  else if (type === "idea_comment") return <IdeaCommentNotification notification={notification} />;
+  else if (type === "reply_to_idea_comment")
+    return <IdeaCommentReplyNotification notification={notification} />;
+  else if (type === "person_joined_idea")
+    return <PersonJoinedIdeaNotification notification={notification} />;
+  else if (type === "project_like") return <ProjectLikeNotification notification={notification} />;
+  else return <></>;
 }
 
 const PersonJoinedIdeaNotification = ({ notification }) => {
```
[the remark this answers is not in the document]
piperchester: Added to list of follow ups 👍

backend/organization/urls.py
```diff
@@ -156,6 +165,16 @@
         project_views.LeaveProject.as_view(),
         name='leave-project'
     ),
+    path(
+        'projects/<str:project_slug>/request_membership/<str:user_slug>/',
+        project_views.RequestJoinProject.as_view(),
+        name = 'request-join-entity'
+    ),
+    path(
+        'projects/<str:project_slug>/request_membership/<str:request_action>/<str:request_id>/',
```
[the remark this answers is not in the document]
piperchester: Added to list of follow ups 👍

frontend/src/components/communication/notifications/Notification.js
```diff
@@ -172,25 +170,41 @@ const MentionNotification = ({ notification, texts, locale }) => {
...
           primary={
-            sender.first_name +
-            " " +
-            sender.last_name +
-            " " +
-            texts.mentioned_you_in_comment_about_project
+            // TODO: should refactor to support richer notification requests; see
+            // // https://github.com/climateconnect/climateconnect/issues/898
+            notification.text.includes("wants to join your project")
+              ? "A user wants to join your project!"
```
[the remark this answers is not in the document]
piperchester: Added in 201ce48b.

backend/organization/permissions.py
```diff
@@ -241,7 +241,7 @@ def has_permission(self, request, view):
 
         # Detect for AnonymousUser
         if not request.user.is_authenticated:
-            return True
+            return False
```
piperchester: @positiveimpact to be followed up, I actually made this change and then started getting a 401 with this error: https://stackoverflow.com/questions/26906630/django-rest-framework-authentication-credentials-were-not-provided -- changed back in: a7b52d3f

positiveimpact: Maybe that's some cookie setup problems? We definitely can't return True here, otherwise ever user who isn't authenticated will have admin priviledges :D Maybe it's worth to pair on this issue?

backend/organization/permissions.py
```diff
@@ -227,9 +227,38 @@ def has_permission(self, request, view):
...
+        # TODO: this seems to sporadically be throwing an error:
+        #
+        #  raise TypeError('Cannot cast AnonymousUser to int. Are you trying to use it in place of User?')
+        # TypeError: Cannot cast AnonymousUser to int. Are you trying to use it in place of User?
+        #
+        # When calling from the post() in ManageJoinProject. Verify
+        # that we don't have duplicate requests causing the error to be thrown,
+        # or some other method.
+
+        # TODO: request.user is Anonymous
```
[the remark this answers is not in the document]
piperchester: Should be resolved now.

backend/organization/permissions.py
piperchester: So I don't believe this was actually needed, see note in: [c9dddfa](https://github.com/climateconnect/climateconnect/pull/674/commits/c9dddfaf0a4c8a2b37a94e31e20adb11e54e73c2).

ddhanesha:
@piperchester you have a migration missing here. Can you run `python manage.py makemigrations` and `python manage.py migrate`?

ddhanesha (approved):
I had chat with @piperchester and tested it locally. The functionality is working as expected and code is looking good. There are few followup work that we will do after this PR is merged. 

Approved!

PR #676, opened 2021-08-01
fix(project.py): AttributeError on project.language

piperchester:
Noticed this on the latest `develop`. This was causing no viewable projects in my local setup:

Before:

Screen Shot 2021-08-01 at 12 43 37 PM

After:

Screen Shot 2021-08-01 at 12 44 07 PM


positiveimpact (approved):
LGTM!

PR #680, opened 2021-08-02
chore: remove duplicate yarn.lock; run yarn-deduplicate; misc updates

piperchester:
Noticed we had a duplicate yarn.lock checked in. Also ran [yarn-deduplicate](https://www.npmjs.com/package/yarn-deduplicate) which is a really useful way to reduce the number of dependencies we ship.

## Before landing

1. PR has meaningful title
1. `yarn lint` passes
1. `yarn format` passes


positiveimpact (approved):
LGTM, everything looking fine locally!

ddhanesha (dismissed):

PR #682, opened 2021-08-05
Actually call create_language_test_data

samschaevitz:
Languages weren't initializing; this calls the language creation function in create_test_data.


positiveimpact (approved):
LGTM

PR #723, opened 2021-10-18
fix: remove redundant hints on /share page

piperchester:
This removes the "Use the search bar..." `<Typography />` components under `/share` to remove redundancy and clean up the space some. See them here:

Screen Shot 2021-10-18 at 2 02 10 PM

The helper text under the search bar, as well as the placeholder text within the search bar, should suffice here.

Also
- Pulls in updates from https://github.com/climateconnect/climateconnect/pull/722
- Misc. readability fixes like prop alphabetization and delineating 1st and 3rd party imports

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.md
+++ b/.github/ISSUE_TEMPLATE/bug.md
@@ -1,0 +1,40 @@
+---
+name: Bug
+about: Create a report to help us improve
+title: 'Bug: '
+labels: ''
+assignees: ''
+
+---
+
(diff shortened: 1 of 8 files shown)


positiveimpact:
I personally don't feel that it's redundant for 2 reasons:
1. The placeholder just tells you _how_ to search but I don't think that it would be clear for everybody that you look for people to add to your team in this search bar
2. The now removed text really drew the eye to itself making it much harder to not understand what the search bar is for

I worry that removing the text could make less projects add teams to be honest.

PR #757, opened 2021-11-05
fix: show list of liking users when clicking the link below the LikeButton

### LINKED ISSUE #756: Bug: Link below the like button does not show the list of users who liked
**Describe the bug**
Clicking on the link below the like button you should be shown a list of the users who liked the project. But the list is empty even if there are likes already.

**To Reproduce**
1. Go to any project page that already has project likes
2. Click on the link below the button ("X Like(s)")
3. Instead of the list of the users who liked the project you get: "This project does not have any likes yet."

verena-e-p:
Closes #756 

When clicking the link below the LikeButton, it didn't show the list of users who are liking the project. This is now fixed.

## Before landing

1. PR has meaningful title
1. `yarn lint` passes
1. `yarn format` passes

### DIFF
--- a/backend/organization/views/project_views.py
+++ b/backend/organization/views/project_views.py
@@ -888,10 +888,13 @@ class ListProjectLikesView(ListAPIView):
     serializer_class = ProjectLikeSerializer
 
     def get_queryset(self):
-        project = Project.objects.filter(url_slug=self.kwargs['url_slug'])
-        if not project.exists():
+        try:
+            project = Project.objects.get(url_slug=self.kwargs['url_slug'])
+        except Project.DoesNotExist:
             return None
-        likes = ProjectLike.objects.filter(project=project[0])
+            
+        
+        likes = ProjectLike.objects.filter(project=project)
         return likes        
 
 class LeaveProject(RetrieveUpdateAPIView):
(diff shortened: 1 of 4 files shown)


backend/organization/views/project_views.py
```diff
@@ -888,10 +888,10 @@ class ListProjectLikesView(ListAPIView):
     serializer_class = ProjectLikeSerializer
 
     def get_queryset(self):
-        project = Project.objects.get(url_slug=self.kwargs['url_slug'])
+        project = Project.objects.filter(url_slug=self.kwargs['url_slug'])
         if not project.exists():
             return None
```
positiveimpact, line 893:
```suggestion
        try:
            project = Project.objects.get(url_slug=self.kwargs['url_slug'])
        except Project.DoesNotExist:
            return None
            
        
```

backend/organization/views/project_views.py
```diff
@@ -888,10 +888,10 @@ class ListProjectLikesView(ListAPIView):
...
-        likes = ProjectLike.objects.filter(project=project)
+        likes = ProjectLike.objects.filter(project=project[0])
```
positiveimpact, line 894:
```suggestion
        likes = ProjectLike.objects.filter(project=project)
```

positiveimpact (changes requested):
Great job on discovering this and thanks for your initiative for fixing this. Your fix definitely works but I wouldn't revert to `filter` because `get` makes more sense if we just want to retrieve one object. Instead we should wrap the get statement into a try-catch block as suggested

Issue #504, opened 2021-02-23
Go back button


tobiasrehm:
button for easy access to the previous page. Design needed

PR #759, opened 2021-11-08
Feat: GoBackButton

verena-e-p:
# Description #504 #

The GoBackButton component is now implemented and added to the project page. When clicked it either directs the user back to the hub he came from or to the browse page. It uses a query string for navigation.

## Design ##
* different designs for different screen sizes

## Functionality ##
* if the project is entered through a hub page, the button directs the user back to that page
* in every other case the user is directed to the browse page, when clicking on the button

## Further Implementations ##
* the HubsSubHeader was added to the project page

## Before landing

1. PR has meaningful title
1. `yarn lint` passes
1. `yarn format` passes


frontend/src/components/project/Buttons/GoBackButton.js
```diff
@@ -0,0 +1,67 @@
...
+import React from "react";
+import useElementProps from "../../hooks/useElementProps";
+import ArrowBackIosIcon from "@material-ui/icons/ArrowBackIos";
+import { useRouter } from "next/router";
+
+const useStyles = makeStyles((theme) => ({
+  root: (props) => ({
+    position: "absolute",
+    height: props.tinyScreen ? 0 : props.containerHeight,
+    top: props.tinyScreen ? 0 : props.containerTop,
```
positiveimpact, line 12: Why are you using position `absolute` and these hook-calculated offsets here?
It seems like this should also be possible with just using display: `flex` on the parent element on desktop - and for mobile you don't seem to need the calculation.

positiveimpact (commented):
Good work overall, but for one point I didn't understand why you made it quite complicated. It looks overcomplicated but maybe I'm missing something and it's necessary.
For the HubsSubHeader, I'd not show any hub names but only the options "SectorHubs" and "CityHubs". Otherwise there's a bit too much going on that takes attention away from the project imo.

frontend/src/components/project/Buttons/GoBackButton.js
verena-e-p: Thank you for your feedback. I attempted to implement it (see recent commits). I hope the code is more suitable now. I'm still happy to take further feedback! :)

frontend/src/components/project/Buttons/GoBackButton.js
```diff
@@ -0,0 +1,56 @@
...
+      },
+    },
+  },
+}));
+
+export default function GoBackButton({ texts, tinyScreen, locale, containerClassName }) {
+  const classes = useStyles();
+
+  const router = useRouter();
+  const goBack = () => {
```
verena-e-p, line 27: The goBack-function only works for the projects page right now. Should I generalize it?

frontend/src/components/project/Buttons/GoBackButton.js
```diff
@@ -0,0 +1,56 @@
...
+        backgroundColor: theme.palette.primary.main,
+      },
+    },
+  },
+}));
+
+export default function GoBackButton({ texts, tinyScreen, locale, containerClassName }) {
+  const classes = useStyles();
+
+  const router = useRouter();
```
verena-e-p, line 26: The goBack-function only works for the projects page right now. Should I generalize it?

frontend/src/components/project/ProjectOverview.js
```diff
@@ -41,6 +42,16 @@ const useStyles = makeStyles((theme) => ({
...
     color: theme.palette.primary.main,
   },
+  largeScreenHeader: {
+    paddingTop: theme.spacing(4),
+    paddingBottom: theme.spacing(4),
+    textAlign: "center",
+  },
+  goBackButtonContainer: {
+    marginLeft: theme.spacing(1),
+    marginTop: theme.spacing(1),
```
verena-e-p, line 52: Is there a way to position the button on the image without using `absolute`?

frontend/src/components/project/Buttons/GoBackButton.js
positiveimpact: Good point. I think it's quite hard to generalize it unless you pass the page you want to go back to manually. Therefore it would maybe be better to rename it so that it's clear that it's not generic. E.g. `GoBackFromProjectPageButton`

frontend/src/components/project/ProjectOverview.js
positiveimpact: That would only be possible if you use a div with the background property rather than an image. If you'd like to use this option I can explain it more in detail in a call :)

positiveimpact (approved):
LGTM! Merging.

PR #762, opened 2021-11-10
fix: always update state when Followers/Likes are set

### LINKED ISSUE #761: Bug: Dialog window (to display followers/likes on project page) lost loading animation
**Describe the bug**
The dialog window which lists the likes/followers on the project page shows a wrong text instead of its loading animation when loading. 

**To Reproduce**
1. Go to any project page that has likes or followers.
2. Click on the 'Like(s)/Follower(s)'-Link below the corresponding button (on desktop screens) or tap and hold the button (on mobile screens).
3. The window displays the statement: "This project does not have any likes/followers yet".
4. After opening the dialog window you need to refresh your browser to reproduce the error.

**Expected behavior**
"There should be a loading animation if the likes/followers are not displayed immediately.

verena-e-p:
When the updateFollowers-/updateLikes-function was called outside of toggleShowFollowers/toggleShowLikes (in this case from within toggleFollowProject/toggleLikeProject) not all corresponding states were updated. This is now fixed.

Closes #761 

## Before landing

1. PR has meaningful title
1. `yarn lint` passes
1. `yarn format` passes

### DIFF
--- a/frontend/src/components/project/ProjectPageRoot.js
+++ b/frontend/src/components/project/ProjectPageRoot.js
@@ -304,7 +304,7 @@ export default function ProjectPageRoot({
   const toggleShowFollowers = async () => {
     setShowFollowers(!showFollowers);
     if (!initiallyCaughtFollowers) {
-      updateFollowers();
+      await updateFollowers();
       handleReadNotifications(NOTIFICATION_TYPES.indexOf("project_follower"));
     }
   };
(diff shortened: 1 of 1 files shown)


frontend/src/components/project/ProjectPageRoot.js
```diff
@@ -306,12 +306,12 @@ export default function ProjectPageRoot({
     if (!initiallyCaughtFollowers) {
       updateFollowers();
```
positiveimpact, line 307:
```suggestion
      await updateFollowers();
```

The problem here seems to be that you don't wait for updateFollowers to finish. 
I think it makes more sense to leave it here and use the await keyword. Same below with likes

positiveimpact (approved):
LGTM!

PR #763, opened 2021-11-10
Feat: New section for the project page 'DiscussionPreview'

### LINKED ISSUE #505: Discussion preview
Implement a new section to the project tab, After the Project description. 
image
-It shows the latest comment from the discussion tab 
- shows the number of comments next to "Discussion" 
- clicking anywhere on this section leads to the Discussion (Comments) Tab

verena-e-p:
Closes #505 

# Description #505 #

The DiscussionPreview is now implemented and added to the project page. Its heading is the same as for the discussion tab and it displays the latest parent comment (including profile picture and name of the commenting user, date of the comment).

## Design ##
* visual indication when the user is hovering over or clicking on the section

## Functionality ##
* clicking anywhere on the preview section takes the user to the discussion tab on the project page

## Before landing

1. PR has meaningful title
1. `yarn lint` passes
1. `yarn format` passes

### DIFF
--- a/frontend/src/components/communication/Post.js
+++ b/frontend/src/components/communication/Post.js
@@ -123,7 +123,6 @@ export default function Post({
             />
           </Link>
           <span className={classes.messageWithMetaData}>
-            {type !== "preview" && (
               <div className={classes.metadata}>
                 <Link
                   color="inherit"
@@ -144,7 +143,6 @@ export default function Post({
                   <DateDisplay date={new Date(post.created_at)} />
                 </Typography>
               </div>
-            )}
             {type === "preview" ? (
               <Typography>
                 <Truncate lines={truncate} ellipsis={"..."}>
--- a/frontend/src/components/communication/Post.js
+++ b/frontend/src/components/communication/Post.js
@@ -118,7 +118,7 @@ export default function Post({
             />
           </Link>
           <span className={classes.messageWithMetaData}>
-            {type != "preview" && (
+            {type !== "preview" && (
               <div className={classes.metadata}>
                 <Link
                   color="inherit"
--- a/frontend/src/components/communication/Post.js
+++ b/frontend/src/components/communication/Post.js
@@ -3,6 +3,7 @@ import { makeStyles } from "@material-ui/core/styles";
 import ExpandLessIcon from "@material-ui/icons/ExpandLess";
 import ExpandMoreIcon from "@material-ui/icons/ExpandMore";
 import React, { useContext } from "react";
+import Truncate from "react-truncate";
 import { getLocalePrefix } from "../../../public/lib/apiOperations";
 import { getImageUrl } from "../../../public/lib/imageOperations";
 import getTexts from "../../../public/texts/texts";
@@ -17,9 +18,10 @@ const useStyles = makeStyles((theme) => ({
   postDate: {
     color: theme.palette.grey[700],
   },
-  commentFlexBox: {
+  commentFlexBox: (props) => ({
     display: "flex",
-  },
+    alignItems: props.preview ? "center" : "stretch",
+  }),
   messageWithMetaData: {
     display: "flex",
     flexDirection: "column",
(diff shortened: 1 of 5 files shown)


verena-e-p:
Issue to be solved: When the browser window's width is decreased the text is decreased as well. But when the browser window's width gets increased again, the text continues to be decreased.

frontend/src/components/project/ProjectContent.js
```diff
@@ -357,3 +401,43 @@ function CollaborateContent({ project, texts }) {
...
+              <Truncate lines={3} ellipsis={"..."}>
+                {latestParentComment.content}
+              </Truncate>
+            </Typography>
+          </div>
+        </div>
+      </Link>
+    </>
+  );
+}
```
positiveimpact, line 443: I would create a new file for this

frontend/src/components/project/ProjectContent.js
```diff
@@ -357,3 +401,43 @@ function CollaborateContent({ project, texts }) {
...
+                  ? getImageUrl(latestParentComment.author_user.image)
+                  : getImageUrl(latestParentComment.author_user.thumbnail_image)
+              }
+            />
+            <Typography className={classes.commentTextDiscussionPreview}>
+              <Truncate lines={3} ellipsis={"..."}>
+                {latestParentComment.content}
+              </Truncate>
+            </Typography>
+          </div>
```
positiveimpact, line 438: For this it would make sense to use `frontend/src/components/communication/Posts`. This includes all the functionality necessary for displaying comments properly, e.g. parsing links and user-tagging. You could then adapt this function to also fit your needs (while keeping it working as is for the current use-cases), e.g. by adding a variable `truncate` which could indicate whether and after how many lines a text should be truncated.

positiveimpact (changes requested):
I think the CommentsPreview needs to be implemented a bit differently to incorporate the existing `Posts` component. For now I didn't investigate the bug further and would check if it persists after using the `Posts`component.

frontend/src/components/project/ProjectContent.js
verena-e-p: The Posts component is now used for the DiscussionPreview! Using the component indeed got rid of the bug mentioned above :slightly_smiling_face:

frontend/src/components/project/ProjectContent.js
verena-e-p: The DiscussionPreview-function has shrunk quite a bit since using the post component. Do you still prefer to put it in its own file?

positiveimpact: Yeah I think that makes sense since it's logically it's own component

verena-e-p: Alright, it's done

positiveimpact (changes requested):
This is actually not working reliably for me or at least has quite some delay.
Also some feedback on hover and click similar to material UI buttons would make the experience much more interactive. Other than that great work! :)

frontend/src/components/communication/Post.js
```diff
@@ -115,28 +118,39 @@ export default function Post({
...
           <span className={classes.messageWithMetaData}>
-            <div className={classes.metadata}>
-              <Link
-                color="inherit"
-                href={getLocalePrefix(locale) + "/profiles/" + post.author_user.url_slug}
-                target="_blank"
-              >
-                <Typography variant="body2" className={classes.username}>
-                  {post.author_user.first_name + " " + post.author_user.last_name}
+            {type != "preview" && (
```
positiveimpact, line 121:
```suggestion
            {type !== "preview" && (
```
Always use `===` or `!==` unless you concrete don't want to do that. It's a bit safer since it makes it less likely to have an unexpected equals.

positiveimpact: Same goes for all other instances of `==` or `!=`

positiveimpact (approved):
Works really well now!
I left 2 nits which I think would make sense to fix.
One more remark: I think I wouldn't link to the author in the preview. That might be a bit confusing if you wanted to jump to the comment section, click on the user's avatar, and are suddenly on their profile page.

positiveimpact (approved):
LGTM! :)

PR #765, opened 2021-11-11
Feat: Icon FollowButton - Designupdate

### LINKED ISSUE #764: FollowButton design update
Add the earth design to the icon of the FollowButton (used when a user is following). 
FollowButtonDesign

verena-e-p:
Closes #764 

## Description
The icon of the FollowButton now has the earth-design when a user is following.

## Additional changes 
These were also added to the **LikeButton**! 
* The [visibility issues](https://github.com/climateconnect/climateconnect/pull/765#pullrequestreview-804088849) are fixed.
* The CircularProgress, which is indicating the pending state of the button, wasn't visible anymore. This is now fixed as well. 

## Before landing

1. PR has meaningful title
1. `yarn lint` passes
1. `yarn format` passes


positiveimpact (changes requested):
Great idea to generalize the ButtonIcon! There is just some small visibility issues, I messaged you about that on Slack.

frontend/src/components/project/Buttons/LikeButton.js
```diff
@@ -113,13 +135,15 @@ export default function LikeButton({
...
-          startIcon={<LikeIcon size={20} color={isUserLiking ? "earth" : "white"} />}
+          startIcon={<ButtonIcon icon="like" size={26} color={isUserLiking ? "earth" : "white"} />}
           color={isUserLiking ? "secondary" : "primary"}
           disabled={likingChangePending}
           className={classes.largeLikeButton}
         >
-          {likingChangePending && <CircularProgress size={13} className={classes.fabProgress} />}
-          {isUserLiking ? texts.liked : texts.like}
+          <div className={classes.buttonLabel}>
+            {likingChangePending && <CircularProgress size={20} className={classes.fabProgress} />}
```
positiveimpact, line 144:
```suggestion
            <CircularProgress size={20} className={`${classes.fabProgress} ${!likingChangePending && classes.hidden}`} />
```

For me the loading animation actually doesn't show load. It only works when I directly set likingChangePending to true. It seems like the time isn't enough to load the SVG. Therefore this workaround could be a good solution: actually rendering but hiding the CircularProgress directly on page load.

positiveimpact (commented):
LGTM other than my comment

positiveimpact (approved):
LGTM!

Issue #769, opened 2021-11-19
Visual refresh to organization preview cards


positiveimpact:
# Overview
The preview cards for organization accounts should be reworked to make them more visually appealing and to make it easier to quickly get an overview of the organizations on Climate Connect without clicking them.

# Current state
The current organization preview cards can be seen [here](https://climateconnect.earth/browse?&#organizations). The code for them is located in `frontend/src/components/organization/OrganizationPreviews.js`.

# Desired state after reword
A sketch of how organization cards should look like can be found [here](https://xd.adobe.com/view/263c33d4-4ee4-441b-7c68-f4533c94f5ee-500d/screen/0efaf8a1-7aed-4972-bb12-ce433aa43c29/). The most important changes are:
- changing the background color
- adding box shadow
- Displaying the organization type in a Material UI <Chip /> component directly under the image
- Showing a preview of the organization's short_description on the card. This should be cut off at some max length by using react-truncate. If an organization has a description but no short_description, the description should be shown here.

Margins should not be exactly like on the design, it's just a draft. Also the font-weight for location and short_description should be normal (in the design they are bold). Generally it's not necessary to exactly implement the design as is but just to implement the general idea into code.

PR #770, opened 2021-11-21
Organization cards redesign

TheMeerCat:
[Issue #769: Visual refresh to organization preview cards](https://github.com/climateconnect/climateconnect/issues/769)

**This is the mockup:** https://xd.adobe.com/view/263c33d4-4ee4-441b-7c68-f4533c94f5ee-500d/screen/0efaf8a1-7aed-4972-bb12-ce433aa43c29/
**This is how it looks:** https://ibb.co/HN1LX68

This is just a proposal, if you want me to do it exactly as the mockup, I am happy to change it.


frontend/src/components/organization/OrganizationPreview.js
```diff
@@ -48,28 +56,18 @@ const useStyles = makeStyles((theme) => {
...
-          alt={organization.name}
-          size="large"
-          src={getImageUrl(organization.thumbnail_image)}
-          className={classes.media}
-          component="div"
+        <OrganizationPreviewHeader organization={organization} />
+        <OrganizationPreviewBody
+          organization={organization}
+          showOrganizationType={showOrganizationType}
         />
```
positiveimpact, line 70: I like the split into two components here :+1:

frontend/src/components/organization/OrganizationPreview.js
```diff
@@ -1,25 +1,33 @@
...
+      backgroundSize: "calc(100% - 1px) 100%",
+      borderRadius: "5px",
       textAlign: "center",
-      height: 250,
+      height: "350px",
+      boxShadow: "rgba(99, 99, 99, 0.2) 0px 2px 8px 0px;",
+      padding: "0 12px",
     },
     bold: {
       fontWeight: "bold",
```
positiveimpact, line 33: It seems like this class is now unused. Would be great if you go through the classes here that aren't used in the component anymore and remove them.

frontend/src/components/organization/OrganizationPreviewBody.js
```diff
@@ -0,0 +1,92 @@
...
+import { makeStyles } from "@material-ui/core/styles";
+import PlaceIcon from "@material-ui/icons/Place";
+
+const useStyles = makeStyles((theme) => {
+  return {
+    bold: {
+      fontWeight: "bold",
+    },
+    locationName: {
+      fontWeight: "bold",
```
positiveimpact, line 12: I tested it locally and I think making the location bold highlights it a bit too much. I think semi-bold is the better choice here.
```suggestion
      fontWeight: 600,
```

frontend/src/components/organization/OrganizationPreviewBody.js
```diff
@@ -0,0 +1,92 @@
...
+      </Box>
+      <Box
+        className={classes.summaryBox}
+        ref={(textBox) => {
+          const linesEstimation = textBox?.clientHeight / 25;
+          setLinesOfText(Math.floor(linesEstimation));
+        }}
+      >
+        <Typography className={classes.shortenedSummary} style={{ WebkitLineClamp: linesOfText }}>
+          {organization.short_description ?? organization.description}
```
positiveimpact, line 87: Nice, I actually didn't know about the `??` operator. Will use that in the future as well.

frontend/src/components/organization/OrganizationPreviewBody.js
```diff
@@ -0,0 +1,92 @@
...
+      overflow: "hidden",
+      WebkitBoxOrient: "vertical",
+      display: "-webkit-box",
+    },
+    summaryBox: {
+      overflow: "hidden",
+    },
+    contentWrapper: {
+      padding: 0,
+      display: "grid",
```
positiveimpact, line 51: Generally, we usually work with flexbox rather than grid unless we're displaying a longer list of the same component but using grid is also fine.

positiveimpact (commented):
LGTM overall, except the few nits I made. 
However, it would be nice to also display the number of members and number of projects of the organization like in the design. For that you'd have to update the OrganizationStubSerializer in `backend/organization/serializers/organization.py`. I'm happy to walk you through that in a call.

backend/organization/serializers/organization.py
```diff
@@ -120,10 +120,12 @@ class OrganizationCardSerializer(serializers.ModelSerializer):
     types = serializers.SerializerMethodField()
     location = serializers.SerializerMethodField()
     name = serializers.SerializerMethodField()
+    hubs_count = serializers.SerializerMethodField()
     class Meta:
         model = Organization
         fields = (
-            'id', 'name', 'url_slug', 'thumbnail_image', 'location', 'types', 'short_description'
+            'id', 'name', 'url_slug', 'thumbnail_image', 'location', 'types', 'short_description', 'organization_size',
```
positiveimpact, line 127: Instead of  `organization_size`we should instead count the number of `OrganizationMember`s that are part of this organization

positiveimpact: Because organization_size is an optional field organization creators can fill out so not every organization has a value for that. Also it's quite motivating to see this number grow as more of the organization members join Climate Connect.

positiveimpact (commented):
LGTM other than 2 things I noted in my comments. These are both misunderstandings that are mainly due to unclear communication on my side.

backend/organization/serializers/organization.py
TheMeerCat: Ah, sure. I didn't know that. It definitely makes sense :-)

backend/organization/serializers/organization.py
```diff
@@ -120,10 +120,12 @@ class OrganizationCardSerializer(serializers.ModelSerializer):
     types = serializers.SerializerMethodField()
     location = serializers.SerializerMethodField()
     name = serializers.SerializerMethodField()
+    hubs_count = serializers.SerializerMethodField()
```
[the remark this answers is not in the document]
TheMeerCat: Sure, I changed it, too. :-)

TheMeerCat:
> LGTM other than 2 things I noted in my comments. These are both misunderstandings that are mainly due to unclear communication on my side.

I fixed it. No, all good. I misunderstood it a bit and at least I learned something new now. :-)

positiveimpact (approved):
Thanks for the changes. LGTM now! :)

Issue #780, opened 2021-11-25
Add button to CityHubs to message local ambassador


positiveimpact:
# Overview
For Desktop screens we want to add an easy opportunity for users to get in contact with the local ambassador of a CityHub with one click. For that purpose we want to add a contact button on the bottom right of Desktop screens on CityHubs.
[Design draft link](https://xd.adobe.com/view/b12f4514-bd4b-47b7-b21c-0e4922baee92-e47c/screen/3cf11f50-778c-4d45-9ecd-b6a823a6f56d/)

# Steps to achieve this
I would suggest the following steps to achieve this:
- Create a new model `LocalAmbassador` in django. This can be placed in the `hubs` app. It should just have a foreign key to `User` and `Hub` as well as a `title` and `title_de` column which should contain the text written under the name on the button in the design (e.g. "Lokaler Ansprechpartner")
- When serializing a CityHub the local ambassador should also be passed to the frontend
- For implementing the actual button in the frontend I'd use the component `frontend/src/components/project/buttons/ContactCreatorButton`. Potentially we'll have to add some additional options to it to make it fit this use-case

# Impact
This feature can make a huge difference in establishing the local ambassador within the city and allowing users to easily get support or help.

ddhanesha:
This feature has been completed as far as I can tell. Moving this to done.

PR #785, opened 2021-11-29
780 Contact local ambassador

TheMeerCat:
Issue #780 

I would like to discuss the current progress. :-) 
- I've created the new table that represents the new relation between user and hub. As the relation is 1:1, I think maybe we would like to add this information to 'hub' table and avoid having a separate table for this?
- I am a bit unsure about the ContactCreatorButton component. I tried to split it and reuse it but maybe it would make sense to split the huge component even more? Or do we prefer to adjust the big component to fit this use case? Personally, I am more fan of small components. :-)
- How do we set the local ambassador? 😅 I guess there should be some field in configuration?

Aaand the pipeline is failing but looks like it has nothing to do with my code 🤔


positiveimpact (commented):
Looks good so far! 
1. Yeah, we could definitely add this to the hub table. I thought smaller tables would be a bit easier to read but since we already have humongous tables like `UserProfile` it probably makes sense to just pull it in there. And I can't think of a use-case where we'd need more than 1 local ambassador for 1 hub atm. 
2. Yeah you're right, it's getting pretty big :D As long as the code that's the same is generalized and not written multiple times it probably makes more sense to split it up, so we can go with smaller components here.
3. I don't quite understand the question :smile: I'd just create a database entry for the local ambassador, the easiest way would be through django admin (localhost:8000/admin/). But maybe that's not what the question is getting at :D

marketajancova:
1. As I've already implemented the option with the separate table (and it seems to work), you can decide which option you prefer. :-)
2. The component combines 4 different modes of one button, I would maybe split it into 5 components then and these buttons would be easier to reuse. I will check it once more, maybe it's not as simple as I think :-D 
3. Nono, I meant how we want to configure it "real life". I currently set it manually in the database to test it but I guess that's not the way we want to go in the production environment (?). So there should also be an UI option.

marketajancova:
(Ops, commenting using work account) 😂

positiveimpact:
Sorry for the late reply!

> 1. As I've already implemented the option with the separate table (and it seems to work), you can decide which option you prefer. :-)

If you've already done it like this, let's just keep it like that for simplicity :)

> 2. The component combines 4 different modes of one button, I would maybe split it into 5 components then and these buttons would be easier to reuse. I will check it once more, maybe it's not as simple as I think :-D

Makes sense :+1: 

> 3. Nono, I meant how we want to configure it "real life". I currently set it manually in the database to test it but I guess that's not the way we want to go in the production environment (?). So there should also be an UI option.

I think for the start I'd just do it through django admin. In the future it could be worth it to create UI options for this and other things for admins but so far it's not worth it I think

marketajancova:
Nice, sounds good (and it makes my job easier 😇 ). I will update the PR accordingly. Thanks @positiveimpact

TheMeerCat:
https://user-images.githubusercontent.com/18369329/145074090-97355462-52d2-4cc5-92db-00751e114b55.mov

I changed the button a tiny bit. I made it expandable on hover so it is not that huge. I find it a bit better like that (I think it was a bit disturbing before). :-)
I started refactoring the "big" ContactCreatorButton and I started to be a little bit afraid as there were more and more changes. I feel that it would be very easy to introduce a bug (e.g., by forgetting passing a parameter) so I decided to don't do it now. I would suggest considering it a technical debt and fix it later (e.g. during the fix it week) but I think we don't want to introduce more bugs right now. :-)
If you don't mind, I would therefore keep it as it is.

frontend/public/texts/getHubTexts.js
```diff
@@ -120,6 +120,14 @@ export default function getHubTexts({ hubName }) {
       en: "Add Hubs (sectors) in which your organization is active",
       de: "Füge Hubs (Sektoren) hinzu, in denen deine Organisation aktiv ist",
     },
+    please_create_an_account_or_log_in_to_contact_the_ambassador: {
+      en: "Please sign up to get in contact with others.",
+      de: "Bitte melde dich an, um Menschen zu kontaktieren.",
+    },
+    contact_ambassador: {
+      en: "Contact Local Ambassador",
+      de: "Lokalen Ambassador Kontaktieren",
```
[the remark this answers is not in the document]
positiveimpact: Saving the pronouns could be done by adding a column to the database model

frontend/src/components/hub/ContactAmbassadorButton.js
```diff
@@ -0,0 +1,69 @@
...
+      });
+    }
+
+    const chat = await startPrivateChat(localAmbassador.user, token, locale);
+    Router.push("/chat/" + chat.chat_uuid + "/");
+  };
+
+  return (
+    <>
+      {localAmbassador && (
```
positiveimpact, line 55: Maybe we could make this check outside of the component? So only use this component if there is a local ambassador.

frontend/src/components/hub/ContactAmbassadorButton.js
```diff
@@ -0,0 +1,69 @@
...
+  const { locale, user } = useContext(UserContext);
+  const cookies = new Cookies();
+  const token = cookies.get("token");
+  const texts = getTexts({ page: "hub", locale: locale });
+
+  const handleClickContact = async (e) => {
+    e.preventDefault();
+
+    if (!user) {
+      return redirect("/signup", {
```
positiveimpact, line 43:
```suggestion
      return redirect("`${getLocalePrefix(locale)}/signup`, {
```
Always use localization for links :) (Don't forget to import getLocalePrefix

positiveimpact (commented):
LGTM except my comments. Sorry for the super late review, I was really slow on reviewing PRs recently. I feel like it would be nice if at least the image of the local ambassador is always visible to create a bit more of a connection to the local ambassador. We could either achieve this by making the button always expanded or by putting a small image of the local ambassador into the button itself

TheMeerCat:
I am closing this PR because the branch is already very old and it's difficult to solve all the conflicts. I have created a new one. Let's continue there: https://github.com/climateconnect/climateconnect/pull/874

PR #799, opened 2021-12-05
Preventing state update in a loop

TheMeerCat:
State of the main page component is constantly getting updated when the WebSocket connection keeps failing. This causes an infinite loop of re-rendering, which contains sending some Web Requests to our backend from the child components – these two variables are propagated into children that rely on them. 
I think we should avoid this state and let the WebSocket connection keep failing more silently (if it has to happen).

### DIFF
--- a/frontend/pages/_app.js
+++ b/frontend/pages/_app.js
@@ -181,10 +181,13 @@ export default function MyApp({
       //
       // Revisit this code after the most recent state testing from
       // https://github.com/climateconnect/climateconnect/pull/709
-      setSocketConnectionState("closed");
+      if (socketConnectionState !== "closed") {
+        setSocketConnectionState("closed");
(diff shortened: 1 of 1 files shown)


positiveimpact (approved):
LGTM! On the main server the websocket connection usually shouldn't constantly fail and didn't for me when I just checked the production server. But it's still nice for it to not fail as loudly if it does since we always have the fallback to normal requests

marketajancova:
Interestingly, I could not reproduce the problem anymore (on prod) today but I am sure that it was there on Saturday in cases I described you in chat 🤔. Never mind; in any way, I would keep an eye on it and if the problem happens again, it might be worth looking into these components and there might be some other modifications like that, causing some regular/unwanted re-rendering.

Issue #831, opened 2021-12-13
Bug: random "true" printed on the org page of a sub-organization


positiveimpact:
**Describe the bug**
When you go to the page of an organization that has a parent organization you will see the ": true" on the actual website which is definitely not intended

**To Reproduce**
1. Make sure you have at least 2 organization profiles
2. Make one organization the parent organization of the other
3. The child organization will have this bug on its organization page 

**Expected behavior**
The ": true" should not appear

**Screenshots**
![Screenshot from 2021-12-13 14-48-45](https://user-images.githubusercontent.com/10531079/145824338-ad48556a-461f-4637-8f3f-74f11bbb2463.png)

positiveimpact:
Closed by #835

PR #835, opened 2021-12-13
fix: 831 remove weird things that are printed

marketajancova:
#831


positiveimpact (approved):
LGTM! :) Great job! Just added some styling for the display of the parent org.

PR #833, opened 2021-12-13
Fixed: YouTube link can't be cut off anymore when truncating project description

verena-e-p:
* The index to cut off the project description for the show more / show less functionality is now calculated to avoid cutting in between YouTube links. ( Closes #506 )

* Also ran `yarn lint` and `yarn format`

## Before landing

1. PR has meaningful title
1. `yarn lint` passes
1. `yarn format` passes

PR #843, opened 2021-12-14
Fixed: only calculate length when there is a description

verena-e-p:
Follow up to #833 
The `maxDescriptionLength` was calculated even if there was no description which caused a 'properties of null' error. This is now fixed.


positiveimpact (approved):
LGTM! :)

PR #850, opened 2021-12-15
Setting up caching and cache frequently fetched and unchanged APIs. 

ddhanesha:
## Description

We have few APIs that we use very frequently, but the response data does not change very often in our system. These are very database heavy operations that takes long timer to fetch in the frontend. This PR is to cache those frequently used APIs. 

Note: You need to run redis server for this. You can run it using docker. We have docker file configured in our system. 

## Test plan
* You need to setup `REDIS_URL` in `.backup_env` file. Example: `REDIS_URL="redis://127.0.0.1:6379/0"`.
* The go to /browse page on frontend and look for members tab. First time you load the page it will make a database call but later if you visit them same page it will render data from the cache file. 
* You can check if the cache is set properly or not by opening django shell. Run following commands.
```
python manage.py shell
>>> from django.core.cache import cache
>>> cache.keys('*')
```
This will return cache keys that we store for APIs. You should also test filters and searches to make sure it works as expected. 

## Before landing

1. PR has meaningful title
1. `yarn lint` passes
1. `yarn format` passes

### DIFF
--- a/backend/climateconnect_api/views/user_views.py
+++ b/backend/climateconnect_api/views/user_views.py
@@ -152,7 +152,7 @@ class ListMemberProfilesView(ListAPIView):
     search_fields = ['name']
     serializer_class = UserProfileStubSerializer
 
-    @method_decorator(cache_page(60*60))
+    @method_decorator(cache_page(settings.DEFAULT_CACHE_TIMEOUT))
     def dispatch(self, *args, **kwargs):
         return super(ListMemberProfilesView, self).dispatch(*args, **kwargs)
 
--- a/backend/climateconnect_api/views/user_views.py
+++ b/backend/climateconnect_api/views/user_views.py
@@ -1,4 +1,3 @@
-
 from ideas.serializers.idea import IdeaFromIdeaSupporterSerializer
 from ideas.models.support import IdeaSupporter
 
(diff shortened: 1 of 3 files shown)


backend/climateconnect_main/settings.py
```diff
@@ -299,3 +299,16 @@
         }
     }
 }
+
+# Setting up cache
```
piperchester, line 303: Nit:
```suggestion
# Setting up cache. Read more at
# https://docs.djangoproject.com/en/4.0/topics/cache/
```

backend/climateconnect_api/views/user_views.py
```diff
@@ -151,6 +152,10 @@ class ListMemberProfilesView(ListAPIView):
     search_fields = ['name']
     serializer_class = UserProfileStubSerializer
 
+    @method_decorator(cache_page(settings.DEFAULT_CACHE_TIMEOUT))
```
piperchester, line 155: Could we get away with just:

```suggestion
    @cache_page(settings.DEFAULT_CACHE_TIMEOUT)
```

backend/climateconnect_api/views/user_views.py
```diff
@@ -151,6 +152,10 @@ class ListMemberProfilesView(ListAPIView):
...
+    def dispatch(self, *args, **kwargs):
+        return super(ListMemberProfilesView, self).dispatch(*args, **kwargs)
```
piperchester, line 157: Nit: does this capture it? Anything else you might want to add for Django newbies like myself?
```suggestion
    # We use a custom dispatch to define behavior that returns the cached HTTP response.
    # https://docs.djangoproject.com/en/4.0/ref/class-based-views/base/#django.views.generic.base.View.dispatch
    def dispatch(self, *args, **kwargs):
        return super(ListMemberProfilesView, self).dispatch(*args, **kwargs)
```

piperchester (approved):
Awesome work! Thanks for providing the testing instructions too. It'd be fantastic if we looked at any perf instrumentation / dashboards we have before landing this change, and then after too to see if there's a noticeable change.

positiveimpact (approved):
LGTM now!

PR #853, opened 2021-12-16
Fix unexpected error in chat

### LINKED ISSUE #830: Bug: "Application error: a client-side exception has occured" when using messages
**Describe the bug 1**
A: When I am in a chat with another user at climateconnect.earth and then log out from there I get the error: "Application error: a client-side exception has occurred".
B: When I log out from my inbox a loading graphic (Climate Connect logo + text "loading ...") is shown forever. Refreshing the page gives me a 404 error in the browser window and a failed get request in the console (error 500).

**To Reproduce**
Steps to reproduce the behavior A:
1. Open a chat with another user on climateconnect.earth
2. Log out

And B:
1. Open your inbox on climateconnect.earth
2. Log out

**Describe the bug 2**
This seems to happen randomly, about 50% of the time. When sending a message to another user on climateconnect.earth the message is first sent and displayed correctly but then the "Application error: a client-side exception has occurred" turns up. After refreshing, everything seems normal.

**To Reproduce**
1. Send a message to another user on climateconnect.earth
2. You might need to repeat since the error does not happen every time

Not sure whether bugs 1 and 2 are related.

positiveimpact:
## Description

- Fixed index out of bound error in backend which through a 500 error. 
- For some reason the websocket sends each chat message twice. Fixed frontend implications of that by only showing a message once per id. 
- Fixed the unexpected error when logging out while in the chat window. This was caused by trying to access a property of the user object which is obv. null if you log out.

Closes #830 

## Test plan

1. Open a private chat (ideally with 2 accounts in 2 browsers)
2. Write a message
3. Make sure there is no error on either browser and the message arrives exactly once. 

Fixes were made to websocket mode, so you shouldn't be using the fallback HTTP requests

### DIFF
--- a/backend/chat_messages/utility/chat_setup.py
+++ b/backend/chat_messages/utility/chat_setup.py
@@ -26,15 +26,17 @@ def set_read(messages, user, is_private_message):
     # Here we are assuming that all messages passed to this function are from the same chat
     message_notifications = Notification.objects.filter(chat=messages[0].message_participant)
     if message_notifications.exists():
-        unread_user_notifications = UserNotification.objects.filter(
-            notification__in=message_notifications,
-            read_at=None,
-            user=user
-        )
-        if unread_user_notifications.exists():
-            user_notification = unread_user_notifications[0]
+        try:
+            unread_user_notification = UserNotification.objects.get(
+                notification__in=message_notifications,
+                read_at=None,
+                user=user
+            )
+            user_notification = unread_user_notification
             user_notification.read_at = datetime.now()
             user_notification.save()
+        except UserNotification.DoesNotExist:
+            print("there is no user notification for "+user.first_name)
 
 
 def create_private_or_group_chat(
--- a/backend/chat_messages/utility/chat_setup.py
+++ b/backend/chat_messages/utility/chat_setup.py
@@ -36,7 +36,7 @@ def set_read(messages, user, is_private_message):
             user_notification.read_at = datetime.now()
             user_notification.save()
         except UserNotification.DoesNotExist:
-            print("there is no user notification for "+user.first_name)
+            logger.error("there is no user notification for "+user.first_name)
 
 
 def create_private_or_group_chat(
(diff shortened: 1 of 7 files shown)


backend/chat_messages/utility/chat_setup.py
```diff
@@ -26,15 +26,17 @@ def set_read(messages, user, is_private_message):
...
+            unread_user_notification = UserNotification.objects.get(
+                notification__in=message_notifications,
+                read_at=None,
+                user=user
+            )
+            user_notification = unread_user_notification
             user_notification.read_at = datetime.now()
             user_notification.save()
+        except UserNotification.DoesNotExist:
+            print("there is no user notification for "+user.first_name)
```
samschaevitz, line 39: do you want to leave this as print() vs a logging statement?

frontend/src/components/layouts/layout.js
```diff
@@ -16,11 +16,11 @@ const useStyles = makeStyles((theme) => ({
     textAlign: "center",
     margin: `${theme.spacing(4)}px 0`,
   },
-  alert: {
+  alert: props => ({
     width: "100%",
     zIndex: 100,
-    marginTop: theme.spacing(-2),
-  },
+    marginTop: !props.donationCampaignRunning && theme.spacing(-2),
```
samschaevitz, line 22: Is this related to the error checking above ? If not maybe split into another PR...

samschaevitz (approved):
Looks mostly fine, PTAL before merging

frontend/src/components/layouts/layout.js
positiveimpact: I think I would leave it in here because this is another bug that happens when logging out while in the chat window: You'll get redirected to /signin and the page will look weird because the donation banner goes over the message telling you that you need to log in to chat

backend/chat_messages/utility/chat_setup.py
positiveimpact: Changed it to logging, thanks for pointing that out

Issue #663, opened 2021-07-08
Fix "create idea" popup from the dashboard


piperchester:
The dashboard also still has one small thing not working: when you click "create idea" it doesn't actually open the popup where you can create an idea but just redirects you to the ideas page.

positiveimpact:
Closed by #855

PR #855, opened 2021-12-16
fix: 663 create idea popup

TheMeerCat:
#663

### DIFF
--- a/frontend/src/components/dashboard/Dashboard.js
+++ b/frontend/src/components/dashboard/Dashboard.js
@@ -206,7 +206,7 @@ export default function Dashboard({ allHubs, hubData, className, headline, locat
       const userOrgsFromServer = await getUserOrganizations(token, locale);
       setUserOrganizations(userOrgsFromServer || []);
     }
-  });
+  }, []);
 
   const { user, locale } = useContext(UserContext);
   const texts = getTexts({ page: "dashboard", locale: locale, user: user, location: location });
--- a/frontend/src/components/dashboard/Dashboard.js
+++ b/frontend/src/components/dashboard/Dashboard.js
@@ -255,7 +255,7 @@ export default function Dashboard({ allHubs, hubData, className, headline, locat
                   {
                     name: texts.create_idea,
                     // url_slug: `${getLocalePrefix(locale)}/${item.url_slug}/`,
-                    url_slug: "#ideas",
+                    url_slug: "#",
                     onClick: () => setCreateIdeaOpen(true),
                   },
                   {
--- a/frontend/src/components/dashboard/Dashboard.js
+++ b/frontend/src/components/dashboard/Dashboard.js
@@ -15,6 +15,7 @@ import AssignmentIcon from "@material-ui/icons/Assignment";
 import EmojiObjectsIcon from "@material-ui/icons/EmojiObjects";
 import GroupAddIcon from "@material-ui/icons/GroupAdd";
 import React, { useContext, useEffect, useRef, useState } from "react";
+import Cookies from "universal-cookie";
 import { getLocalePrefix } from "../../../public/lib/apiOperations";
 import getTexts from "../../../public/texts/texts";
 import theme from "../../themes/theme";
@@ -23,7 +24,6 @@ import UserContext from "../context/UserContext";
 import UserImage from "./UserImage";
 import CreateIdeaDialog from "../ideas/createIdea/CreateIdeaDialog";
 import { getUserOrganizations } from "../../../public/lib/organizationOperations";
-import Cookies from "universal-cookie";
 
 const useStyles = makeStyles((theme) => {
   return {
--- a/frontend/src/components/dashboard/Dashboard.js
+++ b/frontend/src/components/dashboard/Dashboard.js
@@ -255,7 +255,7 @@ export default function Dashboard({ allHubs, hubData, className, headline, locat
                   {
                     name: texts.create_idea,
                     // url_slug: `${getLocalePrefix(locale)}/${item.url_slug}/`,
-                    url_slug: "",
+                    url_slug: "#ideas",
                     onClick: () => setCreateIdeaOpen(true),
                   },
                   {
--- a/frontend/src/components/dashboard/Dashboard.js
+++ b/frontend/src/components/dashboard/Dashboard.js
@@ -14,13 +14,16 @@ import ArrowDropDownIcon from "@material-ui/icons/ArrowDropDown";
 import AssignmentIcon from "@material-ui/icons/Assignment";
 import EmojiObjectsIcon from "@material-ui/icons/EmojiObjects";
 import GroupAddIcon from "@material-ui/icons/GroupAdd";
-import React, { useContext, useRef, useState } from "react";
+import React, { useContext, useEffect, useRef, useState } from "react";
 import { getLocalePrefix } from "../../../public/lib/apiOperations";
 import getTexts from "../../../public/texts/texts";
 import theme from "../../themes/theme";
 import OpenClimateMatchButton from "../climateMatch/OpenClimateMatchButton";
 import UserContext from "../context/UserContext";
 import UserImage from "./UserImage";
+import CreateIdeaDialog from "../ideas/createIdea/CreateIdeaDialog";
+import { getUserOrganizations } from "../../../public/lib/organizationOperations";
+import Cookies from "universal-cookie";
 
 const useStyles = makeStyles((theme) => {
   return {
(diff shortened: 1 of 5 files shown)


frontend/src/components/dashboard/Dashboard.js
```diff
@@ -14,13 +14,16 @@ import ArrowDropDownIcon from "@material-ui/icons/ArrowDropDown";
...
+import React, { useContext, useEffect, useRef, useState } from "react";
 import { getLocalePrefix } from "../../../public/lib/apiOperations";
 import getTexts from "../../../public/texts/texts";
 import theme from "../../themes/theme";
 import OpenClimateMatchButton from "../climateMatch/OpenClimateMatchButton";
 import UserContext from "../context/UserContext";
 import UserImage from "./UserImage";
+import CreateIdeaDialog from "../ideas/createIdea/CreateIdeaDialog";
+import { getUserOrganizations } from "../../../public/lib/organizationOperations";
+import Cookies from "universal-cookie";
```
piperchester, line 26: Nit: do you mind colocating this 3rd party import to the others, up with line 18? And then delineating relative imports and 3rd party with a newline? I find this helps aid with readability some :-)

frontend/src/components/hub/HubContent.js
```diff
@@ -133,7 +135,13 @@ export default function HubContent({
             <div className={`${user && classes.dashboardAndStatboxWrapper}`}>
               <div>
                 {user ? (
-                  <Dashboard location={location} headline={headline} hubUrl={hubUrl} />
+                  <Dashboard
+                    allHubs={allHubs}
+                    hubData={hubData}
+                    location={location}
+                    headline={headline}
+                    hubUrl={hubUrl}
```
piperchester, line 143: Nit: do you mind alphabetizing these? If you're using VS Code, this is really easy by pressing `cmd+shift+p` and then typing "sort".

frontend/src/components/dashboard/Dashboard.js
```diff
@@ -161,16 +164,24 @@ const DropDownList = ({ buttonRef, handleOpen, handleClose, items, open }) => {
...
   return (
     <Popper open={open} anchorEl={buttonRef.current}>
       <Paper onMouseEnter={handleOpen} onMouseLeave={handleClose} className={classes.menu}>
         <MenuList>
           {items?.map((item) => (
-            <Link key={item.url_slug} href={item.url_slug} onClick={handleClick}>
+            <Link
+              key={item.url_slug}
+              href={item.url_slug}
+              onClick={() => handleClick(item.onClick)}
```
piperchester, line 183: Is there a benefit here to using inline arrow functions over the bound handler?

marketajancova: I am not sure whether I understand what you mean but `onClick` requires `Function` type and doing `onClick={() => handleClick(item.onClick)}` would just call `handleClick(item.onClick)` during rendering and insert the output as the parameter. I could insert `onClick={handleClick}` without the arrow but then I would not be able to pass the `item.onClick` parameter. Do you mean some other way? :-)

frontend/src/components/hub/HubContent.js
marketajancova: Sure, no problem. Personally, I don't alphabetize shortcuts but if it is our standard, I will definitely follow it. :-)

marketajancova: Hmmm I've checked it in code and we have "hubUrl" everywhere. Do we really want to do alphabetize it?

frontend/src/components/dashboard/Dashboard.js
marketajancova: Sure, no problem. :-)

marketajancova:
Thank you for your review/comments @piperchester :-)

frontend/src/components/hub/HubContent.js
positiveimpact: I feel like doing this manually doesn't make too much sense at this point since we've already not been doing it in the rest of the codebase. Instead we should probably update our configuration for `prettier` so this gets done automatically.

frontend/src/components/dashboard/Dashboard.js
```diff
@@ -182,9 +193,21 @@ const DropDownList = ({ buttonRef, handleOpen, handleClose, items, open }) => {
...
+  const [isCreateIdeaOpen, setCreateIdeaOpen] = useState(false);
+  const token = new Cookies().get("token");
+
+  useEffect(async function () {
+    if (userOrganizations === null) {
+      setUserOrganizations("");
+      const userOrgsFromServer = await getUserOrganizations(token, locale);
+      setUserOrganizations(userOrgsFromServer || []);
+    }
+  });
```
positiveimpact, line 209: Does this have to run multiple times? Because if it only has to run once, I'd add the following suggestion:
```suggestion
  useEffect(async function () {
    if (userOrganizations === null) {
      setUserOrganizations("");
      const userOrgsFromServer = await getUserOrganizations(token, locale);
      setUserOrganizations(userOrgsFromServer || []);
    }
  }, []);
```

frontend/src/components/dashboard/Dashboard.js
```diff
@@ -161,16 +164,24 @@ const DropDownList = ({ buttonRef, handleOpen, handleClose, items, open }) => {
...
 
   return (
     <Popper open={open} anchorEl={buttonRef.current}>
       <Paper onMouseEnter={handleOpen} onMouseLeave={handleClose} className={classes.menu}>
         <MenuList>
           {items?.map((item) => (
-            <Link key={item.url_slug} href={item.url_slug} onClick={handleClick}>
+            <Link
+              key={item.url_slug}
+              href={item.url_slug}
```
positiveimpact, line 182: The href here needs to be removed if item.url_slug is empty. Otherwise, it will open the idea popup for a second and then just refresh the current page

positiveimpact (commented):
LGTM except my comments. One thing that would be ideal would be if it also switched tab to the ideas tab automatically.

frontend/src/components/dashboard/Dashboard.js
marketajancova: For the idea popup, the href is empty string. I tested it and it does not redirect you anywhere

frontend/src/components/dashboard/Dashboard.js
marketajancova: True, I fixed it

frontend/src/components/dashboard/Dashboard.js
piperchester: Nope that makes sense! Thanks for explaining :-)

piperchester (approved):
Nice work!

TheMeerCat:
> LGTM except my comments. One thing that would be ideal would be if it also switched tab to the ideas tab automatically.

I changed it based on your suggestion. Unfortunately, the redirect at the end is a bit problematic as we have to load "ideas" data. Do you think we are ok with refreshing the page at the end?

positiveimpact:
> > LGTM except my comments. One thing that would be ideal would be if it also switched tab to the ideas tab automatically.
> 
> I changed it based on your suggestion. Unfortunately, the redirect at the end is a bit problematic as we have to load "ideas" data. Do you think we are ok with refreshing the page at the end?

I'd merge it without the tab switch for now. It would be possible to switch to the tab and load the data in the background without refreshing. But since tab switching is handled in BrowseContent that would be a lot of refactoring to be able to trigger it without a redirect from the dashboard component. And after creating an idea you're redirected to the ideas tab anyways so from my side this can be merged as is.

marketajancova:
> > > LGTM except my comments. One thing that would be ideal would be if it also switched tab to the ideas tab automatically.
> > 
> > 
> > I changed it based on your suggestion. Unfortunately, the redirect at the end is a bit problematic as we have to load "ideas" data. Do you think we are ok with refreshing the page at the end?
> 
> I'd merge it without the tab switch for now. It would be possible to switch to the tab and load the data in the background without refreshing. But since tab switching is handled in BrowseContent that would be a lot of refactoring to be able to trigger it without a redirect from the dashboard component. And after creating an idea you're redirected to the ideas tab anyways so from my side this can be merged as is.

Yeh, I do agree. So I reverted it. All good now?

positiveimpact:
> > > > LGTM except my comments. One thing that would be ideal would be if it also switched tab to the ideas tab automatically.
> > > 
> > > 
> > > I changed it based on your suggestion. Unfortunately, the redirect at the end is a bit problematic as we have to load "ideas" data. Do you think we are ok with refreshing the page at the end?
> > 
> > 
> > I'd merge it without the tab switch for now. It would be possible to switch to the tab and load the data in the background without refreshing. But since tab switching is handled in BrowseContent that would be a lot of refactoring to be able to trigger it without a redirect from the dashboard component. And after creating an idea you're redirected to the ideas tab anyways so from my side this can be merged as is.
> 
> Yeh, I do agree. So I reverted it. All good now?

:+1:

PR #792, opened 2021-12-01
Feat: SocialMediaShareButton for organizations and ideas

verena-e-p:
SocialMediaShareButton and SocialMediaShareDialog (see PR #772) are now generalized and implemented on
* organization page
* idea card

**To Do: Check whether the database entry for the native share dialog works correctly.**

Closes #481

## Before landing

1. PR has meaningful title
1. `yarn lint` passes
1. `yarn format` passes

PR #857, opened 2021-12-17
Fix: Select the correct language for texts (SocialMediaShareButton on the organization page)

verena-e-p:
Follow up to #792 

The organization texts were missing the locale prop and therefore they were always in English. Now the locale is added and the correct language is displayed.


piperchester (approved):

ddhanesha:
Since it approved I am going to merge this PR to reduce clutter in our PR list.

PR #858, opened 2021-12-17
Serving WebP images in lieu of JPEG when applicable

FnayouSeif:
## Description

This PR introduces a new feature that converts uploaded images to webp. For every image that is uploaded, a file will be created normally with a `.jpeg` extension. In addition, a new file with the same name will be created with a `.webp` extension.

As webp is not supported by all browser versions, we need to provide .jpeg in case of incompatibility. The logic will be handled on the node side. It servers webp or jpeg based on the compatibility check.  

##TODO 

1. Add a function in imageOperations.js that detects webp browser compatibility. More info is found here https://developers.google.com/speed/webp/faq#how_can_i_detect_browser_support_for_webp. We have 2 options: Client side and Server side detection (using headers). 
2. Write a script that will convert already existing images in Azure Storage to their webp equivalent.

## Test plan

_Include relevant test configuration details for the reviewer(s), and steps to test and verify new feature._

## Before landing

1. PR has meaningful title
1. `yarn lint` passes
1. `yarn format` passes

### DIFF
--- a/frontend/public/lib/imageOperations.js
+++ b/frontend/public/lib/imageOperations.js
@@ -2,7 +2,15 @@ const DEVELOPMENT = ["development", "develop", "test"].includes(process.env.ENVI
 import imageCompression from "browser-image-compression";
 
 export function getImageUrl(url) {
+
+  var IsWebPCompatibleBrowser = false;
   if (!url) return;
+
+
+  if (IsWebPCompatibleBrowser===true) {
+    url = url.replace('.jpeg','.webp')
+  }
+  console.log(url);
   if (DEVELOPMENT && !url.includes("http:") && !url.includes("https:")) {
     return process.env.API_URL + url;
   } else {
--- a/backend/climateconnect_main/utility/general.py
+++ b/backend/climateconnect_main/utility/general.py
@@ -1,7 +1,6 @@
 # python standard lib
 import base64
 import io
-from os import path
 import secrets
 
 from django.core.files.base import ContentFile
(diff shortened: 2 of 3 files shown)


frontend/public/lib/imageOperations.js
```diff
@@ -2,7 +2,15 @@ const DEVELOPMENT = ["development", "develop", "test"].includes(process.env.ENVI
...
 export function getImageUrl(url) {
+
+  var IsWebPCompatibleBrowser = false;
   if (!url) return;
+
+
+  if (IsWebPCompatibleBrowser===true) {
+    url = url.replace('.jpeg','.webp')
+  }
+  console.log(url);
```
piperchester, line 13: Nit: just some minor rewrite following more modern JS best practices!
```suggestion
  let isWebPCompatibleBrowser = false;
  if (!url) return;


  if (isWebPCompatibleBrowser) {
    url = url.replace('.jpeg', '.webp');
  }
```

piperchester (commented):
💯 excited for this improvement!

positiveimpact:
Saving the images as webp works perfectly for me :ok_hand: 
I think client-side detection makes most sense since the client just needs to add the correct ending to the URL retrieved from the server.

positiveimpact:
Hey, I was thinking whether we could already merge part of this but I think we first need client-side detection in order to do that, right? Do you need any support on the frontend part?

ddhanesha:
@FnayouSeif @positiveimpact what is the state of this branch? Anything we can do to move this forward!

ddhanesha:
@positiveimpact what is the state of this PR?

PR #841, opened 2021-12-14
fix: [838] header size

TheMeerCat:
#838

PR #859, opened 2021-12-17
Fix: Refactor ContactCreatorButton Component

verena-e-p:
## Description

Refactoring of the ContactCreatorButton component which was introduced in PR #717.

* introduced new properties to control the button, instead of returning different buttons depending on the screen size

### DIFF
--- a/frontend/src/components/climateMatchResults/ClimateMatchSuggestionInfo.js
+++ b/frontend/src/components/climateMatchResults/ClimateMatchSuggestionInfo.js
@@ -174,7 +174,7 @@ export default function ClimateMatchSuggestionInfo({
             withStartIcon={!isNarrowScreen}
             withAvatar={!isNarrowScreen}
             customWidth={!isNarrowScreen ? 220 : 300}
-            withCard={true}
+            withInfoCard={true}
             collapsable={!isNarrowScreen}
           />
         )}
--- a/frontend/src/components/climateMatchResults/ClimateMatchSuggestionInfo.js
+++ b/frontend/src/components/climateMatchResults/ClimateMatchSuggestionInfo.js
@@ -129,6 +129,7 @@ export default function ClimateMatchSuggestionInfo({
 }) {
   const classes = useStyles({ displayContactButton: displayContactButton });
   const suggestionInfoUnderImage = isInSlider || useMediaQuery("(max-width:1525px)");
+  const isSmallScreen = useMediaQuery((theme) => theme.breakpoints.down("sm"));
   const isNarrowScreen = useMediaQuery((theme) => theme.breakpoints.down("xs"));
   return (
     <div className={classes.wrapper}>
@@ -174,6 +175,7 @@ export default function ClimateMatchSuggestionInfo({
             large={isNarrowScreen}
             withStartIcon={!isNarrowScreen}
             withAvatar={!isNarrowScreen}
+            customWidth={isSmallScreen ? 220 : 300}
           />
         )}
       </div>
--- a/frontend/src/components/climateMatchResults/ClimateMatchSuggestionInfo.js
+++ b/frontend/src/components/climateMatchResults/ClimateMatchSuggestionInfo.js
@@ -172,6 +172,8 @@ export default function ClimateMatchSuggestionInfo({
             explanationBackground={background}
             className={classes.contactCreatorButton}
             large={isNarrowScreen}
+            withStartIcon={!isNarrowScreen}
+            withAvatar={!isNarrowScreen}
           />
         )}
       </div>
--- a/frontend/src/components/climateMatchResults/ClimateMatchSuggestionInfo.js
+++ b/frontend/src/components/climateMatchResults/ClimateMatchSuggestionInfo.js
@@ -129,7 +129,6 @@ export default function ClimateMatchSuggestionInfo({
 }) {
   const classes = useStyles({ displayContactButton: displayContactButton });
   const suggestionInfoUnderImage = isInSlider || useMediaQuery("(max-width:1525px)");
-  const isSmallScreen = useMediaQuery((theme) => theme.breakpoints.down("sm"));
   const isNarrowScreen = useMediaQuery((theme) => theme.breakpoints.down("xs"));
   return (
     <div className={classes.wrapper}>
@@ -175,8 +174,9 @@ export default function ClimateMatchSuggestionInfo({
             large={isNarrowScreen}
             withStartIcon={!isNarrowScreen}
             withAvatar={!isNarrowScreen}
-            customWidth={isSmallScreen ? 220 : 300}
+            customWidth={!isNarrowScreen ? 220 : 300}
             withCard={true}
+            collapsable={!isNarrowScreen}
           />
         )}
       </div>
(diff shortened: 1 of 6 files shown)


frontend/src/components/project/Buttons/ContactCreatorButton.js
```diff
@@ -99,85 +98,52 @@ export default function ContactCreatorButton({
...
-        <div className={classes.buttonWithCollapseContainer}>
-          <Collapse in={hoveringButton} timeout={550}>
+  return (
+    <div
+      className={withInfoCard ? `${classes.root} ${className}` : null}
+      onMouseEnter={handleMouseEnter}
+      onMouseLeave={handleMouseLeave}
+      onClick={handleClickContact}
+    >
+      <div className={collapsable ? classes.collapsableContainer : classes.nonCollapsableContainer}>
```
marketajancova, line 108: `nonCollapsableContainer` is a subset of `collapsableContainer`. Maybe I would extract this 'width' property to a separate class, which is always present, and add the  `collapsableContainer` as an optional second class (if required)

frontend/src/components/project/Buttons/ContactCreatorButton.js
```diff
@@ -8,12 +8,11 @@ import theme from "../../../themes/theme";
 import UserContext from "../../context/UserContext";
 
 const useStyles = makeStyles({
-  root: {
+  root: (props) => ({
     position: "relative",
-    height: 40,
-    width: 200,
+    width: props.customWidth ? props.customWidth : "auto",
```
marketajancova, line 13: Please, check whether this property is really needed in all these places (I would expect we needed on root and that's it). Additionally, I would maybe prefer having 200px as default (like it was before) to have under control how it behaves on non-standard screens.

frontend/src/components/climateMatchResults/ClimateMatchSuggestionInfo.js
```diff
@@ -171,7 +171,11 @@ export default function ClimateMatchSuggestionInfo({
             handleClickContact={handleClickContact}
             explanationBackground={background}
             className={classes.contactCreatorButton}
-            large={isNarrowScreen}
+            withStartIcon={!isNarrowScreen}
+            withAvatar={!isNarrowScreen}
```
marketajancova, line 175: Some of these properties seem to be always the same (at least withAvatar and withStartIcon), maybe we can consider merging them. I understand the idea of reusability but I think it's good to avoid including more props than needed.

marketajancova (commented):
I had some minor comments but I don't require changing any of them (only check the button width, please). You did really nice job here. :-)

frontend/src/components/climateMatchResults/ClimateMatchSuggestionInfo.js
verena-e-p: Good point! I merged withAvatar and withStartIcon into one prop (withIcons) :)

frontend/src/components/project/Buttons/ContactCreatorButton.js
verena-e-p: I added a default width but I could not get rid of the width property in the child classes. The only way (I found) to make sure the button and card have the same width and the elements around the button know its width was to define the width within all three classes. But simply setting it to 100% in the child classes seems to be enough so I did that.

frontend/src/components/project/Buttons/ContactCreatorButton.js
verena-e-p: Not sure if what I did is what you meant, see [commit](https://github.com/climateconnect/climateconnect/pull/859/commits/552213ec8dd98fce42c8f7236afc09643167ccf5)

verena-e-p:
One additional issue I found is that I don't know how to control what happens to the card if there is a word that is so long that it doesn't fit on the card. Right now the card just gets cut off (left and right), I could not make its contents shrink. Maybe you have a tip for this? And thanks a lot for your feedback once again :slightly_smiling_face:

frontend/src/components/project/Buttons/ContactCreatorButton.js
```diff
@@ -99,85 +95,58 @@ export default function ContactCreatorButton({
...
+      onMouseEnter={handleMouseEnter}
+      onMouseLeave={handleMouseLeave}
+      onClick={handleClickContact}
+    >
       <div
-        className={`${classes.root} ${className}`}
-        onMouseEnter={handleMouseEnter}
-        onMouseLeave={handleMouseLeave}
-        onClick={handleClickContact}
+        className={
```
marketajancova, line 106: Yes, it's what I meant :-). I would maybe simplify this:

```
collapsable  ? `${classes.collapsableContainer} ${classes.applyCustomCardWidth}` : classes.applyCustomCardWidth
```
using:
```
`${collapse ? classes.collapsableContainer : ''} ${classes.applyCustomCardWidth}`
```
but that's only my preference, you can keep it as it is. :-)

frontend/src/components/project/Buttons/ContactCreatorButton.js
```diff
@@ -33,16 +32,17 @@ const useStyles = makeStyles({
...
     fontWeight: "bold",
   },
-  buttonWithCollapseContainer: {
+  collapsableContainer: {
     position: "absolute",
     bottom: 0,
     right: 0,
   },
+  applyCustomCardWidth: {
+    width: "100%",
```
marketajancova, line 41: If there is a reason why you need these width properties, you can ofc keep them! :-) As I was doing the review, it seemed that some of them were redundant but it's fully ok to have them. Thank you for checking it. :-)

frontend/src/components/project/Buttons/ContactCreatorButton.js
```diff
@@ -69,14 +61,18 @@ export default function ContactCreatorButton({
   creator,
   contactProjectCreatorButtonRef,
   handleClickContact,
-  tiny,
-  small,
-  isFixed,
-  large,
   contentType,
   explanationBackground,
+  withIcons,
```
marketajancova, line 66: Looks much simpler now. Thanks for doing it! :-)

marketajancova (approved):
I will already approve, I don't have any additional points here. Nice work and thanks for considering my comments. :-)

frontend/src/components/project/Buttons/ContactCreatorButton.js
verena-e-p: It looks cleaner the way you suggested, I changed it to that :)

TheMeerCat:
@verena-e-p  sorry for late reply! This is quite problematic topic, it depends on circumstances and there is no ultimate-guide. If you have single-line text that is fix-sized (i.e. it has at least "maxWidth"), you can do it similarly as I did it here: https://github.com/climateconnect/climateconnect/pull/841/files, I solved the problem for Title names in organization cards.
Problem comes if you have multiple lines of text in an element of undefined height and you want to prevent the height expansion. Then you have to be very creative :-D, for example, what I did here: https://github.com/climateconnect/climateconnect/pull/770/files#diff-924b766e723b479b50fe184ec24d04cafab22edbee235e650f5094826fa96c84R79, I tried to use a bit of maths to estimate the number of lines that can fit to the div without overflowing based on presence or non-presence of other elements (e.g. you have a card with optional header, location, and description, respectively; and you want to prevent height expansion, so you check whether you have header and location and based on this you can estimate how many lines of description fit into the box). But this is kind of crazy already.
I guess what you want to do is preventing width expansion, which can be done with `text-overflow: ellipsis`, such as here: https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/ (this problem is called "string truncation", it took me a while before I found out how to name it :-D )

TheMeerCat (approved):
As always, I forgot to swich my account and approved it with the work account. Oh well... So once more. :-D

Issue #860, opened 2021-12-19
Investigate + optimize bundle size in order to improve initial load time


TheMeerCat:
Bundle size for some pages is quite big. For example, '/browse' page has around 445kb on the first load. This number gets lower when the user browses through the website but we should aim to make it as fast as possible, when the user reaches the websibe for the first time. 

**There are multiple techniques we can use:**
- babel plugins – find "huge" dependecies and search for babel plugins that can optimize their size
- splitting big components – avoiding having huge files may generate smaller bundles (we aim having more small bundles than fewer big bundles)
- dynamic imports – there are some components importing lots of stuff; this could be imported in runtime dynamically. That means the specific component is imported only when it is actually supposed to be rendered. E.g., this is very handy for tabs.
- ... be creative

PR #861, opened 2021-12-19
Fix/814 browse optimizations

TheMeerCat:
#860 

I have played around and identified a few problems:
- /browse bundle size is too big because we import lots of components (projects, hubs, etc.). I played around with dynamic imports. If my understanding is correct, tabs that are not shown should not be imported (this happens later – when user switches the tab). Bundle size of /browse was reduced from cca 450 to 295kb. 
- We use lodash and import the whole library, which is 500kb. I used "lodash" babel plugin, which picks only necessary parts (during build) and packs them to our output files. The result is that lodash is reduced to 0 and related files have ca. 1kb more.
- I added 'lazy load' property to images, which means they are downloaded on demand – when they are about to be shown on the visible screen
- I used `useMemo` for our texts fetching. This means that we fetch it once (during the first rendering) and cache it. You can understand its behavior similarly as `useEffect` hook. `useMemo` then uses memoization and values are not re-calculated during each re-rendering. This is very handy for complex calculations.
- I fixed "useStyles" and "withStyles" to be imported from the correct place (not core but core/styles), which is required for further material-ui optimizations in babel. Unfortunately, the official babel plugin for material ui did not optimize much for us so I haven't included it for now.
- There is a new plugin that visualizes bundles for both client and server rendering. You can use it with `ANALYZE=true yarn build` and it opens both graphs in your browser (it lasts up to minute or two before it computes them)

Big topic are our text translations: we have a file that imports all texts to one place. Dev experience is cool but this approach might cause performance problems once we include more translations or support more languages. What happes is that when you visit a page, it downloads all translations (also these that are fully irrelevant). As a result, there are hundreds of textlines that we don't need and are downloaded everywhere). We could fix it using dynamic imports, which would be probably the most elegant way, I started with PoC, then realized how much work it is to fix it everywhere and gave up :-D. Here is the idea, how the new texts file would like: https://github.com/climateconnect/climateconnect/compare/fix/814-text-optimization-dont-merge#diff-39a10bb64c37fc0811e27b005e3c1d96c58ec5773586b2acff6a1de28249de85 note that we do not import any text-file directly. When we do it, it reduces bundle size to half. Worth it? I don't know.


ddhanesha:
Please resolve conflicts with the master branch.

positiveimpact (commented):
Here is a partial review I started yesterday but then my computer crashed and I think it's a hardware problem... gonna get that checked while I'm on holiday. This is a really exiting change :) We just need to make sure that all pages are still working after this, e.g. making sure that hooks can be used in all the places where we are using them.

piperchester:
This is some fantastic research @TheMeerCat! Nice work on digging into some area for perf optimizations. Will follow along more in #866.

PR #866, opened 2021-12-27
feat: 860 browse optimizations: add useMemo and bundle-analyzer

TheMeerCat:
Fix of https://github.com/climateconnect/climateconnect/pull/861
We decided to go step-by-step and avoid huge/dangerous changes


frontend/.babelrc.js
```diff
@@ -0,0 +1,6 @@
+const plugins = ["lodash"];
+
+module.exports = {
+  plugins,
+  presets: ["next/babel"],
+};
```
piperchester, line 6: Nit: Nice! For my own understanding, what was the benefit of migrating to a `.babelrc.js`?

frontend/src/components/browse/BrowseContent.js
```diff
@@ -116,7 +119,7 @@ export default function BrowseContent({
     TYPES_BY_TAB_VALUE.push("ideas");
   }
   const { locale } = useContext(UserContext);
-  const texts = getTexts({ page: "general", locale: locale });
+  const texts = useMemo(() => getTexts({ page: "general", locale: locale }), [locale]);
```
piperchester, line 122: Nit: Good call with `useMemo`! Completely for my own curiosity I'm curious if we were able to measure a perceivable difference in text fetching. I think we should probably be using `useMemo` more widely.

frontend/src/components/browse/BrowseContent.js
```diff
@@ -491,106 +495,108 @@ export default function BrowseContent({
...
-              organizations={state.items.organizations}
-              parentHandlesGridItems
-            />
-          </TabContentWrapper>
-          {!hideMembers && (
-            <TabContentWrapper type={"members"} {...tabContentWrapperProps}>
-              <ProfilePreviews
-                hasMore={state.hasMore.members}
-                loadFunc={() => handleLoadMoreData("members")}
+      <Suspense fallback={<LoadingSpinner isLoading={true} />}>
```
piperchester, line 498: Nice! I think this is our first usage of `<Suspense />` in the project. I personally haven't used it yet. Is it working well so far?

piperchester (approved):
This looks great! Few micro suggestions but otherwise no blockers from me :) 
1. Do you mind renaming the PR title to something that describes the changes? Maybe "feat: add useMemo and bundle-analyzer" or something similar. I think this'll help if we ever need to look back in Git history and find this PR after we squash and merge.
2. What do you think about adding a simple entry point to the analyze tool in `package.json` and in the `README.md`? Like an npm script, `analyze` or something similar? 
3. Have you ran [`yarn-deduplicate`](https://www.npmjs.com/package/yarn-deduplicate) recently? I'm curious if we can save on download time by reducing some duplicate dependencies. I've been meaning to add this to one of our CI checks.

frontend/.babelrc.js
TheMeerCat: https://babeljs.io/docs/en/configuration#javascript-configuration-files
It extends the json one. Not a huge benefite but as far as I know, it's the more standardized way :-)

frontend/src/components/browse/BrowseContent.js
TheMeerCat: Definitely! Well, without it, these texts are computed with each re-render, which sounds a bit creepy. :-)
It's dangerous to try to refactor everything now but maybe it's something we should keep in mind with future changes.

frontend/src/components/browse/BrowseContent.js
TheMeerCat: Suspense is there because of the lazy loading. We fetch data asynchronously and we need some substitution while the content is being fetched. I am also wondering how much it will be faster in production :-)

TheMeerCat:
> This looks great! Few micro suggestions but otherwise no blockers from me :)
> 
>     1. Do you mind renaming the PR title to something that describes the changes? Maybe "feat: add useMemo and bundle-analyzer" or something similar. I think this'll help if we ever need to look back in Git history and find this PR after we squash and merge.
> 
>     2. What do you think about adding a simple entry point to the analyze tool in `package.json` and in the `README.md`? Like an npm script, `analyze` or something similar?
> 
>     3. Have you ran [`yarn-deduplicate`](https://www.npmjs.com/package/yarn-deduplicate) recently? I'm curious if we can save on download time by reducing some duplicate dependencies. I've been meaning to add this to one of our CI checks.

Thanks @piperchester for all the comments. :-)
1. Will do!
2. Good idea, I added it
3. Nope but I did it now :-)

positiveimpact (commented):
Hey, sorry for the super late feedback! Overall the changes look good to me but the user experience when loading the site (that the filter bar appears later) but especially when changing tabs (the filter bar and tab overview disappear for a second) isn't ideal in my opinion. Once this is resolved the PR would be ready for merge. Maybe it could make sense to not lazy-load the filter bar and just lazy-load the tab content files (ProjectContent etc.)

frontend/src/components/browse/BrowseContent.js
```diff
@@ -491,106 +495,108 @@ export default function BrowseContent({
...
+          <FilterSection
+            filtersExpanded={isMobileScreen ? filtersExandedOnMobile : filtersExpanded}
+            onSubmit={handleSearchSubmit}
+            setFiltersExpanded={isMobileScreen ? setFiltersExpandedOnMobile : setFiltersExpanded}
+            type={TYPES_BY_TAB_VALUE[tabValue]}
+            customSearchBarLabels={customSearchBarLabels}
+            filterButtonRef={filterButtonRef}
+            searchValue={filters.search}
+            hideFilterButton={tabValue === TYPES_BY_TAB_VALUE.indexOf("ideas")}
+          />
```
positiveimpact, line 509: Instead of one big suspense object we could use 3 small ones: one here, one wrapping the content in the tab and one wrapping the tutorial. This could help with not always putting a loading animation over the whole content if we're only waiting for a small part of it.
```suggestion
<Suspense fallback={<></>}>
          <FilterSection
            filtersExpanded={isMobileScreen ? filtersExandedOnMobile : filtersExpanded}
            onSubmit={handleSearchSubmit}
            setFiltersExpanded={isMobileScreen ? setFiltersExpandedOnMobile : setFiltersExpanded}
            type={TYPES_BY_TAB_VALUE[tabValue]}
            customSearchBarLabels={customSearchBarLabels}
            filterButtonRef={filterButtonRef}
            searchValue={filters.search}
            hideFilterButton={tabValue === TYPES_BY_TAB_VALUE.indexOf("ideas")}
          />
          </Suspense>
```

positiveimpact:
I added a small change that still allows showing the loading spinner for the part under the tab selection because I think that's necessary for user experience. Merging now, let's see what the performance results are :)

PR #867, opened 2021-12-28
Fix: Position of ContactCreatorButton in ProjectOverview

verena-e-p:
## Description

Follow up to #859 
When refactoring I removed the height from the root container because it seemed duplicate to me. That introduced a visual bug for projects with likes or followers: The button aligns with the likes- / followers-link instead of with the other buttons. This PR fixes that by defining the height of the root container for the collapsable version of the ContactCreatorButton.

## Test plan

Check the button position of the ContactCreatorButton in ProjectOverview on a project which does have likes or followers. Without this fix, the positioning of the button is wrong.


TheMeerCat (approved):
LGTM :-)

PR #868, opened 2021-12-30
Fix: Tutorial: Switch tab to the one the referenced element is on (if necessary)

### LINKED ISSUE #520: Automatically switch to projects tab in tutorial for the step "The project cards!"
The tutorial contains a step "The project cards" that described how project cards work. This is pretty confusing if you are in a different tab than "projects". Therefore the tab should automatically switch to "projects" when going to that step in the tutorial

Screenshot from 2021-12-12 22-33-12

verena-e-p:
## Description
Closes #520 :
This fix makes sure that the correct tab is opened if an element that is referenced by a tutorial step is located on a tab 

* Additional fix: The placeholder text had a different size than the actual text. Therefore the text did sometimes overlap with the buttons (eg. last step of tutorial on browse page when logged in). Now the sizes are equal to keep the text from overlapping.


frontend/src/components/tutorial/Tutorial.js
```diff
@@ -104,6 +111,18 @@ export default function Tutorial({ fixedPosition, pointerRefs, nextStepTriggered
     [step]
   );
 
+  const handleSetStep = (nextStep) => {
+    const tabLocationHash = tutorialSteps[nextStep]?.tabOfRef;
+    if (tabLocationHash && tabLocationHash !== window.location.hash) {
+      handleTabChange(null, typesByTabValue.indexOf(tabLocationHash.replace("#", "")));
+      setTimeout(() => {
```
piperchester, line 118: Guessing we needed to wait a moment before updating the step? WDYT about adding a one-line comment highlighting that?

frontend/src/components/tutorial/Tutorial.js
```diff
@@ -33,7 +33,14 @@ const useStyles = makeStyles((theme) => ({
...
 }));
 
-export default function Tutorial({ fixedPosition, pointerRefs, nextStepTriggeredBy, hubName }) {
+export default function Tutorial({
+  fixedPosition,
+  pointerRefs,
+  nextStepTriggeredBy,
+  hubName,
+  handleTabChange,
+  typesByTabValue,
```
piperchester, line 42: Nit: WDYT about sorting imports and props (when possible)? I find it helps with readability and thereby maintainability.
```suggestion
  fixedPosition,
  hubName,
  handleTabChange,
  nextStepTriggeredBy,
  pointerRefs,
  typesByTabValue,
```

VS Code makes it really easy to do with `Cmd+shift+p` and then typing "sort" 🎉

frontend/src/components/tutorial/TutorialStep.js
```diff
@@ -103,6 +103,7 @@ const useStyles = makeStyles((theme) => ({
   },
   textPlaceholder: {
     visibility: "hidden",
+    fontSize: 17.5,
```
piperchester, line 106: Thanks for getting this fix too!

piperchester (approved):
This looks great! Do you mind attaching a small GIF or clip of the fix to the PR?

frontend/src/components/tutorial/Tutorial.js
marketajancova: Considering [line 99](https://github.com/climateconnect/climateconnect/pull/868/files#diff-d9f7a62367d8875cc5d090ea51e3f3cb201f93c7186d562a5e5e4a5da390225aR99), I would say this should not be needed anymore 🤔 . Have you tried it without the timeout?

frontend/src/components/tutorial/Tutorial.js
```diff
@@ -138,7 +157,7 @@ export default function Tutorial({ fixedPosition, pointerRefs, nextStepTriggered
       sameSite: "lax",
     });
     const nextStep = getTutorialStepFromCookie(tutorialSteps, newCookieValue, user);
-    setStep(!isStartingStep || nextStep > 0 ? nextStep : 1);
+    handleSetStep(!isStartingStep || nextStep > 0 ? nextStep : 1);
```
marketajancova, line 160: I think this could be replaced with `handleSetStep(nextStep ? nextStep : +isStartingStep);`? :-)

marketajancova: (`+` is an operator that converts `boolean` to `integer`)

frontend/src/components/tutorial/Tutorial.js
marketajancova: If we are motivated, we might want to optimize [line 109](https://github.com/climateconnect/climateconnect/pull/868/files#diff-d9f7a62367d8875cc5d090ea51e3f3cb201f93c7186d562a5e5e4a5da390225aR109) to avoid that long waiting (set it to 50ms and if the element is not yet there after a certain period of time, wait another 50ms, etc.).
But for now, it's fully fine!

marketajancova (approved):
I tried it locally and it looks great! I was thinking about whether we could do it without including a new property but I think it would be too complicated and you did it in the most readable way. So I don't have any additional points. :-)
Please, verify whether we can avoid the additional timeout and rely on the timeout provided in useEffect (it should be that case). 
Great work!

frontend/src/components/tutorial/TutorialStep.js
verena-e-p: :slightly_smiling_face:

frontend/src/components/tutorial/Tutorial.js
verena-e-p: This is good to know! :slightly_smiling_face:
I'm just not sure if we can use it here because `getTutorialStepFromCookie(tutorialSteps, newCookieValue, user)` can return -1 in some cases. In that case the expression `!isStartingStep || nextStep > 0 ? nextStep : 1` would become 1 and the expression you suggested -1. Since setting the step to -1 would close the tutorial it is maybe safer as it is. Or should I rather check if the return value of -1 will actually occur at this point?

frontend/src/components/tutorial/Tutorial.js
verena-e-p: > If we are motivated, we might want to optimize [line 109](https://github.com/climateconnect/climateconnect/pull/868/files#diff-d9f7a62367d8875cc5d090ea51e3f3cb201f93c7186d562a5e5e4a5da390225aR109) to avoid that long waiting (set it to 50ms and if the element is not yet there after a certain period of time, wait another 50ms, etc.).
> But for now, it's fully fine!

If I understand it correctly the element that the function checks for is either there or not. I guess the 500 ms are only there to make the scroll appear smoother.

verena-e-p: @marketajancova, I tried again and for me, it does not work reliably without the timeout. I think that is because the function with the timeout in line 99 doesn't keep the tutorial from being displayed, but I'm not sure, to be honest. So I only added the comment like @piperchester suggested and kept the timeout. I'm also happy to upload a clip of the fix. How can I do it without needing to ad the file to the repository?
Also thanks to both of you for your reviews! :slightly_smiling_face:

frontend/src/components/tutorial/Tutorial.js
TheMeerCat: Ah, didn't know about this case. All good then :-) keep it as it is to make sure we are not introducing a new bug

frontend/src/components/tutorial/Tutorial.js
TheMeerCat: That's super weird. But sure, go ahead and keep it :-)

TheMeerCat (approved):

frontend/src/components/tutorial/Tutorial.js
piperchester: > How can I do it without needing to ad the file to the repository? 
If you attach it as part of the comment on this open PR, that would be great! But no worries too :-)

verena-e-p: https://user-images.githubusercontent.com/90782656/147917783-7c8abd48-ee51-44fe-9f6d-39c5f91cfff6.mp4

Here is a short clip of the tab switch!

PR #871, opened 2022-01-04
Fix: Chat: Keep scrollbar scrolled down when writing messages

### LINKED ISSUE #459: Keep chat scrolled down when adding a new line by typing
When the Text input field extends during typing it overlays the previous chat message. Chat should be kept scrolled down when typing. 

### To reproduce:
1. Start a chat with a user or group
2. Write enough messages for a scrollbar to appear in the chat messages box
3. Write a message that's more than one line
4. The message will go over the previously written message

verena-e-p:
## Description

Closes #459 

When the user was typing a message the text field would sometimes overlap with the latest message. (When there were enough messages for the scrollbar to be displayed and the new message would be longer than one line.)

Now the scrollbar is kept fully scrolled down while typing a new message, but only if it was fully scrolled down when the user started typing.

## Test plan
1. Start a chat with a user or group
2. Write enough messages for a scrollbar to appear in the chat messages box
3. Make sure the scrollbar is scrolled down fully
4. Write a message that's more than one line, but do not send it: The latest message should not be covered by the text of the message you wrote
6. Repeat step 4 when the scrollbar is not fully scrolled down: Scrollposition should not change, messages at the bottom might be covered by the message you wrote


verena-e-p:
Just let me know if there is a better solution :)

positiveimpact (approved):
LGTM, great job! Merging.

PR #874, opened 2022-01-07
feature: 780 Contact local ambassador button

TheMeerCat:


backend/hubs/serializers/hub.py
```diff
@@ -68,6 +72,35 @@ def get_location(self, obj):
...
+    def get_local_ambassador(self, obj):
+            localAmbassador = LocalAmbassador.objects.filter(hub_id=obj.id)
+            if localAmbassador.exists():
+                return LocalAmbassadorSerializer(localAmbassador[0]).data
+            return None
+
+class LocalAmbassadorSerializer(serializers.ModelSerializer):
+    title = serializers.SerializerMethodField()
+    user = serializers.SerializerMethodField()
+    title_de = serializers.SerializerMethodField()
```
positiveimpact, line 84: Instead of returning both `title` and `title_de` I'd detect the language of the user here and just return the title attribute in the correct language similar to here: https://github.com/climateconnect/climateconnect/blob/bc7269b9551771e8ae95ff7caec8009bd112e512/backend/climateconnect_api/serializers/badge.py#L19-L20

frontend/public/texts/getHubTexts.js
```diff
@@ -120,6 +120,14 @@ export default function getHubTexts({ hubName }) {
       en: "Add Hubs (sectors) in which your organization is active",
       de: "Füge Hubs (Sektoren) hinzu, in denen deine Organisation aktiv ist",
     },
+    please_create_an_account_or_log_in_to_contact_the_ambassador: {
+      en: `Please sign up to contact ${localAmbassador?.first_name}.`,
+      de: `Bitte melde dich an, um ${localAmbassador?.first_name} zu kontaktieren.`,
+    },
+    contact_ambassador: {
+      en: "Contact Local Ambassador",
+      de: "Lokaler Klimavernetzer Kontaktieren",
```
positiveimpact, line 129: I'd make this generic, e.g. if the local ambassador is a woman it would be "Klimavernetzerin" in german. We could just pull this from the title
```suggestion
      en: `Contact ${localAmbassador.title}`,
      de: `${localAmbassador.title} kontaktieren`,
```

frontend/src/components/hub/ContactAmbassadorButton.js
```diff
@@ -0,0 +1,58 @@
...
+import getTexts from "../../../public/texts/texts";
+import Cookies from "universal-cookie";
+import ContactCreatorButtonInfo from "../communication/contactcreator/ContactCreatorButtonInfo";
+
+const useStyles = makeStyles(() => ({
+  root: {
+    zIndex: 1,
+    position: "fixed",
+    bottom: 0,
+    right: "200px",
```
positiveimpact, line 17: The positioning looks a bit weird on smaller screen sizes. Perhaps we could reuse the code from the contact project creator button and make it just not be all the way at the bottom when you're scrolled down all the way.
And we need a solution for mobile @tobiasrehm

positiveimpact (commented):
Great job on this so far! There are a few things that still need to be fine-tuned, especially the positioning of the button but overall ot looks great!

frontend/src/components/hub/ContactAmbassadorButton.js
TheMeerCat: I changed it to 10% so it deals better with different screen sizes :-)

TheMeerCat:
Awesome that I was wondering where my commit is and then realized I was working with the old branch and it took me like 20min to fix my stupidity... 👏 👏 😂 I should not work at the evening…

positiveimpact (commented):
LGTM overall now! But after looking at it on mobile again I feel like it will be quite annoying to always have that button there taking up ~15-20% of the screen. Therefore I'd just hide it on screen sizes smaller than xs and potentially think of a better solution in the future. Other than that I think it's ready to push!

ddhanesha:
@positiveimpact @TheMeerCat This PR has been inactive for a while. Is this something we can work on to move this forward?

PR #884, opened 2022-01-31
Redirect logged in users from '/' to '/browse' without memory leak

positiveimpact:
## Context

We recently removed our `getServerSideProps` function from `index.js`. All this function was handling was redirecting users from `/` to `/browse` if they were already logged in because the `/` page is just a landing page for new users. However this function seemed to be called over and over even after the redirect was done and was therefore causing a server-side memory leak.
We removed the redirect in a prior update to make sure the memory leak was caused by this which turned out to be the case. 

## Changes
- The described redirect is now happening through the "redirects" function in next.config.js. This seemed to be the easiest and most performant solution
- If the `token` cookie is set you will be redirected to `/browse`. If it's not set you will just go to `/`

## Test plan

- potentially create a production build for testing by using `yarn build` and then running it with `yarn start`
- Try to access `localhost:3000` while not logged in. You should not be redirected
- Try to access `localhost:3000`while logged in. You should be redirected to `/[your_locale]/browse`
- Try to change the language of the website and visit `localhost:3000` again while logged in. The redirect should point to the `browse` page in the correct language

## Before landing

1. PR has meaningful title


marketajancova (approved):

frontend/package.json
```diff
@@ -50,14 +50,14 @@
     "html-react-parser": "^1.4.0",
     "humanize-duration": "^3.22.0",
     "lodash": "^4.17.21",
-    "next": "^11.1.3",
+    "next": "^12.0.9",
```
piperchester, line 53: So looks like we needed v12 to handle redirects - any other breaking changes we should be aware of with the major upgrade? Didn't exhaustively look, but here's the upgrade guide: https://nextjs.org/docs/upgrading

piperchester (commented):
Nice!

frontend/package.json
positiveimpact: No breaking changes :) 
Actually we would have needed that for middleware redirects which we didn't opt for in the end but since there were no breaking changes anywhere I just left it in

positiveimpact: No changes that break out code to be precise. There are changes that could be breaking but we don't use any of the affected features :)

Issue #712, opened 2021-10-11
Send users regular automated update emails to show them what they missed


positiveimpact:
# Problem
Many users don't check the platform actively all the time. Therefore they may miss projects that are really interesting to them or may even completely forget about Climate Connect. I've heard even from people who really like the concept that they "forget to check". We, therefore, need a way to draw users on the platform again. 
# Suggested Fix
A regular (weekly, bi-weekly or monthly) automated update email with new interesting projects or updates to projects you follow could be really interesting. It would be much easier to show relevant updates if we had all of the following features:
- [Project update posts](https://github.com/climateconnect/climateconnect/issues/500) because then we show users updates on projects they follow which is definitely relevant
- [The new sign up process](https://github.com/climateconnect/climateconnect/issues/667) because with this process users are going to be asked for their [interests](https://github.com/climateconnect/climateconnect/issues/650) on sign-up which helps us recommend relevant projects
- The [Like Button](https://github.com/climateconnect/climateconnect/issues/513) which helps us to offer users an easy way to interact with projects and helps us find out what they are most interested in
- The [sidebar showing similar projects on the project page](https://github.com/climateconnect/climateconnect/issues/514) to be able to show users one new project but still let them see that other projects exist as well

tobiasrehm:
not on roadmap

PR #894, opened 2022-02-16
Feature/automated email

ddhanesha:
## Description
This PR includes feature that sends automated email reminders to users who have unread notifications that users have not checked in two days or more. 

## Test plan
You will need to have mailjet ENV variables in `.backend_env` If you don't have this information please reach out to me to get the credentials. 
To test this PR I would recommend **not to use** celery worker because if you have production data it will start sending email to all those users from your local machine. Instead follow these steps to verify that email is being sent properly. 

1. Open django shell and import following things
```
>>> from django.contrib.auth.models import User
>>> from climateconnect_api.models import UserNotification
>>> from climateconnect_api.utility.email_setup import send_email_reminder_for_unread_notifications
>>> user = User.objects.get(email="<email_address>")
>>> notifications = UserNotification.objects.filter(read_at__isnull=True, user=user)
>>> send_email_reminder_for_unread_notifications(user, notifications)
```
This should send an email to your account if you have everything setup correctly.


backend/climateconnect_api/utility/email_setup.py
```diff
@@ -272,3 +285,50 @@ def remove_contact_from_list(contact_id, list_id):
...
+
+def send_email_reminder_for_unread_notifications(
+    user: User,
+    user_notifications: List[UserNotification]
+):
+    total_notifications = user_notifications.count()
+    language_code = get_user_lang_code(user=user)
+    subject_by_language = {
+        "en": f"You have {total_notifications} unread notifications",
+        "de": f"Sie haben {total_notifications} ungelesene Benachrichtigungen"
```
positiveimpact, line 298: We usually use the less formal way to address people

```suggestion
        "de": f"Du hast {total_notifications} ungelesene Nachrichten"
```

backend/climateconnect_api/utility/email_setup.py
```diff
@@ -272,3 +285,50 @@ def remove_contact_from_list(contact_id, list_id):
...
+    }
+    subject = subject_by_language.get(language_code, "en")
+    website_link_by_language = {
+        "en": "https://climateconnect.earth/inbox",
+        "de": "https://climateconnect.earth/de/inbox"
+    }
+    website_link = website_link_by_language.get(language_code, "en")
+    email_text_by_language = {
+        "en": f"<p>Dear {user.first_name},</p><p>You have {total_notifications} new notifications. Please <a href={website_link}>click here</a> to check your inbox.</p><p>See you soon,</p><p>The Climate Connect Team</p>",  #NOQA
+        "de": f"<p>Liebe {user.first_name},</p><p>Sie haben {total_notifications} neue Benachrichtigungen. Bitte <a href={website_link}>klicken Sie hier</a>, um Ihren Posteingang zu überprüfen.</p><p>Bis bald,</p><p>Deine Climate Connect Team</p>"  #NOQA
```
positiveimpact, line 308: Is this advantageous over making an (unstyled) mailjet template? This seems a bit more difficult to work with to me.
Suggestion for the text:

```suggestion
        "en": f"<p>Dear {user.first_name},</p><p>You have {total_notifications} new messages. Please respond to the people who reached out because we can only limit climate change if we work together and exchange knowledge.</p> <p><a href={website_link}>Click here</a> to check your inbox.</p><p>See you soon,</p><p>The Climate Connect Team</p>",  #NOQA
        "de": f"<p>Hallo {user.first_name},</p><p>Du hast {total_notifications} ungelesene Nachrichten von anderen Klimaschützer*innen. Bitte beantworte die Nachrichten, denn gemeinsam und durch Zusammenarbeit und Wissensaustausch können wir das 1,5 Grad Ziel erreichen.</p><p><a href={website_link}>Klicke hier</a>, um deinen Posteingang anzusehen.</p><p>Bis bald,</p><p>Deine Climate Connect Team</p>"  #NOQA
```

backend/climateconnect_api/tasks.py
```diff
@@ -1,9 +1,50 @@
...
+
+        unread_user_notifications = UserNotification.objects.filter(
+            user_id=u_id, read_at__isnull=True
+        )
+
+        if unread_user_notifications.exists():
+            send_email_reminder_for_unread_notifications(
+                user=user,
+                user_notifications=unread_user_notifications
+            )
```
positiveimpact, line 50: Here I'd check whether they disabled the `email_on_private_chat_message` setting. Otherwise there'd be no way to disable these emails.

backend/climateconnect_api/tasks.py
```diff
@@ -1,9 +1,50 @@
...
 
 @app.task
-def add(a, b):
-    logger.info(f"testing.... {a+b}")
+def schedule_automated_reminder_for_user_notifications():
+    # Get all user_ids for people who have not checked their notification
+    all_user_ids = list(
+        UserNotification.objects.filter(
+            read_at__isnull=True,
+            created_at__lte=(timezone.now() - timedelta(days=2))
```
positiveimpact, line 23: I'd only do this for the private message notifications. Doing this for project likes etc. will not be necessary imo.

backend/climateconnect_main/celery.py
```diff
@@ -12,3 +13,10 @@
 #   should have a `CELERY_` prefix.
 app.config_from_object('django.conf:settings', namespace='CELERY')
 app.autodiscover_tasks()
+
+app.conf.beat_schedule = {
+    'schedule_automated_email_reminder_for_notifications': {
+        'task': 'climateconnect_api.tasks.schedule_automated_reminder_for_user_notifications',
+        'schedule': timedelta(days=2)  # Task runs every two days
```
positiveimpact, line 20: Should we only run this every 2 days? This would mean that if you get a message right after the task ran it would take almost 4 days until you get a reminder. Maybe instead we could log when we sent reminders in the db and run this task more often? Then we could also track better how many reminders we're sending.

backend/climateconnect_main/settings.py
```diff
@@ -270,9 +270,9 @@
 
 # For Celery we use Redis as a broker URL
 CELERY_BROKER_URL = env('CELERY_BROKER_URL')
-CELERY_BROKER_USE_SSL = {
-    'ssl_cert_reqs': ssl.CERT_REQUIRED
-}
+# CELERY_BROKER_USE_SSL = {
+#     'ssl_cert_reqs': ssl.CERT_REQUIRED
+# }
```
positiveimpact, line 275: Did you just comment these out for development or is this supposed to stay commented out?

positiveimpact (commented):
The code works perfectly for me! :) 
I made some comments about how I think the reminders should work, especially that I'd only send them for private notifications.
Also the email arrived in my promotions tab on gmail. Likely that is because noreply@climateconnect.earth is already marked as an email that goes to promotions tab. Therefore it would probably be nicer to add some styling and use a template since removing the styling doesn't seem to make a difference

backend/climateconnect_api/tasks.py
ddhanesha: Not even group messages?

backend/climateconnect_main/celery.py
ddhanesha: We can run this task everyday, but I don't think so we should run this task more than once a day.

backend/climateconnect_main/settings.py
ddhanesha: Yeah, locally you don't have ssl cert so I commented it out.

ddhanesha:
> The code works perfectly for me! :) I made some comments about how I think the reminders should work, especially that I'd only send them for private notifications. Also the email arrived in my promotions tab on gmail. Likely that is because [noreply@climateconnect.earth](mailto:noreply@climateconnect.earth) is already marked as an email that goes to promotions tab. Therefore it would probably be nicer to add some styling and use a template since removing the styling doesn't seem to make a difference

I am getting email on my primary setting. We can use another email like an alias for @tobiasrehm or @positiveimpact's email. Example: `tobias@climateconnect.earth` or `chris@climateconnect.earth`

ddhanesha:
@positiveimpact proof of the email being received in primary email folder
Screen Shot 2022-03-03 at 8 41 10 PM

backend/climateconnect_api/tasks.py
ddhanesha: Done

backend/climateconnect_api/tasks.py
positiveimpact: In the first step, I'd only do it for private messages because those are for sure addressed to you while group messages don't need to be addressed to you. If we also do it for group messages we have to make sure that we still respect their settings, if they disabled notifications on group messages. In this case we could maybe say You missed messages in group x and you missed private messages.

backend/climateconnect_main/celery.py
positiveimpact: Is that so that we don't send too many emails? Because for when to send an email I'd just save in the db when we last sent one. But sure, once per day seems good!

backend/climateconnect_api/utility/email_setup.py
```diff
@@ -295,7 +295,7 @@ def send_email_reminder_for_unread_notifications(
     language_code = get_user_lang_code(user=user)
     subject_by_language = {
         "en": f"You have {total_notifications} unread notifications",
```
positiveimpact, line 297: We also need to change this one to unread messages, otherwise it doesn't fit to the email body

backend/climateconnect_main/settings.py
positiveimpact: > > The code works perfectly for me! :) I made some comments about how I think the reminders should work, especially that I'd only send them for private notifications. Also the email arrived in my promotions tab on gmail. Likely that is because [noreply@climateconnect.earth](mailto:noreply@climateconnect.earth) is already marked as an email that goes to promotions tab. Therefore it would probably be nicer to add some styling and use a template since removing the styling doesn't seem to make a difference
> 
> I am getting email on my primary setting. I think removing styling worked. But we can use another email like an alias for @tobiasrehm or @positiveimpact's email. Example: `tobias@climateconnect.earth` or `chris@climateconnect.earth`

Ah ok, that's nice! Then sure, we can leave it like this!

positiveimpact:
> > The code works perfectly for me! :) I made some comments about how I think the reminders should work, especially that I'd only send them for private notifications. Also the email arrived in my promotions tab on gmail. Likely that is because [noreply@climateconnect.earth](mailto:noreply@climateconnect.earth) is already marked as an email that goes to promotions tab. Therefore it would probably be nicer to add some styling and use a template since removing the styling doesn't seem to make a difference
> 
> I am getting email on my primary setting. I think removing styling worked. But we can use another email like an alias for @tobiasrehm or @positiveimpact's email. Example: `tobias@climateconnect.earth` or `chris@climateconnect.earth`

Ah ok, that's nice! Then sure, we can leave it like this!

backend/climateconnect_api/utility/email_setup.py
ddhanesha: You mean change text from`unread notifications` to `unread messages`?

backend/climateconnect_api/tasks.py
ddhanesha: Done

backend/climateconnect_main/celery.py
ddhanesha: Changed this!

ddhanesha:
@positiveimpact can you re-review?

positiveimpact (approved):
LGTM! From my point of view this is ready to merge :)

piperchester:
@ddhanesha does this close out https://github.com/climateconnect/climateconnect/issues/712?

PR #896, opened 2022-02-18
Adding back redis cache to user profile GET request

ddhanesha:
## Description

This PR is to add back redis cache back user profiles

## Test plan
You need to setup redis instance locally to test this. Use docker to run redis server, setup `.backend_env` to add REDIS_URL. Example: `REDIS_URL="redis://127.0.0.1:6379/0"`

The improvement you will see is members page in frontend will have better performance. 

## Before landing

1. PR has meaningful title


ddhanesha:
> I'd try to directly implement this in a way that the cache gets overwritten if a new member profile is created. Otherwise new members won't show up in the list for up to two days and that's a functionality that we'll need for caching other pages anyways

Figured this out! I added a logic where if new member is signed up we remove cache keys so when the first API call is made the keys gets recreated.

positiveimpact:
> > I'd try to directly implement this in a way that the cache gets overwritten if a new member profile is created. Otherwise new members won't show up in the list for up to two days and that's a functionality that we'll need for caching other pages anyways
> 
> Figured this out! I added a logic where if new member is signed up we remove cache keys so when the first API call is made the keys gets recreated.

Awesome! :raised_hands:

ddhanesha:
> > > I'd try to directly implement this in a way that the cache gets overwritten if a new member profile is created. Otherwise new members won't show up in the list for up to two days and that's a functionality that we'll need for caching other pages anyways
> > 
> > 
> > Figured this out! I added a logic where if new member is signed up we remove cache keys so when the first API call is made the keys gets recreated.
> 
> Ah, so it removes all cache keys or just the one that was changed

As discussed online we have to remove all keys because its caching the API page not a specific object of UserProfile table.

ddhanesha:
@positiveimpact will have time to check this? Otherwise I can merge this for now.

Issue #802, opened 2021-12-06
Bug: Organization page: edit button overlaps with organization image on tiny screens


verena-e-p:
When the user is logged in on their own organization page and the screen is smaller than around 400 px in width, the edit organization button overlaps with the organization image.

(_As mentioned here: https://github.com/climateconnect/climateconnect/issues/792#issuecomment-986853834_)

After relocating the button, check if the CSS style change to the container of the SocialMediaShareButton on the organization account is still necessary. (_See commit: https://github.com/climateconnect/climateconnect/pull/792/commits/a74ded30d3c743997e95c0ab06993511144a6c23_)

positiveimpact:
Steps to reproduce:
1. Create an account by signing up on your local platform
2. Create an organization on /createorganization
3. Go to the organization's page with a screen resolution of less than 400px

Additionally to the edit button overlapping, the "Share a project" button is also looking a bit weird on small mobile screens. It could be worth it to just display "Share" instead of "Share a project" on tiny screens (breakpoint xs)

Image for context:
![Uploading Screenshot from 2022-02-25 18-36-05.png…]()

tobiasrehm:
Was fixed. But there is still a bug related to this: #1199. Closing this one

PR #903, opened 2022-03-07
Bug 802 fix

kreider94:
## Description

#802 

Moved the `edit` button to the center of the screen and below the avatar info. 

Screenshot 2022-03-07 at 15 58 09

## Before landing

1. PR has meaningful title

### DIFF
--- a/frontend/src/components/account/AccountPage.js
+++ b/frontend/src/components/account/AccountPage.js
@@ -1,4 +1,4 @@
-import { Avatar, Button, Chip, Container, Link, Tooltip, Typography } from "@material-ui/core";
+import { Avatar, Button, Chip, Container, Link, Tooltip, Typography, Divider } from "@material-ui/core";
 import { makeStyles } from "@material-ui/core/styles";
 import PlaceIcon from "@material-ui/icons/Place";
 import React, { useContext } from "react";
@@ -94,11 +94,16 @@ const useStyles = makeStyles((theme) => ({
     marginRight: theme.spacing(1),
   },
   editButton: {
-    position: "absolute",
-    right: theme.spacing(1),
-    top: theme.spacing(12),
+    display: "flex",
+    justifyContent: "center",
+    margin: "auto",
+    marginBottom: theme.spacing(3),
+    marginTop: theme.spacing(1),
     [theme.breakpoints.up("sm")]: {
       top: theme.spacing(1),
+      position: "absolute",
+      right: theme.spacing(1),
+      top: theme.spacing(12),
     },
     [theme.breakpoints.down("xs")]: {
       width: theme.spacing(14),
@@ -303,16 +308,6 @@ export default function AccountPage({
         )}
       </div>
       <Container className={classes.infoContainer}>
-        {isOwnAccount && (
-          <Button
-            className={classes.editButton}
-            color="primary"
-            variant="contained"
-            href={editHref}
-          >
-            {editText ? editText : texts.edit_profile}
-          </Button>
-        )}
         <Container className={classes.avatarWithInfo}>
           <div className={classes.avatarContainer}>
             {account.badges?.length > 0 ? (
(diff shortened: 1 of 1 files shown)


positiveimpact:
Thanks for creating this PR so quickly! :) Putting it under the org name is a great solution! However I think we can put all the text in one line so that the button looks a bit more natural. Also it would be great if you could change "Share a project" to "Share" on mobile and screen sizes where "Share a project" doesn't fit into one line :)

frontend/pages/organizations/[organizationUrl].js
```diff
@@ -205,9 +210,15 @@ function OrganizationLayout({
...
-          </Button>
+          <Link
+            href={getLocalePrefix(locale) + "/share"}
+          >
+            <ControlPointSharpIcon
+              className={classes.button}
+              variant="contained"
+              color="primary"
+            />
+          </Link>
```
Comment (positiveimpact) [inline frontend/pages/organizations/[organizationUrl].js:221]: I'd use the `<IconButton>` component instead of link here - similar to the share button. This will add some interactivity on hover and click. Also I'd only display the plus icon on mobile On larger screens the "Share a project" button is more descriptive. 
```suggestion
          {!isTinyScreen && (
            <Button variant="contained" color="primary" href={getLocalePrefix(locale) + "/share"}>
              {texts.share_a_project}
            </Button>
          ) : (
            <IconButton
              href={getLocalePrefix(locale) + "/share"}
            >
              <ControlPointSharpIcon
                className={classes.button}
                variant="contained"
                color="primary"
              />
            </IconButton>
          )
          }
```

frontend/pages/organizations/[organizationUrl].js
```diff
@@ -224,14 +235,17 @@ function OrganizationLayout({
...
-              {texts.manage_members}
-            </Button>
+              <EditSharpIcon
+                className={classes.button}
+                variant="contained"
+                color="primary"
+              >
+                {texts.manage_members}
+              </EditSharpIcon>
+            </Link>
```
Comment (positiveimpact) [inline frontend/pages/organizations/[organizationUrl].js:248]: Same as above. Would use a button for normal screens and an IconButton for mobile. As an icon, I'd use the `GroupAddIcon` to not have the same Icon as with edit and because we're also using that to manage members of a group chat.

frontend/src/components/account/AccountPage.js
```diff
@@ -285,6 +294,21 @@ export default function AccountPage({
...
+          >
+            <EditSharpIcon
+              className={classes.editButton}
+              color="primary"
+              variant="contained"
+            >
+              {editText ? editText : texts.edit_profile}
+            </EditSharpIcon>
+          </Link>
+        )}
```
positiveimpact, line 311: Same as above here.

positiveimpact (commented):
Added some more comments because I think we need to display things differently on mobile and desktop. Sorry for being so unspecific in the issue and in my previous comments.

kreider94:
@positiveimpact I have made those changes you requested. I also thought the icons are rather descriptive, so I have added these icons to the desktop versions: 
Screenshot 2022-03-20 at 17 05 42

Please let me know if you like this idea. If not, I can remove the icons within the buttons. 
I have noticed that there are many similar pages which could benefit from using the same styling template (using headers, buttons and items) which would result in less code and fewer files, and would mean we can reuse the code more. But maybe I will save this for another ticket.

positiveimpact (approved):
Thanks for implementing the changes! Looks great now, I love the icons on the desktop buttons as well :ok_hand: :heart_eyes: 
Since both the profile page and the organization page use the `AccountPage` component the changes also apply to the profile page already :100: It could definitely make sense to also apply this to the project page or potentially other places but I'd focus this PR just on the AccountPage.
On desktop it's working perfectly but I'm having some issues on mobile now.

1. Half of the "edit" button is hidden behind the "share" button on smaller screen sizes
2. The "edit" button isn't clickable on small screensizes, it's clickbox is hidden behind the "infoContainer" element. The same issue applies to the "share" button so it doesn't seem to be new.
Peek 2022-03-20 20-30

A potential fix to these problems could be to place the share and edit button into one container, making the container's position absolute and adding a zIndex to it but I'm not sure if this is the best option.

After addressing that you can merge it. I'll already approve it now because I'm on vacation next week. Thanks for your work! :)

Issue #382, opened 2020-11-11
Counter for number of characters in the "Summarize your project" section when sharing/editing a project


ImpactMaker-2:
The summary of a climate project on Climate Connect only allows up to 240 characters (1 tweet) `frontend/src/components/shareProject/AddSummarySection.js` should show a counter of how many characters your summary has so far. 
This will help users to see how many characters they have left while writing their project's summary

positiveimpact:
@KParrish193 Are you still actively working on this issue?

KParrish193:
I am but am cool moving to something else if someone else needs to work on this, or it's not needed anymore.

positiveimpact:
> I am but am cool moving to something else if someone else needs to work on this, or it's not needed anymore.

Not, it's still needed and I'm fine with you working on it, just wanted to make sure the issue is still actively being worked on.

positiveimpact:
Fixed by #910

PR #906, opened 2022-03-08
Add counter to textfield

atlas033:
## Description
added a counter to the textfield short description in share project

github issue #382 

## Before landing

1. PR has meaningful title

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ frontend/.vscode
 backend/.azure
 
 yarn.lock
+frontend/.eslintcache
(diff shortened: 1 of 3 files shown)


positiveimpact (commented):
The counter is working very smoothly! 
2 changes I'd still make:
1. The counter looks a bit lost currently. Maybe we could instead add it to the `helperText`. It could look like this: "Briefly summarize what you are doing (0 / 280 characters)" and then update live.
2. We also need to implement this for editing a project - specifically in the EditProjectOverview

Issue #496, opened 2021-02-17
Remove the explicit indication of missing username or wrong password at login


FnayouSeif:
In the login process, if the user does not exist, a message will indicate that the account does not exist. 
If the password is incorrect, a message will explicitly indicate so. 
This is a security risk. It is a common practice to issue a standard message "Invalid Email or Password".

positiveimpact:
![Screenshot from 2023-02-10 08-39-10](https://user-images.githubusercontent.com/10531079/218031294-8137d862-b732-435d-a114-739173792f4f.png)
Seems to be fixed. Closing

PR #907, opened 2022-03-08
Fix: Changed incorrect login message

atlas033:
## Description

removed the explicit indication of missing username or missing password to a more generic "invalid username or password" 
also added german translation

github issue #496 

## Before landing

1. PR has meaningful title


frontend/src/components/shareProject/AddSummarySection.js
```diff
@@ -68,6 +68,7 @@ export default function AddSummarySection({
           value={projectData.short_description}
         />
       </div>
+      <Typography> {projectData.short_description ? projectData.short_description.length: 0} / 280</Typography>
```
positiveimpact, line 71: Make sure to not mix changes from different PR's in here.

backend/climateconnect_api/views/user_views.py
```diff
@@ -67,12 +67,8 @@ def post(self, request, format=None):
                 user_profile.save()
             return super(LoginView, self).post(request, format=None)
         else:
-            if not User.objects.filter(username=request.data['username']).exists():
-                return Response({
-                    'message': 'Username does not exist. Have you signed up yet?'
-                }, status=status.HTTP_401_UNAUTHORIZED)
             return Response({
-                'message': 'Invalid password.'
+                'message': _('Invalid username or password.')
```
positiveimpact, line 71: Since users don't have a username but only an email, I'd also change that here (even though it was also wrong before). Accordingly in german we could then also use "Ungültige E-Mail oder Passwort"
```suggestion
                'message': _('Invalid email or password')
```

positiveimpact (commented):
Great job on changing the message and correctly implementing the translation through Django's system. I left 2 comments, once these are addressed, we can merge.

frontend/src/components/shareProject/AddSummarySection.js
atlas033: Thank you for your review i reverted the changes and it should be clean now.

backend/climateconnect_api/views/user_views.py
atlas033: I changed it and updated the django language files.

positiveimpact (approved):
LGTM now! Great job! :)

PR #910, opened 2022-03-09
added textfield counter

atlas033:
## Description

#382 

Added a textfield counter when editing a project and when sharing a project.

## Before landing

1. PR has meaningful title


atlas033:
Apologies for opening a new pull request. Here is positiveimpact's comment on the previous pull request: 


[positiveimpact](https://github.com/positiveimpact) left a comment
> The counter is working very smoothly!
> 2 changes I'd still make:
> 
>    The counter looks a bit lost currently. Maybe we could instead add it to the helperText. It could look like this: "Briefly summarize what you are doing (0 / 280 characters)" and then update live.
>    We also need to implement this for editing a project - specifically in the EditProjectOverview

positiveimpact (approved):
LGTM other than a very small wording suggestions that I'll quickly commit. Good work! :)

Issue #454, opened 2021-01-26
Rework project, organization and profile url generation


positiveimpact:
# Current behavior
We generate urls for profiles, projects and organizations by making their names into url strings. Currently they aren't properly escaped, for example this recently created organization name "https://www.spiritofndanifor.com" resulted in this url `https://climateconnect.earth/organizations/https://www.spiritofndanifor.com` which then didn't work. Generally urls look really bad and are very long

# Expected future behavior
- Url's should be properly escaped so that all urls work
- project urls should not have numbers attached in the end unless this is necessary. This can be achieved by trying to insert without a number and only attaching the id to the end if the project url_slug already exists

FnayouSeif:
1. How would you describe url escaping? Would you encode it? If yes, that will result in some confusing urls I think.
2. I think this will create an overhead for inserting a record. I mean, ideally, the database should be able to guarantee uniqueness at the moment of creation. Does the url slug need to be human readable?

positiveimpact:
> 1. How would you describe url escaping? Would you encode it? If yes, that will result in some confusing urls I think.
I would only encode characters that make the url break and are currently not being encoded. The two characters that fit that scheme that I'm aware of are `&` and `/`
> 2. I think this will create an overhead for inserting a record. I mean, ideally, the database should be able to guarantee uniqueness at the moment of creation. Does the url slug need to be human readable?
I don't think the overhead will be too big for now since we very rarely have duplicate names on projects, orgs and members. I think nice looking urls are a bit better for linking stuff and look a bit cleaner but obviously this is just a very minor improvement and should be reverted to `formattedProjectName + projectId` once we have so many duplicate entries that it's taking a toll on the db

positiveimpact:
Fixed by #915

PR #915, opened 2022-03-23
fix/added new slug generation to profile, organization and project

atlas033:
## Description

A new slug generation was added to Profile, Organization and Project generation. It's similar to the slug generation in idea. Its implemented straight forward, but also could probably be generalized.  

#454

## Test plan
needs to be tested manually because of the simple implementation

## Before landing

1. PR has meaningful title
Screenshot from 2022-03-23 12-49-52
Screenshot from 2022-03-23 12-53-45

### DIFF
--- a/backend/climateconnect_api/views/user_views.py
+++ b/backend/climateconnect_api/views/user_views.py
@@ -27,6 +27,7 @@ from django.contrib.gis.db.models.functions import Distance
 from django.db.models import Count, Q
 from django.utils import timezone
 from django.utils.translation import gettext as _
+from django.utils.text import slugify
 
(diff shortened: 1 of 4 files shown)


positiveimpact (commented):
Great idea to use slugify there! I think it would make sense to generalize this in a utility function before merging it - and to also only add the id to the profile url slug if somebody else with this url slug already exists. Other than that it looks good! :)

atlas033:
added suggested changes

positiveimpact (approved):
Made one small fix to the user profile url slug generation. LGTM now, great work on this really important update :rocket:

PR #922, opened 2022-03-30
feature: short description input of organization made mandatory

siuting0511:
TextField in DetailledDescriptionInput component changed to required, error managed with native error throw.

### DIFF
--- a/frontend/src/components/account/DetailledDescriptionInput.js
+++ b/frontend/src/components/account/DetailledDescriptionInput.js
@@ -27,6 +27,7 @@ export default function DetailledDescriptionInput({ title, helpText, value, onCh
         </Tooltip>
       </Typography>
       <TextField
+        required = {true}
         variant="outlined"
         fullWidth
         multiline


positiveimpact (changes requested):
Thanks for working on this! This is a good solution in general :ok_hand:  
But the PR was refering to the summary, not the about section. The About section should not be mandatory as it's for explaining more in detail what your organization is doing. But a summary should be mandatory.
922_explanation

positiveimpact:
@siuting0511 Yep, sure! I'm just seeing that there is basically no documentation for this component so I'm gonna add some.
The summary is part of the accountInfo which is generated by the `displayAccountInfo` function in `EditAccountPage`.
Normal textfields such as summary are being handled [here](https://github.com/climateconnect/climateconnect/blob/671643d394975a64cdfe568708a7d12cd697ca59/frontend/src/components/account/EditAccountPage.js#L596) based on the information provided by the corresponding infoMetadata object. For organizations you can find this object in the [`organization_info_metadata`](https://github.com/climateconnect/climateconnect/blob/master/frontend/public/data/organization_info_metadata.js) file. The summary is in the "short_description" entry. 

You could add a property `required: true` here and adapt the code in displayAccountInfo to add a required property if required is set to true. 

Sorry for the badly commented code, let me know if you need more clarification!

siuting0511:
Thank you for the explanation! I have made the following changes:
1. add `required:true` to organisation meta data
2. add attribute `required` to the corresponding TextField component if the meta data had `required:true` attribute

frontend/src/components/account/EditAccountPage.js
```diff
@@ -594,7 +594,9 @@ export default function EditAccountPage({
                 </Tooltip>
               )}
             </Typography>
-            <TextField fullWidth value={i.value} multiline onChange={handleChange} />
+            {i.required? 
+              <TextField required fullWidth value={i.value} multiline onChange={handleChange} />
+              : <TextField fullWidth value={i.value} multiline onChange={handleChange} />}
```
positiveimpact, line 599:
```suggestion
              <TextField required={i.required} fullWidth value={i.value} multiline onChange={handleChange} />
```
I'd use this annotation as it saves 2 lines in this context.

positiveimpact (commented):
Works great for me now, thanks for making the change! I added one notation suggestion. Once that is implemented, this looks ready to merge! :)

siuting0511:
yes that's a better solution!

Issue #919, opened 2022-03-28
Bug: runtime error while adding location and switching back and forth when signing up


atlas033:
**Describe the bug**
runtime error at sign up (not crucial) 
happens when you switch back and forth after you added a location

**To Reproduce**
1. Go to sign up
2. Procede to the second sign up page
3. Add a location
4. go back to the first sign up page
5. press continue to procede to the second sign up page again
6. runtime error 

**Expected behavior**
It should save the location value
quick fix: remove the location value when going back

**Screenshots**
![Screenshot from 2022-03-28 16-17-55](https://user-images.githubusercontent.com/56408063/160419637-8153c053-c25f-4387-85ad-b54c5ba9466c.png)

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

siuting0511:
Hi, it could probably fixed by adding a useState hook to store the value in AddInfo component, may I give it a try?

atlas033:
> 

Hi, yes that would be great!

PR #926, opened 2022-04-01
fix: sign up page location bug

siuting0511:
#919 

variable `value` is `undefined` in run-time useEffect function, use [optional chain operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) so there wouldn't be run-time error if the `value` is `null`, return `undefined` instead.


atlas033 (approved):
LGTM! Thanks for your help!

PR #936, opened 2022-04-19
Migrating Django to 3.2

FnayouSeif:
## Description

In order to migrate to 3.2, these warnings were issued and they are related to these new changes: 

1. AutoField type default behavior 
As of 3.2, the default type for PK should be explicitly defined on djanog, app or model level config, 
Documentation: https://docs.djangoproject.com/en/4.0/releases/3.2/#customizing-type-of-auto-created-primary-keys

2. Added http protocol in `backend/climateconnect_main/routing.py` as recommended in the warning message of migration 
3. Upgraded 2 django plugins : `djangorestframework` and `channels-redis`. The new django versions will not support some configurations with those 2 libraries so an upgrade was needed (based on warning messages)  

## Test plan

Manual testing was done on some essential features. 

## Before landing

1. PR has meaningful title


backend/chat_messages/apps.py
```diff
@@ -2,4 +2,5 @@
 
 
 class ChatMessagesConfig(AppConfig):
+    default_auto_field = 'django.db.models.BigAutoField'
```
ddhanesha, line 5: What does this field do? Why do you have this at an AppConfig level?

ddhanesha:
Can you resolve conflicts with master branch?

PR #938, opened 2022-04-20
Fix CI actions for frontend code; rename GitHub actions

piperchester:
## What and Why

This updates our GitHub action files. It adds a `format` step for frontend code, and it renames the files. On every new code change, we should both lint, and format, and fail the build if there are any failures or diffs. This will ensure that each new diff that lands will match the pre-existing codebase.

Used https://github.com/nektos/act to test actions changes locally.


ddhanesha:
@piperchester is this ready for review?

piperchester:
Yep! Should be ready @ddhanesha.

ddhanesha (approved):
Let's get this merged!

PR #942, opened 2022-05-11
Added file we should ignore

ddhanesha:
## Before landing

1. PR has meaningful title


ddhanesha:
Merging as there are no breaking changes

PR #943, opened 2022-05-11
Changed command to run celery scheduled job server on production

ddhanesha:
## Description

PR includes an update on script to run celery beat server on production.

### DIFF
--- a/start_celery.sh
+++ b/start_celery.sh
@@ -9,5 +9,5 @@ cd backend
 # Install requirements
 pip install -r requirements.txt
 
-# starting celery beat (scheduled job) server
+# starting celery worker
 celery -A climateconnect_main -l INFO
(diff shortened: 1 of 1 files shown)


FnayouSeif (commented):
@ddhanesha 
I am not sure but running `beat` will only start the beat service which will trigger tasks. Nevertheless, the trigger task will require a worker for it to be processed. I believe that this script should launch the 2 apps (beat + worker). No?

ddhanesha:
> @ddhanesha I am not sure but running `beat` will only start the beat service which will trigger tasks. Nevertheless, the trigger task will require a worker for it to be processed. I believe that this script should launch the 2 apps (beat + worker). No?

Yes, you are correct @FnayouSeif. I will update the command to launch 2 apps. Totally missed that!

PR #965, opened 2022-07-13
Fix: Inbox alert only opened once

JoshKornfeld:
## Description

When clicking the button "Start Chat" while the searchbar is empty, an alert would appear, indicating the user needs to search for another user to message. Closing the alert and repeating the same step does not open the alert again.

The wording for the alert was also changed in both English and German.

Potential conflict could occur with ["frontend/pages/editprofile.js"] as it also contains a component which uses the WideLayout component which was slightly modified.

## Test plan

- Sign in 
- Navigate to inbox
- Press the "New Chat" button
- Press the "Start Chat" button
- Close the alert
- Press the "Start Chat" button again

## Before landing

1. PR has meaningful title

### DIFF
--- a/frontend/pages/inbox.js
+++ b/frontend/pages/inbox.js
@@ -96,6 +96,10 @@ export default function Inbox({ chatData, next }) {
     next: next,
   });
 
+  const resetErrorMessage = (e) => {
+    setErrorMessage(e);
+  };
+
   const handleGroupNameChange = (e) => {
     setGroupName(e.target.value);
   };
@@ -119,6 +123,7 @@ export default function Inbox({ chatData, next }) {
     if (newChatMembers.length >= 1) {
       const payload = {};
       if (isGroupchat) {
+        //The if statement (line 127) is never entered due to TextField having a the required property set (line 216)
         if (!groupName) {
           setErrorMessage(
             texts.please_set_a_group_name_if_youre_starting_a_chat_with_more_than_one_member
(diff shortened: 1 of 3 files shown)


frontend/pages/inbox.js
```diff
@@ -96,6 +96,10 @@ export default function Inbox({ chatData, next }) {
     next: next,
   });
 
+  const resetErrorMessage = (e) => {
+    setErrorMessage(e);
+  };
+
```
positiveimpact, line 102: If this is generic it should be called `setErrorMessage`. If it's always clearing the message then I wouldn't allow to pass a value but always default to ""

positiveimpact (approved):
Good catch! Commented a small nit, other than that it looks good to me!

frontend/pages/inbox.js
```diff
@@ -179,7 +184,12 @@ export default function Inbox({ chatData, next }) {
 
   return (
     <div>
-      <WideLayout title={texts.inbox} messageType="error" message={errorMessage}>
+      <WideLayout
+        title={texts.inbox}
+        messageType="error"
+        message={errorMessage}
+        resetAlertMessage={resetErrorMessage}
```
JoshKornfeld, line 191:
```suggestion
        resetAlertMessage={updateErrorMessage}
```

frontend/pages/inbox.js
```diff
@@ -96,6 +96,10 @@ export default function Inbox({ chatData, next }) {
     next: next,
   });
 
+  const resetErrorMessage = (e) => {
```
JoshKornfeld, line 99:
```suggestion
  const updateErrorMessage = (e) => {
```

positiveimpact (approved):
LGTM! Good catch and great work :+1:

Issue #991, opened 2022-07-20
missing translation on about page because FAQ is only availabel in english right now


tobiasrehm:
FAQ would need to be translated or this section removed from the about page 

![Image](https://user-images.githubusercontent.com/64414429/179974675-e064006a-a97a-4e24-a5af-d446a7d77071.png)

JoshKornfeld:
Just a quick addition, all the FAQs aren't translated in the FAQ page either.

JoshKornfeld:
Closed by #995

PR #995, opened 2022-07-22
Fix: FAQ translations

JoshKornfeld:
## Before merging
Run migrations and go to django admin to select which section should be shown on the about page

## Description
solves issue #991 
Backend changes:
- to FAQ which updates the serializer to access the translations for the FAQs. 
- a new url to access a new FAQ view (AboutFaqView)
- new field added to FaqSection that identifies which section should be shown on the About page

Frontend changes include:
 - Filtering questions is now case insenstive
 - Previously unseen text is now properly displayed
 - A link is provided to send an email if a question isn't listed in the FAQ
 - In the German version of the FAQ, the background image properly loads
 - HowItWorks section added small left margin to text in the carousel
 - About page makes a different api call to the new view to get the FAQs by section

## TODO: 
- (On production) Question: How do i create an account? -> Should be: How do I create an account? 
- Live Database needs to be modified to use the new field in the FaqSection model (Basics section ticked to True)

## Test plan

You can see the FAQ changes by doing of the following:
1. Open the about page and scroll down to the "got a question section" and then change the language and repeat
2. Open the FAQ page and then change the language and repeat

Filtering Questions:
- Start typing the FAQ you want to search for

Email a Question:
- Press the link above the searchbar

## Before landing

1. PR has meaningful title


frontend/pages/about.js
```diff
@@ -118,15 +117,19 @@ export default function About({ faqQuestions }) {
...
-const getQuestionsWithAnswers = async () => {
+const getQuestionsWithAnswers = async (token, locale) => {
   try {
-    const resp = await apiRequest({ method: "get", url: "/api/list_faq/" });
+    const resp = await apiRequest({
+      method: "get",
+      url: "/api/about_faq/",
+      token: token,
+      locale: locale,
+    });
```
positiveimpact, line 127: Why do we need the token for this request?

positiveimpact (approved):
LGTM other than some minor things that I commented.

Issue #990, opened 2022-07-20
Visual bug when loggin out on the climatehub page


tobiasrehm:
When clicking log out while on the climatehub this visual bug happens:

![Image](https://user-images.githubusercontent.com/64414429/179972731-b93f6767-7f50-4e3f-b6e2-13a62bc39888.png)

JoshKornfeld:
Closed by #1001

PR #1001, opened 2022-07-25
Fix: ClimateHub visual bug on logout

JoshKornfeld:
## Description
Solves issue #990 
When logging out while viewing a ClimateHub, the HubHeader would move upward and required a page refresh to be in the correct location. This is now fixed.

The HubHeaders are now much more similar and their location on the page is matching. 

A fine grey line and some whitespace used to display on screen when the screen width was less than 960. This is no longer the case. This is also fixed for sector hubs.

A function that would highlight certain words within the title of the header was removed as it wasn't working as intended and hindered the banners being similar when logged in or out.

## Test plan
1. Sign in
2. Open a ClimateHub
3. Log out
4. The Header container should no longer "jump" upward

## Before landing

1. PR has meaningful title


frontend/src/components/hub/HubHeaderImage.js
```diff
@@ -16,7 +16,23 @@ const useStyles = makeStyles((theme) => ({
...
+    backgroundPosition: "bottom center",
+    zIndex: -1,
+    width: "100%",
+    [theme.breakpoints.down("sm")]: {
+      minHeight: 100,
+      backgroundSize: "cover",
+    },
+    position: "relative",
+    [theme.breakpoints.up("md")]: {
+      position: "relative", // we want to have relative positioning when its a location hub and user is logged out
```
JoshKornfeld, line 35:
```suggestion
      position: "relative", // we want to have relative positioning when user is logged out
```

frontend/src/components/layouts/WideLayout.js
```diff
@@ -97,6 +98,7 @@ export default function WideLayout({
         transparentHeader={transparentHeader}
         noSpacingBottom={isStaticPage}
         background={headerBackground}
+        isHubPage={isHubPage}
```
ddhanesha, line 101: Hey @JoshKornfeld this is a common WideLayout page that is used by multiple other pages. I don't think it is good a idea to add Hub specific arguments here. cc @positiveimpact

ddhanesha (changes requested):
The bug is fixed but I am curious to know if there is an alternative option where we don't have to change a common layout page. cc @positiveimpact

frontend/src/components/layouts/WideLayout.js
positiveimpact: I think since we also have `isStaticPage` and hubs are quite central to our concept, this is not a huge issue imo

positiveimpact (changes requested):
On medium screensizes it doesn't look as expected. In the screenshot below you can see the behavior on 850px width while logged in. The left screen is this PR while the right screen is production
- The box is way too low
- In the new design of the box there is a border in there which is pointless since no buttons are displayed below at this screensize
- There is still a mysterious spacing between the actual header and the green subheader but this might be out of the scope of this PR
Screenshot from 2022-08-16 17-36-17

ddhanesha:
@JoshKornfeld any updates on the comments made by @positiveimpact and myself?

ddhanesha:
The logged out layout is all messed up here. Attached screenshots.
Screen Shot 2022-08-27 at 12 08 52 PM
Screen Shot 2022-08-27 at 12 09 02 PM

ddhanesha (approved):
Functionality is working as expected! Approved

JoshKornfeld:
Resolved conflicts, but require approval to merge @ddhanesha @positiveimpact

positiveimpact (approved):
LGTM, all issues see to be resolved. Great work! :rocket:

PR #1007, opened 2022-07-28
Fix: "All Projects" link now displays onscreen while looking at a hub

JoshKornfeld:
## Description

When viewing a location or sector hub the navigation sub-header would have an invisible button. If you highlight the text in the sub navigation sub-header you would see "All projects" that when clicked, would route the user to the browse page. This button is now visible.

Additionally, all the items in the navigation sub-header are properly inline. (The dropdown buttons used to be a few pixels lower)

## Test plan

- For sub-header items, simply open the browse page and look at the navigation sub-header

- To see the "All Projects" link:
1. Open the browse page and select "All hubs" and in the next page select any of the hubs. Or simply click any of the hubs already displayed on the navigation sub-header. Lastly, you can open a hub by opening the "ClimateHubs/SectorHubs" dropdown buttons and picking one.
2. Look at the navigation-subheader to see the "All Projects" link.

## Before landing

1. PR has meaningful title

### DIFF
--- a/frontend/src/components/hub/NavigationSubHeader.js
+++ b/frontend/src/components/hub/NavigationSubHeader.js
@@ -94,6 +94,7 @@ export default function NavigationSubHeader({ hubName, allHubs, isLocationHub })
             locale={locale}
             isNarrowScreen={isNarrowScreen}
             showAllProjectsButton
+            linkClassName={classes.link}
           />
         </Typography>
       </Container>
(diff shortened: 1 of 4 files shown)


positiveimpact (approved):
The "all projects" button and the HubsDropDown weren't centered. I made a commit and used position: "flex" and alignItems: "center" to achieve this. LGTM now.

Issue #970, opened 2022-07-14
Can't save user profile (still investigating)


tobiasrehm:
It seems like a problem with the automatic translation. 
It's currently not possible to save any changes to a profile 

while trying around I got this: 

![Image](https://user-images.githubusercontent.com/64414429/178942612-dbbbbb5a-8631-4683-a55a-343e6472b6dc.png)

PR #1010, opened 2022-08-02
Fix: User Profile not saving

JoshKornfeld:
## Description
Solves #970 #1011

Currently 3 bugs:
- Bug 1: If you change the left side text and try to click save or back -> error.

Cause: Typing in the bio field after pressing the check translations button overrides the "info" key which contains _location, availability, bio, skills_ and  _website_, causing them to be null when read later. Data is now preserved when typing into the field.

- Bug 2:  Availability isn't sticking -> change it then go back into edit user profile it will just display initial value set or one 
manually set in Django admin UI.

Cause: When parsing the profile before posting it to the sever, the availability value was being filtered by name instead of by key value.

- Bug 3: Changing your name does not instantly update the url slug. This means that you cannot directly go to "My Profile" / "My organizations" / "My Projects" or edit profile again without refreshing the page manually. This is now changed so the url_slug doesn't change when changing user name.

Cause: url_slug changes when changing name

Replaced hard coded cancel text for the cancel button within the edit profile UI with one in general texts which includes the German translation.

Added a link to the email test for users to delete their account.

## Test plan

1. Sign in
2. Navigate to "My Profile"
3. Click "Edit Profile"

Things you can test:
- Bug 1:  Navigate to "check translations", type something into the field and click "save" or click "back"

- Bug 2:  Change your availability from the dropdown and click "save". The new value should be visible to you.

- Bug 3:  Change either your first or last name and then save. The edit profile button should be visible. Clicking "My Profile" / "My organizations" / "My Projects" should no longer lead to a dead page.

- Cancel button should read " Abbrechen" when using the German page
- Deleting account information text should now include an email link

## Before landing

1. PR has meaningful title


frontend/src/components/account/EditAccountPage.js
```diff
@@ -919,7 +923,9 @@ export default function EditAccountPage({
         {deleteEmail && (
           <Typography variant="subtitle2" className={classes.deleteMessage}>
             <InfoOutlinedIcon />
-            {texts.if_you_wish_to_delete} {deleteEmail}
+            {texts.if_you_wish_to_delete}
+            <div className={classes.spaceStrings}> </div>
+            <Link href="mailto:contact@climateconnect.earth">{deleteEmail}</Link>
```
positiveimpact, line 928: This needs to be generic otherwise the generic `deleteEmail` variable is pointless.
```suggestion
            <Link href={`mailto:${deleteEmail}`}>{deleteEmail}</Link>
```

positiveimpact (approved):
LGTM other than my suggestions. Great work!

PR #1018, opened 2022-08-08
Fix null value error on idea comment notification

ddhanesha:
## Description
Fixes null value error on idea comment notifications. 

Screen Shot 2022-08-07 at 6 48 33 PM

## Test plan
Hard to reproduce it locally, You can try by setting idea comment to null in the database. 

## Before landing

1. PR has meaningful title


positiveimpact (approved):
LGTM!

PR #1020, opened 2022-08-10
Fix for inbox alert only showing once

JoshKornfeld:
## Description

Problem: 
When clicking start chat in the inbox with an empty search field, the user would receive an alert, prompting them to search for something. Upon closing the alert and repeating the process, the alert would not reopen. 

A prior solution makes a change to a generic layout component which has side effects on either parts of the website when closing the alert e.g. changing account settings.

Solution:
Make the start chat button change the alert message within the inbox component so it can be repeatedly shown when clicking with an empty search field.

## Test plan

1. Sign in
2. Navigate to the user inbox
3. Press start chat with an empty search field
4. Close the alert
5. Press start chat again

## Before landing

1. PR has meaningful title

### DIFF
--- a/backend/.gitignore
+++ b/backend/.gitignore
@@ -3,3 +3,5 @@
 /gcloud-storage-credentials.json
 
 static/
+
+media/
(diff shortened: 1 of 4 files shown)


positiveimpact:
This is quite a hacky approach because it tries to work around the functionality of `WideLayout`. We should never hack around our own code, often there is a simpler solution. In this case it seems like the problem was that `resetAlertMessage`gets called only after `setAlertOpen`. 
I fixed this [here](https://github.com/climateconnect/climateconnect/commit/d199f2dada4df5cf30e884157c7de81a272239b7) and additionally slightly altered the `resetAlertMessage` function to act more like its name would suggest.

Issue #1021, opened 2022-08-10
Bug: Inbox when logged out says loading


JoshKornfeld:
**Describe the bug**

Logging out while viewing your inbox, provides user with a loading spinner which could be confusing, we could perhaps make it more clear that the user should log in to see the inbox

![Screenshot from 2022-08-10 14-48-57](https://user-images.githubusercontent.com/49129464/183905498-e39def69-f7e9-408f-8040-27ecdd5f9ca1.png)

**To Reproduce**
1. Sign in
2. Navigate to inbox
3. Log out

**Expected behavior**
Provide user with some message saying they should log in to see their inbox

JoshKornfeld:
Closed by #1022

PR #1022, opened 2022-08-10
Fix: Display inbox screen content based on whether user is logged in or not

JoshKornfeld:
## Description

Solves #1021
Changed content shown on screen when user is logged out while their inbox is open

## Test plan

1. Sign in
2. Navigate to inbox
3. Log out

## Before landing

1. PR has meaningful title


positiveimpact (changes requested):
When you try to load the `inbox` page while being logged out you will be sent to login with the message `Du musst eingeloggt sein, um deinen Posteingang zu sehen`. I think logging out while on the page should prompt the same behavior rather than letting you stay on the page with a LoginNudge

positiveimpact (approved):

PR #1023, opened 2022-08-10
Fix: Ideas page comment size and order in which they are shown

### LINKED ISSUE #1024: Bug: Comment threads in Ideas need changing
**Describe the bug**
Currently replies are displayed newest to oldest. Very long comments are fully displayed and decreases the overall readability of the comment thread.

**To Reproduce**
1. Go to a climate hub
2. Click on Ideas
3. Select an Idea with some comments

**Expected behavior**
Comments should be displayed from oldest to newest
Long comments that exceed a certain character number should be hidden and be openable

**Screenshots**
image

JoshKornfeld:
## Description

Closes #1024 

Changes summary:
- Deleting a reply to a comment wouldn't be shown on the UI unless you refreshed the page -> now deletes immediately on the UI when clicked.
- Added a read more / read less feature when a comment exceeds 3 lines.
- Replies to comments are shown oldest at the top, newest at the bottom. 
- When writing a long comment, the send icon in the ideas page would be moved offscreen -> now the width of the comment input is fixed and it will go into a new line, rather than pushing the icon further to the right. This issue was also visible within a project discussion, however less noticeable.
- Moved send/cancel buttons in commentInput into it's own line so that the send/cancel buttons will always be inline.
- Added a title prop to remove warning as title is a require prop for GeneralDialog

## Test plan
- Sign in & navigate to a project discussion or an idea on a ClimateHub.
- Write a comment that would be long enough to use up 3 lines. A long comment will also highlight the icon moving issue.
- In the comments you should see a read more/read less link.

## Before landing

1. PR has meaningful title


ddhanesha:
@JoshKornfeld you have a conflict with `master` branch. Can you resolve that first?

frontend/src/components/communication/CommentInput.js
```diff
@@ -52,6 +54,7 @@ function CommentInput({
   hasComments,
   infoTextSize,
   useIconButton,
+  defaultProjectCommentInput,
```
positiveimpact, line 57: It seems like nothing is ever being passed through this prop. E.g. in `IdeaCommentsSection` it is passed as className (`className={classes.defaultProjectCommentInput}`) even though `CommentInput` doesn't take a className so this just does nothing. If it's not required I'd just remove it everywhere. If it is required I'd check if it really needs to be generic

frontend/src/components/ideas/IdeasBoard.js
```diff
@@ -93,10 +93,27 @@ export default function IdeasBoard({
...
+      });
+      // remove a reply comment
+    } else {
+      const parentCommentIndex = idea.comments.findIndex(
+        (comment) => comment.id === c.parent_comment_id
+      );
+      const filterOutReplies = [
+        ...idea.comments[parentCommentIndex].replies.filter((pc) => pc.id !== c.id),
+      ];
+      idea.comments[parentCommentIndex].replies = filterOutReplies;
```
positiveimpact, line 110: For editing a state you should always use `setIdea()`. Otherwise the state might get corrupted. 
The best way to do this that I can think of is this: 
```suggestion
      setIdea({...idea, comments: [...idea.comments.filter((pc) => pc.id !== c.id, {...c, c.replies.filter((pc) => pc.id !== c.id)}]}
```

frontend/src/components/project/ProjectCommentsContent.js
```diff
@@ -26,7 +29,21 @@ export default function CommentsContent({ user, project, token, setCurComments }
...
+
+      // remove a comment that has a parent comment
+    } else {
+      const parentCommentIndex = project.comments.findIndex(
+        (comment) => comment.id === c.parent_comment_id
+      );
+      const filterOutReplies = [
+        ...project.comments[parentCommentIndex].replies.filter((pc) => pc.id !== c.id),
+      ];
+      project.comments[parentCommentIndex].replies = filterOutReplies;
```
positiveimpact, line 44: Same thing as in my comment in `IdeasBoard` applies here.

frontend/src/components/project/ProjectCommentsContent.js
```diff
@@ -26,7 +29,21 @@ export default function CommentsContent({ user, project, token, setCurComments }
...
+      setCurComments([...project.comments]);
```
positiveimpact, line 45: However with my solution from IdeasBoard this line won't work anymore since the new value of `project` is only applied after the function finishes running.

positiveimpact (dismissed):
As we discussed the solution with isClamped looks a little bit hacky. Would be great if you try to get this to work with [react-truncate](https://www.npmjs.com/package/react-truncate) using the `onTruncate` function to figure out wither to currently show the button or not.
Besides that it looks good other than my smaller comments :+1:

frontend/src/components/communication/CommentInput.js
JoshKornfeld: Removed

frontend/src/components/ideas/IdeasBoard.js
JoshKornfeld: Redid this using a temp variable for the idea and only editing it then setting the idea to the temp variable.

frontend/src/components/project/ProjectCommentsContent.js
JoshKornfeld: Redid this using a temp variable for the project and only editing it then setting the project to the temp variable.

frontend/src/components/ideas/IdeasBoard.js
```diff
@@ -93,10 +93,29 @@ export default function IdeasBoard({
   };
 
   const handleRemoveComment = (c) => {
```
ddhanesha, line 95: It is not best practice to use alphabet characters like this for argument name. Use a readable argument name. I would recommend updating this method.

frontend/src/components/project/ProjectCommentsContent.js
```diff
@@ -26,7 +26,22 @@ export default function CommentsContent({ user, project, token, setCurComments }
   const comments = project.comments;
 
   const handleRemoveComment = (c) => {
```
ddhanesha, line 28: Same here, use readable argument name here.

ddhanesha (changes requested):

ddhanesha (approved):
Feature is working as expected. But I recommended some code best practices to work on

frontend/src/components/project/ProjectCommentsContent.js
JoshKornfeld: updated variable names to something more descriptive.

frontend/src/components/ideas/IdeasBoard.js
JoshKornfeld: Renamed arguments and variables.

Issue #1032, opened 2022-08-16
Bug: Mention notification in Idea/Project comment section not working


JoshKornfeld:
**Describe the bug**
Mentioning a user in the comment section of an idea or project will generate a notification for them. However, the notification type is that of a "joinprojectrequest" instead of "mention". When the user tries to click on the notification, they experience a crash.

**To Reproduce**
1. Go to a project/idea comment section
2. Write a comment and @mention a user by name (an account you can sign into)
3. Swap accounts to the one you just mentioned
4. Click the bell icon

**Expected behavior**
The user should see he was mentioned by another user, when they click the bell and should be able to dismiss it

JoshKornfeld:
Closed by #1033

PR #1033, opened 2022-08-16
Fix: Correct notification type on mention in comments

JoshKornfeld:
## Description
Solves #1032

Changed the notification type from 9 (project join request) to 11 (mention) in a utility function that is called when posting a comment for either a project or idea.

Added a number to each notification type via comment in frontend array for easier comparison to backend array.

## Test plan

- Sign in and go to the comment section for an idea or project
- Write a comment in which you @ the name of a user, ideally for another account you can sign in to
- Sign into the account you just mentioned with @ in the comment
- You should see a notification icon that you can click and it should say that you were mentioned.

## Before landing

1. PR has meaningful title

### DIFF
--- a/backend/.gitignore
+++ b/backend/.gitignore
@@ -3,3 +3,4 @@
 /gcloud-storage-credentials.json
 
 static/
+media/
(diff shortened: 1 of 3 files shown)


ddhanesha (approved):
Bug is fixed with this implementation. Approved!

PR #1041, opened 2022-08-19
Fix: Prevent creation of organizations with same name

### LINKED ISSUE #1040: Bug: Organizations can have the same name
**Describe the bug**
Organization names are supposed to be unique, this can be however by passed by changing the capitalization of letters in the name.

**To Reproduce**
1. Create an organization i.e "Test"
2. Attempt to create a second organization i.e. "test"
3. This will not cause an error

**Expected behavior**
The user should be notified that the name they are trying to use is already taken.

JoshKornfeld:
## Description
Closes #1040

Checking if name already exists changed to ignore case sensitivity i.e. if an organization uses the name "test", another organization cannot be created under the name "TEST".

Added a check when editing an organization name if that organization name is in use. If yes, then an alert will be shown and the user must choose another.

Implementation: 
Added a new view and  API call to the backend to lookup if the input matches an existing organization name or an existing translation of an organization name (basic interface). Additional checks are added for the case of creation in the detailed interface or editing.

## Additional
PR implements a change to project previews. Before displaying # of likes or comments, a check is made if either has at least 3.

## Test plan for creation (basic info step)

1. Go to create organization
2. For the name, enter a name of an existing organization and attempt to continue
3. An error should be thrown

## Test plan for creation (detailed info step)

1. Go to create organization
2. For the name, enter a new name of an organization and continue
3. You should be on the detailed organization creation interface
4. Change the name to one of an organization that is taken
5. Attempt to publish and and error should be shown

Additionally, for German page "veröffentlichen" should open the auto translate interface. Attempting to publish here with an organization name in use should throw an error.

## Test plan for editing

1. Choose an Org to edit
2. Enter the name of another org and attempt to save
3. An alert should pop up

## Before landing

1. PR has meaningful title


backend/organization/views/organization_views.py
```diff
@@ -378,6 +400,40 @@ def post(self, request):
...
+                    "message": "Organization with name {} already exists".format(
+                        query
+                    ),
+                    "url": organization_translation[0].organization.url_slug,
+                },
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        return Response(
+            status=status.HTTP_200_OK
+        )
```
ddhanesha, line 434: @JoshKornfeld why do you need to create an API for looking up existing organization?

backend/organization/views/organization_views.py
```diff
@@ -422,6 +478,27 @@ def patch(self, request, url_slug, format=None):
...
+                name_translation__iexact=request.data["name"]
+            ).exists():
+                return Response(
+                    {
+                        "message": "Organization with name {} already exists".format(
+                            request.data["name"]
+                        )
+                    },
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
```
ddhanesha, line 501: @JoshKornfeld I see the same logic multiple times repeated. Why don't you move this logic into a separate utility function and just call the function?

backend/organization/views/organization_views.py
```diff
@@ -422,6 +478,27 @@ def patch(self, request, url_slug, format=None):
...
+            if OrganizationTranslation.objects.filter(
+                name_translation__iexact=request.data["name"]
+            ).exists():
+                return Response(
+                    {
+                        "message": "Organization with name {} already exists".format(
```
ddhanesha, line 496: @JoshKornfeld are we showing this message in the front end? If yes, you should consult with @positiveimpact or @tobiasrehm for a better messaging verbiage. This error message feel more techie. Our user base is not necessarily technical.

ddhanesha (changes requested):

ddhanesha:
@JoshKornfeld please solve merge conflicts.

backend/organization/views/organization_views.py
JoshKornfeld: The previous implementation didn't handle case sensitivity, or check if the name existed as a translation for the name of another org. Since the previous implementation already makes an API call, I thought it could be suitable to make an API call that specifically does all the checks we need.

backend/organization/views/organization_views.py
JoshKornfeld: Changed the error message to something more descriptive.

backend/organization/views/organization_views.py
JoshKornfeld: Moved reused logic into separated utility functions.

backend/organization/views/organization_views.py
ddhanesha: Okay, that makes sense!

ddhanesha (changes requested):
@JoshKornfeld I am having some migration conflict on this PR. Can you resolve django migrations conflicts?

```python
(climateconnect_env) backend % ./manage.py migrate
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0086_auto_20220827_1839, 0086_auto_20220908_1358 in climateconnect_api).
To fix them run 'python manage.py makemigrations --merge'
```

JoshKornfeld:
Regarding migration files. To fix the merge conflict, I have deleted the commited migration file from this PR `0086_auto_20220908_1358.py` since it has the same body as `0086_auto_20220827_1839.py`. This gets rid of the conflict, but I'm not too sure if this is a safe/suitable solution. I tried to run `python manage.py makemigrations --merge` but ended up with some errors.

backend/organization/utility/organization.py
```diff
@@ -138,3 +140,40 @@ def add_organization_member(organization, user, user_role, role_in_organization)
...
+def check_existing_name_translation(name):
+    if OrganizationTranslation.objects.filter(
+            name_translation__iexact=name
+        ).exists():
+            return True
+    return False
+
+
+def get_existing_name_message(name):
+    return _("The name {} is already in use. Please use another name.").format(name)
```
positiveimpact, line 179: Here instead we should display something along these lines in the frontend "Someone has already created the organization <a href={link_to_org} target="_blank">{orgname}</a>. Please join the organization or use a different name. If you're having problems please contact support@climateconnect.earth.". That's a pretty long message but I think in the rare case that somebody tries to use an organization name that already exists we need to let them know that they can join the existing org if that was the one they were trying to create. Also we should display them a contact option incase somebody else created their organization without their permission.

positiveimpact (approved):
LGTM other than my comment

ddhanesha (dismissed):

backend/organization/utility/organization.py
JoshKornfeld: Made the change addressing this, however, I wasn't able to make the email a mailto link.

piperchester:
The CI should pass after we land #1156.

backend/organization/locale/de/LC_MESSAGES/django.po
```diff
@@ -72,9 +75,12 @@ msgstr "Du folgst diesem Projekt nicht mehr"
...
 
-#: organization/utility/organization.py:179
-msgid "The name {} is already in use. Please use another name."
-msgstr "Der Name {} wird bereits verwendet. Bitte verwende einen anderen Namen."
+#: organization/utility/organization.py:180
+msgid ""
+"Someone has already created the organization {}. Please join the "
+"organization or use a different name. If you're having problems please "
+"contact support@climateconnect.earth"
+msgstr "Jemand hat die Organisation bereits gegründet {}. Bitte trete der Organisation bei oder verwende einen anderen Namen. Falls du Probleme hast, wende dich bitte an support@climateconnect.earth"
```
positiveimpact, line 83:
```suggestion
msgstr "Jemand hat die Organisation {} bereits erstellt. Bitte trete der Organisation bei oder verwende einen anderen Namen. Falls du Probleme hast, wende dich bitte an support@climateconnect.earth"
```

backend/organization/views/organization_views.py
```diff
@@ -269,10 +274,22 @@ def post(self, request):
...
                     status=status.HTTP_400_BAD_REQUEST,
                 )
+        
 
+
+        if check_create_existing_name(request.data["name"]):
+            message = get_existing_name_message(request.data["name"])
+            return Response(
+                {
+                    "message": message,
```
positiveimpact, line 284: Here we could also send the url_slug of the existing org as an additional prop and then the frontend could add a link to the error message so users can look at the existing organization

frontend/pages/createorganization.js
```diff
@@ -146,19 +162,36 @@ export default function CreateOrganization({ tagOptions, rolesOptions, allHubs }
...
+      if (err?.response?.data?.message) {
         handleSetErrorMessages({
-          errorMessages,
+          ...errorMessages,
           basicOrganizationInfo: (
             <div>
-              {texts.an_organization_with_this_name_already_exists}
-              <a href={getLocalePrefix(locale) + "/organizations/" + org.url_slug}>
+              {texts.an_organization_with_this_name_already_exists}{" "}
+              <a href={getLocalePrefix(locale) + "/organizations/" + err?.response?.data?.url}>
```
positiveimpact, line 194:
```suggestion
              <a href={getLocalePrefix(locale) + "/organizations/" + err?.response?.data?.url} target="_blank">
```

positiveimpact:
LGTM! Great work :ok_hand:

Issue #1009, opened 2022-07-29
Bug: "Manage members" button when creating an organization


JoshKornfeld:
**Describe the bug**
On the 2nd step of creating an organization there's a button "manage members" and if you click it, it will say that you are not part of the organization and this feature is not available to you. Going back means you lose all your progress on creating the organization (also clicking the link doesn't help).

**To Reproduce**
1. Go to My profile / My organizations
4. Click on create an organization 
5. Fill out the forms (name, location)
6. In the second form press "manage members"

**Expected behavior**
The process of creating the organization shouldn't crash.

JoshKornfeld:
Closed by #1044

PR #1044, opened 2022-08-22
Fix: Omit manage members button when creating an organization

JoshKornfeld:
## Description
Solves #1009

- Removed manage members button from the edit organization interface.
- When creating an organization, the manage members interface will be opened after they click publish.
- Cancel button text is conditionally changed depending on whether or not the user is coming from the creation stage
- Updated alert message

## Test plan

1. Sign in
2. Go to create organization
3. Go to the detailed organization information portion, enter the necessary fields
4. Press publish
5. Changes should be visible on the manage members page

## Before landing

1. PR has meaningful title


positiveimpact (commented):
LGTM but I would still show the manage members button on the edit organization page if it's not in creation mode.

positiveimpact:
LGTM now, thanks for creating the issue! Great work :muscle:

positiveimpact (approved):

PR #1059, opened 2022-09-14
Uncomment automated email reminder task

ddhanesha:
## Before landing

1. PR has meaningful title


positiveimpact (approved):
LGTM, really looking forward to seeing this in action :heart_eyes: . Made a suggestion for what we could add in the future but for now we can just use Mailjet to track how many emails this feature sends out.

PR #1063, opened 2022-09-14
Fix: Search own name in inbox results

### LINKED ISSUE #1061: Bug: Searching inbox with characters in your own username
**Describe the bug**
If you enter a character that is in your own username in the inbox searchbar, it will not filter the chats and just reveal all of your chats.
E.g. My username is "James" and I have chats with "Jerry" and "Bob". If I type "J", I should only see "Jerry" but instead I still see "Bob".

**To Reproduce**
1. Sign in
2. Go to inbox
3. Click on "Find a chat"
4. Search your own name
5. See error (all chats are displayed)

**Expected behavior**
The user's username should not be included in the filtering. (See screenshot) "Test Group" & "Chris" should not be included

**Screenshots**
Screenshot from 2022-09-14 13-12-44

JoshKornfeld:
## Description
Closes #1061 

Added a .exclude() to the QuerySet which should trigger when the result uses the current user as a possible result.
Included changes from #1082 and suggestions to speed up search query.
In inbox.js removed loading statements that would cause the UI to snap to top every page. 

## Test plan
1. Sign in
2. Navigate to the inbox
3. Click "find a chat"
4. Search your own name
5. You should not see every possible open chat you have
## Before landing

1. PR has meaningful title


backend/chat_messages/views/message_views.py
```diff
@@ -179,40 +172,45 @@ def get_queryset(self):
...
-        
-        participants_matching_query = Participant.objects.annotate(
-            full_name=Concat('user__first_name', V(' '), 'user__last_name')
-        ).filter(
-            chat__in=chat_ids, full_name__icontains=query, is_active=True
+        ).values_list("chat", flat=True)
+
+        participants_matching_query = (
+            Participant.objects.annotate(
+                full_name=Concat("user__first_name", V(" "), "user__last_name"),
```
ddhanesha, line 199: You don't need to do this, check #1082

backend/chat_messages/views/message_views.py
```diff
@@ -179,40 +172,45 @@ def get_queryset(self):
...
+            )
+            .filter(chat__in=chat_ids, full_name__icontains=query, is_active=True)
+            .exclude(
+                chat__in=chat_ids,
+                full_name__icontains=current_user_full_name,
+                is_active=True,
+            )
         )
```
ddhanesha, line 207: You can reduce two queries we make here for Participant and MessageParticipants into one. You can use Prefetch Related: https://docs.djangoproject.com/en/4.1/ref/models/querysets/#prefetch-related

ddhanesha (approved):
Working as expected! Approved

PR #1066, opened 2022-09-16
Feat: Only show 2 active sectors on organization page

### LINKED ISSUE #1065: Only show the first 2 topics an organization has chosen and use a `show more` button to toggle the rest
If an organization is active in basically every field the page looks really bloated and you have to scroll really far to get to the actual content.

Only show the first 2 topics an organization has chosen and use a `show more` button to toggle the rest

**Additional context**
Screenshot from 2022-09-15 18-01-51

JoshKornfeld:
## Description
Closes #1065 

By default, only display 2 active sectors. Subsequent sectors are to be display if the user presses "show more" and can be hidden again by pressing "show less".

Additionally, the size of the HubPreviews has been slightly decreased when not in edit mode. And the user is provided extra information as to what exactly the HubPreviews are.

## Test plan
1. Visit an organization page
2. Scroll down to the sectors
3. Press "show more/show less"

## Before landing

1. PR has meaningful title


frontend/public/texts/organization_texts.js
```diff
@@ -279,5 +279,9 @@ export default function getOrganizationTexts({ organization, locale }) {
       en: "Skip for now",
       de: "Erstmal überspringen",
     },
+    organization_is_active_in_these_sectors: {
+      en: "This organization is active in these sectors:",
+      de: "Diese Organisation ist in diese Aktivitätsbereiche aktiv:",
```
positiveimpact, line 284:
```suggestion
      de: "Diese Organisation ist in diesen Bereichen aktiv:",
```

frontend/src/components/account/AccountPage.js
```diff
@@ -248,7 +260,34 @@ export default function AccountPage({
...
-            return <MiniHubPreviews hubs={i.value} />;
+            if (i.value.length <= 2) return <MiniHubPreviews hubs={i.value} />;
+            else {
+              const hubsToDisplay = showMoreSectors ? i.value : i.value.slice(0, 2);
+              return (
+                <>
+                  <div className={classes.subtitle}>
+                    {" "}
+                    {organizationTexts.organization_is_active_in_these_sectors}{" "}
+                  </div>
```
positiveimpact, line 271: Why do we only need the title incase there is more than 1 sector?

frontend/src/components/account/AccountPage.js
```diff
@@ -248,7 +260,34 @@ export default function AccountPage({
...
+                            {texts.show_less}
+                          </>
+                        ) : (
+                          <>
+                            <ExpandMoreIcon />
+                            {texts.show_more}
+                          </>
+                        )}
+                      </Button>
+                    </div>
```
positiveimpact, line 287: Maybe this could be incorporated into the MiniHubPreviews component. Then we could just add a `maxHubsToShow` prop and the complexity would be hidden and it could be reused.

positiveimpact (changes requested):
Great solution, would just do some restructuring.. Made some comments with things I would restructure

ddhanesha (commented):
Functionality is working as expected. @JoshKornfeld are all @positiveimpact comments addressed here?

positiveimpact:

PR #1067, opened 2022-09-19
Feat: Display child orgs/projects on organization page

### LINKED ISSUE #1064: Display child organizations and projects of child organization on the parent organization's page
When you create an organization you can choose that this organization is a sub-organization of another org.  For example the sustainability council of the city of Erlangen has over 10 forums which are sub-organizations. However when you go to[ their page](https://climateconnect.earth/de/organizations/NachhaltigkeitsbeiratderStadtErlangen162) you won't see any projects on there and you can't see the subforums. 
It would be much more user friendly if you could see the subforums and their projects directly as an overview on the Nachhaltigkeitsbeirat page.

On the organization page of a parent organization the child organizations should be listed and all their projects should show up in the projects section with a little badge containing the profile picture of the corresponding child org so you know who created this project.

JoshKornfeld:
## Description
Closes #1064 

Added backend API call to obtain an organizations sub-orgs. 
In the frontend, we obtain the sub-org projects and display them alongside the sub-orgs, should they exist. 

## Before landing

1. PR has meaningful title

### DIFF
--- a/frontend/pages/organizations/[organizationUrl].js
+++ b/frontend/pages/organizations/[organizationUrl].js
@@ -25,6 +25,7 @@ import UserContext from "./../../src/components/context/UserContext";
 import IconButton from "@material-ui/core/IconButton";
 import GroupAddIcon from "@material-ui/icons/GroupAdd";
 import ControlPointSharpIcon from "@material-ui/icons/ControlPointSharp";
+import OrganizationPreview from "../../src/components/organization/OrganizationPreview";
 
 const DEFAULT_BACKGROUND_IMAGE = "/images/default_background_org.jpg";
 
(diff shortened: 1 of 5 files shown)


frontend/pages/organizations/[organizationUrl].js
```diff
@@ -67,25 +68,45 @@ const useStyles = makeStyles((theme) => ({
...
+
   return {
     props: nullifyUndefinedValues({
       organization: organization,
       projects: projects,
       members: members,
       organizationTypes: organizationTypes,
       rolesOptions: rolesOptions,
+      childOrganizations: childOrganizationInfo?.organizations?.map((o) =>
+        nullifyUndefinedValues(o)
```
Comment (positiveimpact) [inline frontend/pages/organizations/[organizationUrl].js:107]: Why is this necessary?

frontend/pages/organizations/[organizationUrl].js
```diff
@@ -313,6 +362,45 @@ async function getProjectsByOrganization(organizationUrl, token, locale) {
   }
 }
 
+async function getChildOrganizationInfo(organizationUrl, token, locale) {
+  const orgs = await getChildOrganizations(organizationUrl);
+  if (orgs != null) {
+    const projects = await getChildProjects(orgs, token, locale);
```
Comment (positiveimpact) [inline frontend/pages/organizations/[organizationUrl].js:368]: This is not a good idea for performance to wait for one API call to finish and then add another. This needs to either be done in one API call or the projects need to be caught afterwards (The first option is better for SEO since Google can then find the projects on the page)

Comment (positiveimpact) [reply frontend/pages/organizations/[organizationUrl].js:368]: The best solution here is probably to alter the endpoint that catches the projects to also return the child projects. Because then pagination would also work as expected (e.g. only the first 10 projects will be shown without scrolling down)

positiveimpact (commented):
Overall the approach looks good to me except my comments. One interesting question will be how this generally impacts performance since we're retrieving much more data (especially on large orgs with many local groups such as FFF). But I still think it makes sense to catch the child orgs and projects directly rather than after loading the page for SEO reasons

ddhanesha:
@JoshKornfeld please confirm that all @positiveimpact comments are addressed. 

Can you resolve merge conflicts?

PR #1080, opened 2022-09-29
Bugfix: Celery tasks for private message email notifications.

ddhanesha:
## Description

PR fixes the bug found celery task that automatically sends email notifications to climate connect users. 

## Before landing

1. PR has meaningful title


ddhanesha:
Merging this so we can verify if the fix works today.

PR #1097, opened 2022-10-17
Feat: Organization Social Media Links (new version from master)

### LINKED ISSUE #1071: Add social media links to organization page
Add social media icon buttons to the organization page that when clicked, direct the user to their selected social media page(s).
(see organization page UI link).

List of changes for this to be implemented:
Backend:

- Add to the organization model and to the organization translation a field for "social media links".
- Update the view for creating/editing.
- Update serializers if needed.

Frontend:

- In the creation/editing stage, users should be able to add/remove social media links.
- The organization UI page should reflect this change and the social media icons should be visible to users visiting an organization page.

**Additional context**
UI designs can be found here:
Organization page UI:
https://xd.adobe.com/view/861b3ade-6eba-45d8-beb3-506e7afd7699-d348/ (slide 3)
Creation step 2 / Edit:
https://xd.adobe.com/view/b12f4514-bd4b-47b7-b21c-0e4922baee92-e47c/screen/8e3244dd-d74b-4149-b4f6-f6e7672422e6/

JoshKornfeld:
## Description

Closes #1071

Backend: 
Added a SocialMediaChannel & SocialMediaLink model. A channel is a constant that points to a social media site. The link binds this channel to an organization.  
Updated serializers: organization.py and created new ones in social_media.py. Modified organization_views for creating/editing organizations.

Frontend:
Users can click a + icon on a chip that reads add social media. This opens a selectfield dialog from which they can pick from 5 different social medias. Depending on the type of social media, the user will need to either insert the full link or just the handle. A second textfield will open and will display the full URL for the user if only the handle is required. Once added, a chip dispaying the icon and name, with the option to delete it via a close button, will appear on the edit page. When the chip is hovered, the url will display. When viewing the organization page, the social media links will be displayed via their icons that when clicked, will open a new tab with their social media page opened.
The website of the organization has been moved from the center to the left under their avatar. 
Also fixed where selecting empty field, the page doesn't crash for types or social media options.

TODOS: 
Select field for German page has strange formatting. (dialog has a scroll, texts are too long).
Side note: for organization types, I don't believe whatever is being input into additional information is being saved to the backend 

## Test plan
1. Create an organization
2. Go through the steps and in the detailed input screen, chip should be visible. Clicking the + icon should open a dialog in which you can select a type of social media write a valid url for that specific social media.
3. Click create after entering everything
4. Click "edit organization"
5. You can now also delete or modifiy existing urls

## Before landing

1. PR has meaningful title


frontend/src/components/dialogs/SelectDialog.js
```diff
@@ -38,35 +39,80 @@ export default function SelectDialog({
...
+       setElement(typeAsRequired);
+    }
+
     if (supportAdditionalInfo) {
       const value = values.filter((val) => val.name === event.target.value)[0];
-
-      if (value.additionalInfo.length > 0) {
+      if (isSocial) {
+        setTextLabel(
+          value?.ask_for_full_website ? value?.name : "Please enter your " + value?.name + " handle"
```
positiveimpact, line 107: 1. We should use a translation file for this. I'd recommend creating a getSocialMediaTexts file. Also for the website I'd also put "Please enter your {channel} url", otherwise it's a bit confusing imo.

frontend/public/texts/organization_texts.js
```diff
@@ -279,6 +279,39 @@ export default function getOrganizationTexts({ organization, locale }) {
...
+      en: "Select a social media",
+      de: "Wähle ein Soziales Medium aus",
+    },
+    find_us_here: {
+      en: "Find us here:",
+      de: "Finde uns unter:",
+    },
+    does_not_comply_twitter: {
+      en: "Does not comply with a valid Twitter URL",
+      de: "Entspricht nicht einer gültigen Twitter-URL",
```
positiveimpact, line 297:
```suggestion
      en: "This is not a valid Twitter URL",
      de: "Das ist keine gültige Twitter-URL",
```
Would do the same for all the things below as well

frontend/public/lib/socialMediaOperations.js
```diff
@@ -0,0 +1,138 @@
...
+      error = matches ? "" : texts.does_not_comply_instagram;
+      break;
+    case "Facebook": // facebook
+      regex = new RegExp(regexPrefix + "facebook" + regexSuffix);
+      matches = regex.test(url);
+      error = matches ? "" : texts.does_not_comply_facebook;
+      break;
+    default:
+      break;
+  }
```
positiveimpact, line 46: 1. We should use an object and a generic function rather than switch.case here. Switch case is not meant for handling a bunch of cases that are the same except one string.
2. All of these should pass and be sent to the backend as https://twitter.com/test: "twitter.com/test", "www.twitter.com/test". The user should not be required to type out a prefix.

frontend/src/components/account/AccountPage.js
```diff
@@ -399,6 +402,32 @@ export default function AccountPage({
               ))}
             </Container>
           )}
+          {isOrganization && account.info.social_options.length > 0 && (
+            <>
+              <div className={classes.website}>
+                <Typography variant="caption"> {organizationTexts.find_us_here} </Typography>
```
positiveimpact, line 408:
```suggestion
             
```
I feel like this is not really necessary

frontend/public/lib/socialMediaOperations.js
```diff
@@ -0,0 +1,138 @@
...
+  // matches http://, https:// , https://www. ,http://www.
+  const regexPrefix = "^(http)(?:s)?(://)(?:www.)?";
+  // matches.com/ anything
+  const regexSuffix = ".com/.+$";
+  let regex;
+  let matches = false;
+
+  switch (socialChannel.name) {
+    case "Twitter": // twitter
+      regex = new RegExp(regexPrefix + "twitter" + regexSuffix);
```
positiveimpact, line 20: instead of hardcoded options we should use the base url here. Otherwise this will break if we add a new social (e.g. TikTok) in the future. It should be possible to add a new social without making changes to the code

positiveimpact (commented):
Added some comments

frontend/pages/organizations/[organizationUrl].js
```diff
@@ -112,7 +112,7 @@ export default function OrganizationPage({
           organization={organization}
           projects={projects}
           members={members}
-          organizationTypes={organizationTypes}
+          organizationTypes={organizationTypes} // not sure this does anything (not used in organization layout props)
```
Comment (positiveimpact) [inline frontend/pages/organizations/[organizationUrl].js:115]: What happens if you remove it? Are the organization types still shown then?

frontend/public/lib/socialMediaOperations.js
```diff
@@ -0,0 +1,138 @@
...
+        socialMediaLinks.push(createSocialMediaIconButton(social));
+        break;
+      case "Instagram": // instagram
+        socialMediaLinks.push(createSocialMediaIconButton(social));
+        break;
+      case "Facebook": // facebook
+        socialMediaLinks.push(createSocialMediaIconButton(social));
+        break;
+      default:
+        break;
```
positiveimpact, line 70: Why do we have 5 cases that all execute the exact same code? :D

frontend/public/lib/socialMediaOperations.js
```diff
@@ -0,0 +1,138 @@
...
+      method: "get",
+      url: "/api/social_media_channels/",
+      locale: locale,
+    });
+
+    if (resp.data.results.length === 0) return null;
+    else {
+      return resp.data.results.map((t) => {
+        return {
+          name: t.social_media_name, // need to refactor name for the selectfield component to work
```
positiveimpact, line 126: These comments are very helpful, great work! :+1:

frontend/src/components/general/SocialMediaButton.js
```diff
@@ -16,8 +17,17 @@ const useStyles = makeStyles((theme) => ({
   }),
 }));
 
-export default function SocialMediaButton({ href, socialMediaIcon, altText, isFooterIcon }) {
-  const classes = useStyles({ isFooterIcon: isFooterIcon });
+export default function SocialMediaButton({
```
positiveimpact, line 20: The naming "Button" is a bit counter-intuitive here since it seems to just be a link

JoshKornfeld: It's really more of an iconbutton so perhaps we could use that for the name? i.e. SocialMediaIconButton

frontend/pages/organizations/[organizationUrl].js
Comment (JoshKornfeld) [reply frontend/pages/organizations/[organizationUrl].js:115]: The function that is called "getOrganizationTypes" simply returns [] and is not the way types are returned. Types are obtained via API call listing the organization via its url slug. Furthermore we see it is passed as a prop to OrganizationLayout, but OrganizatonLayout does not have it as a parameter, so I'd suggest removing it.

frontend/src/components/dialogs/SelectDialog.js
JoshKornfeld: File added with translations and suggested text.

frontend/src/components/account/AccountPage.js
JoshKornfeld: Added different styling. Additionally, updated the logic so the text "find us here" only shows if there is a social link or website to display.

frontend/public/lib/socialMediaOperations.js
JoshKornfeld: Refactored, removed unnecessary function call.

frontend/public/lib/socialMediaOperations.js
JoshKornfeld: Using baseUrl instead of hardcoded string.

ddhanesha:
@JoshKornfeld please resolve all merge conflicts.

PR #1108, opened 2022-10-22
Improving organization cards to load faster

ddhanesha:
## Description
PR includes improvements to organization cards in the browse page. These are backend changes made to improve API load. 

## Test plan
It would hard to test this locally but try login as a user and explore organization page.

## Before landing

1. PR has meaningful title

### DIFF
--- a/backend/organization/views/organization_views.py
+++ b/backend/organization/views/organization_views.py
@@ -99,7 +99,7 @@ class ListOrganizationsAPIView(ListAPIView):
                 url_slug=self.request.query_params["hub"]
             )
             if hub.exists():
-                if hub[0].hub_type == Hub.SECTOR_HUB_TYPE:
+                if hub.first().hub_type == Hub.SECTOR_HUB_TYPE:
                     project_category = hub.first().filter_parent_tags.all()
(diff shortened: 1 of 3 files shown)


backend/organization/views/organization_views.py
```diff
@@ -84,15 +84,23 @@ def get_serializer_context(self):
...
-        organizations = Organization.objects.all()
+        organizations = Organization.objects.all().prefetch_related(
+            Prefetch(
+                'tag_organization',
+                queryset=OrganizationTagging.objects.select_related(
+                    'organization_tag'
+                )
+            ),
+            'organization_member'
+        ).select_related('language', 'location')
```
positiveimpact, line 95: Should we make this dependent on which query_params were passed? If I understand this correctly, this PR would significantly improve query performance for queries filtering by hub or location but would do unnecessary additional work when simply loading the browse page without filters.

positiveimpact: Generally this can be extremely useful though, e.g. for the project page

ddhanesha: Discussed offline, these tables are used in serializer so it will be helpful to prefetch them with or without filters.

PR #1126, opened 2022-11-10
Convert Frontend to TypeScript

phiresky:
## Description

This PR converts the frontend code from JavaScript to TypeScript.

It tries to keep all semantics the same, i.e. the running frontend should not change in any way.

NextJS has builtin TypeScript support so the compilation / deployment process should stay the same.

## Todo

- [x] Fix all remaining typescript compile errors (most of them should probably be fixed by just adding `any` types where relevant to keep the total effort manage)
- [x] Add typescript compilation to CI

## Future work

After this PR lands, the following this should also be done

- [ ] Add typescript-eslint (including type-checking rules and recommended rules)
- [ ] incrementally remove more instances of any and don't add new ones
- [ ] enable noImplicitAny in tsconfig.json 

## Before landing

1. PR has meaningful title


ddhanesha:
@phiresky can you share pros and cons of using typescript compare to regular javascript? How much learning curve would it be for our interns and volunteers? 

I do not mind using typescript but in my experience, it has a little bit a learning curve so we should make sure that typescript really adds good value and make our development faster. If it slows down our development and increase debugging time, we do not want to that right now.

phiresky:
For me the main benefits would be

- being able to understand existing code much faster without having to think much - it's easier to understand what's happening with types
- more confidence in the code doing what i expect it to because it's explicit what's being passed around and returned
- better IDE support - VSCode is much better at auto-complete, immediately showing issues, etc with TS than with JS
- less bugs, less debugging

specific examples i found in this code base (without much effort):

 - Many React components in this code base receive props that they don't actually use. For example HubHeaderImage does not use the `alt`-text that's passed here: image (TS shows it as an error)
 - Multiple of the MUI styles definitions don't actually have classes that are used in the code, e.g.: 
image
(HoverButtonButton does not exist)
- There were multiple instances of `position: "auto"` in the code base. But "auto" is not a valid CSS value for `position`.
- `parseOrganizationForRequest` is defined as `(o, user, rolesOptions, translations, sourceLanguage) => `. But it's used like this:
image and like this in another place: 
image. Looks like an argument was added or removed at some point. Here it probably doesn't matter much, but it's easy to mess this kind of thing up without typescript

cons / issues:

- migration effort
    - possible merge conflicts until everything is migrated
    - benefit of TS is lower when many things are `any`-typed which they will be in the beginning

- It's a bit of work to understand and learn on top of JS. 
    
    In my opinion this is only a small issue for most usages of typescript that consist of simple types (i.e. objects, numbers, strings, booleans). If a function gets `{foo: number, bar: string}` as an argument you already have to know that whether you're using typescript or javascript, typescript just makes it explicit. The syntax of these types is pretty easy to learn and if you know JS you already know 90% of TS.

    The main issue here is that TypeScript has *can* handle very complex type definitions (generics, mapped types, conditional types, ...). Specifically in this code base the main example I found was the `makeStyles` function from MUI. It's not intuitive how typing this function works. It's also always the same though so can be "copied" from the other usages.
- Time spent writing and figuring out types. For simple types again this is basically just typing time because even with JS you have to know what's being passed around, you're just not writing it down. For complex types this can become an actual issue especially for less-experienced TS devs

company stories:

- airbnb: 38% of bugs preventable with typescript. https://www.youtube.com/watch?v=P-J9Eg7hJwE
- Slack: "Managing large JavaScript codebases is challenging — [...]. To that end, we adopted TypeScript (a statically typed superset of JavaScript) and quickly learned to stop worrying and love the compiler. It’s not just us, either: In the 2017 StackOverflow Developer Survey, TypeScript was the third most-loved programming technology. Given how quickly static type checking is gaining traction, we wanted to share our experiences and practices." https://slack.engineering/typescript-at-slack/
- Strip: "Migrating millions of lines of code to TypeScript" https://stripe.com/blog/migrating-to-typescript
- Sentry: "Slow and Steady: Converting Sentry’s Entire Frontend to TypeScript" https://blog.sentry.io/2021/04/12/slow-and-steady-converting-sentrys-entire-frontend-to-typescript/

Generic quote from one of the company stories:

*    Several classes of bugs could be detected — and eliminated — during compilation.
*    We could improve the developer experience through editor integrations such as auto-completion, faster code navigation, and inline compiler feedback.
*    We could reduce the need for API documentation, as type annotations help produce self-describing code.
*    TypeScript has an active community with a clear and maintained development roadmap in addition to rapid releases.
*    TypeScript can be adopted incrementally. That meant we can start writing new code with TypeScript and incrementally convert over time.

phiresky:
By the way when this one is done we can immediately start adding static type checking of the backend using [mypy](https://github.com/python/mypy)  🤣

phiresky:
if you want more of a meme answer:

Javascript already has all the types. Without TS you're just pretending they don't exist

Benefit for interns/volunteers is also that it looks good on the resume. Is anyone actually still hiring javascript devs? 👀 according to this rando survey 70% of js devs use typescript https://2021.stateofjs.com/en-US/other-tools/#javascript_flavors

ddhanesha:
> By the way when this one is done we can immediately start adding static type checking of the backend using [mypy](https://github.com/python/mypy) 🤣

Haha @phiresky I would rather do mypy migration first. I have slowly started using typing for new functions but wouldn't mind if you make a PR for the existing code.

ddhanesha:
> migration effort
> 
>     * possible merge conflicts until everything is migrated
> 
>     * benefit of TS is lower when many things are `any`-typed which they will be in the beginning

Thank you for the detailed pros and cons list. Yes, I have talked to people working in the companies you mentioned and were part of the migration effort in the bay area. They all have positive things to say about typescript. But, the main concern I have is the migration effort it takes. These tech companies had a big team and a lot of engineers working on the migration. We do not have that leverage here. Looking at what we have in near future work at climate connect, I want our volunteers and full time members to spend time on developing/updating key features.

Having said that, I do not mind testing and merging this PR. But when it comes to migration of our existing code, my recommendation would be to do that process in an organized manner. Maybe in the beginning every type is set to `any` for existing code. Any new development we do should be in typescript but I wouldn't migrate everything to typescript right now. Having done multiple javascript/typescript and python migrations, I would not rush the migration process.

phiresky:
> the migration effort it takes

Sure, makes sense. I'd be willing to champion the migration and to answer any questions the others might have regarding TS.

> my recommendation would be to do that process in an organized manner. Maybe in the beginning every type is set to any for existing code

> but I wouldn't migrate everything to typescript right now

I'm not sure exactly what you mean by these two. Do you mean you'd keep most files as `.js` and not annotate them at all, or rename them to `.ts` and mark any types that aren't automatically correctly inferred as any?

My suggestion for the migration plan would be as follows:

1. Make the build process understand TS in general and add deps etc to support TS code (done in this PR)
2. Rename all existing js code to typescript and ensure the compilation still works (should already be done in this PR as well). There will be compile errors but nextjs (and tsc) will still emit the same JS code regardless
3. Fix all TS compile errors, mainly with trivial changes (add `any` types) (partially done in this pr, still some missing though)
4. For new code: All code should be in TS files, ideally with correct types, but `any` annotations would also be fine at least in the start.
5. Add lints that ensure no new JS code and code quality for TS code

ddhanesha:
@phiresky,
re: I'm not sure exactly what you mean by these two. Do you mean you'd keep most files as .js and not annotate them at all, or rename them to .ts and mark any types that aren't automatically correctly inferred as any?

Yes, rename all to `.ts` and types aren't automatically update refer them as `any`. 


What I mean by "Do not rush" is that lets migrate every files to typescript but update types on existing code in smaller chunks. I hope that makes sense.

phiresky:
Makes sense. I can fix most of the remaining TS errors. Could you approve the workflow runs?

ddhanesha:
> Makes sense. I can fix most of the remaining TS errors. Could you approve the workflow runs?

@phiresky I think I can approve workflows after the merge conflicts have been resolved.

positiveimpact:
+1 on merging this PR once the listed TODOs are done, writing new code in ts and slowly migrating existing code to ts as well. Is this actually ready for review already since there are still two open TODOs?

ddhanesha:
@positiveimpact that is my fault I accidentally clicked on ready for review button.

phiresky:
While fixing errors, I've found many instances of react props not being used or the style class not being defined (or there being a spelling mistake in the property name). I'm commenting them out and marking most of them "TODO(unused)" or "TODO(undefined)", but they can probably all just be removed unless I'm making mistakes.
image

phiresky:
I have fixed all the remaining TS errors. I've also added yarn format check and typescript compile check to the CI and fixed that

ddhanesha:
@phiresky the UI on main page is messed up. Here are the screenshots. 
Screenshot 2022-11-20 at 7 51 49 PM
b.png">

piperchester:
Very excited for this PR! Thanks @phiresky for this monumental lift!

phiresky:
The project page issue should be fixed with https://github.com/climateconnect/climateconnect/pull/1126/commits/bb97e7da1600ac8d57aa56e3b18506fb110eec4e . Are there any other issues?

phiresky:
# Tests:

**User**

- [ ] Sign Up
- [ ] Log In

**Projects**

- [ ] create project
- [ ] view project
- [ ] edit project

**Organizations**

- [ ] create organization
- [ ] view organization
- [ ] edit organization

**Members**

- [ ] create member
- [ ] view member
- [ ] edit member

**Ideas**

- [ ] create idea
- [ ] view idea
- [ ] edit idea

**Hubs**

- [ ] view hub http://localhost:3000/hubs/food
- [ ] filter projects in hub

**Other**

- [ ] switch language de/en

- [ ] private chat
- [ ] group chat
- [ ] /admin page
- [ ] send and receive notification

ddhanesha:
@phiresky can you resolve conflicts with master branch? I will review and test this PR today?

phiresky:
I've fixed the merge conflicts and tested most site functionality except chat and ideas

positiveimpact:
First of all huge apologies for taking so long to finally test this PR!
I merged master and tested everything now and everything seemed great except these two bugs:
- `MentionsInput` has a scrollbar in the comments section on the project project page - even if it's empty
- On a hub page and the browse page, the tab that you didn't start on won't load. E.g. if you load the browse page and switch to the organizations tab it should show a loading animation and then load the organizations. Now it just says there are no organization. But if you reload the page while you're in the organization tab the organizations will load but the projects and members won't load

positiveimpact (approved):
LGTM now! Thanks @phiresky for this important contribution and I'm looking forward to slowly making the code "proper" typescript

PR #1142, opened 2022-11-16
Feat: Reset account page images

### LINKED ISSUE #678: Implement feature to remove background image from editing profile
Currently we don't have a button to remove a user's background image if they add it to their profile. This would be a very small feature to implement that functionality.

Note that this could be split a backend / frontend tickets, but I think it's small enough that anyone that takes this on could own the small feature end to end.

JoshKornfeld:
## Description

Closes #678 

Background image:
Can be reset by clicking an X iconbutton in top right of the image. This button will only show if the account has a background image. Applies to both profiles and organizations.

Avatar image:
If profile:
Avatar has a button "edit" which when clicked offers different other buttons such "edit avatar", "remove avatar" and "add avatar", depending on what is currently selected.
If organization: 
Avatar has a button which allows the user to edit the avatar, if none is selected i.e. creation step it will say add avatar instead.
Screenshot from 2022-11-16 09-56-18

## Test plan

1. Go to edit organization / edit profile to see changes

## Before landing

1. PR has meaningful title


positiveimpact:
@MarcoRehmer If I remove my avatar, my avatar is stilll there after I save. Is this the same for you? Should I have a look at the backend code?

MarcoRehmer:
> 

hm, I tested that. Give me a moment, I take a look at it

frontend/src/components/account/AccountPage.tsx
```diff
@@ -472,10 +451,20 @@ export default function AccountPage({
           <div className={classes.avatarContainer}>
             {account.badges?.length > 0 ? (
               <ProfileBadge badge={account.badges[0]}>
-                <Avatar {...avatarProps} />
+                <UserAvatar
+                  mode={"read"}
+                  isNarrowScreen={isSmallScreen}
```
positiveimpact, line 456: I'm not sure how much sense it makes to pass this as a prop because the component could also just retrieve the screensize itself. I'm not sure what the best practise is here.

positiveimpact, line 456: I'm not sure how much sense it makes to pass this as a prop because the component could also just retrieve the screensize itself. I'm not sure what the best practise is here. Same goes for other places with UserAvatar

frontend/src/components/account/EditAccountPage.tsx
```diff
@@ -378,9 +327,10 @@ export default function EditAccountPage({
           ))}
           {editedAccount.info[key].length < infoEl.maxEntries && (
             <Chip
-              label="Add"
+              label={texts.add}
```
positiveimpact, line 330: Thanks for fixing those in many places!

frontend/src/components/account/UserAvatar.tsx
```diff
@@ -0,0 +1,194 @@
...
+    imageUrl: props.imageUrl,
+    thumbnailImageUrl: props.thumbnailImageUrl,
+  });
+
+  const onImageChanged = async (avatarEvent) => {
+    const file = avatarEvent.target.files[0];
+    if (file && file.type) {
+      if (!ACCEPTED_IMAGE_TYPES.includes(file.type)) {
+        alert(texts.please_upload_either_a_png_or_a_jpg_file);
+      }
```
positiveimpact, line 90: Is this still necessary with the input only accepting png and jpeg?

positiveimpact (approved):
LGTM except the minor comments I made :)
Great job on finishing up the PR and cleaning up the code in the process. It's much simpler and more readable now, thank you!

positiveimpact (approved):
LGTM except the minor comments I made :)
Great job on finishing up the PR and cleaning up the code in the process. It's much simpler and more readable now, thank you!

frontend/src/components/account/AccountPage.tsx
MarcoRehmer: yep, though that too, its an appendix from refactoring :D I will find a better solution

frontend/src/components/account/UserAvatar.tsx
MarcoRehmer: sounds legit. If the format of the image is somehow "currupted", the backend should still check the incoming image for security reasons

PR #1149, opened 2022-11-18
Feat: Label indicating a reply must be sent in chatbox

### LINKED ISSUE #1150: Add a label to inbox indicating if the user was not the last one to write a message
If a user was the last one to receive a chat message, they should be informed that they still need to reply to the message.

JoshKornfeld:
## Description

Closes #1150 

This PR implements a label onto chat previews for messages in which the logged in user was not the last one to write a message. Should a user not want to see the labels, they can use a slider at the top of the inbox to turn them off. A cookie will track this state for 30 days.

The label shows under the condition that the user was not the last one to send a message, or the chat has not had any messages sent in it yet.

The new label is only implemented for non mobile screens.

Interface before:
Screenshot from 2022-11-21 14-52-57

Screenshot displays new interface.

Screenshot from 2022-11-21 14-33-38

## Test plan

1. Go to inbox
2. Send a message to yourself and see the label appear

## Before landing

1. PR has meaningful title


frontend/src/components/communication/chat/ChatPreviews.js
```diff
@@ -59,16 +73,27 @@ const useStyles = makeStyles((theme) => {
...
 
 export default function ChatPreviews({ chats, loadFunc, hasMore, chatSearchEnabled }) {
   const classes = useStyles();
-  const { locale } = useContext(UserContext);
+  const { user, locale } = useContext(UserContext);
   const texts = getTexts({ page: "chat", locale: locale });
-  const [isLoading, setIsLoading] = React.useState(false);
+  const [isLoading, setIsLoading] = useState(false);
   const isNarrowScreen = useMediaQuery((theme) => theme.breakpoints.down("xs"));
+  console.log(user);
```
JoshKornfeld, line 96:
```suggestion

```

ddhanesha (changes requested):
You made a good try but I don't think we need to add a role column to user_profile table to show a label. There is a better way to show the label. You fetch user_id of the last message sent in a chat. Then compare the id of the user calling the API and user_id of the last message. If the user_id does not match then you show the message.

JoshKornfeld:
> You made a good try but I don't think we need to add a role column to user_profile table to show a label. There is a better way to show the label. You fetch user_id of the last message sent in a chat. Then compare the id of the user calling the API and user_id of the last message. If the user_id does not match then you show the message.

We wanted to make this only available to certain users like Thomas, as this is a tool for community building. If this was for regular users we would need to add a way to dismiss the label for the cases in which their conversation is over for the time being and the label may become annoying to those users.

ddhanesha:
> > You made a good try but I don't think we need to add a role column to user_profile table to show a label. There is a better way to show the label. You fetch user_id of the last message sent in a chat. Then compare the id of the user calling the API and user_id of the last message. If the user_id does not match then you show the message.
> 
> We wanted to make this only available to certain users like Thomas, as this is a tool for community building. If this was for regular users we would need to add a way to dismiss the label for the cases in which their conversation is over for the time being and the label may become annoying to those users.

I don't think we established that this requirement should only be for climate connect community building. I believe this can be a very useful tool for everyone on the platform.

JoshKornfeld:
@ddhanesha I agree that this feature could be useful to regular users. I've removed the role column and instead introduced a slider which state is stored in a cookie. Hopefully, this will enable users to also use this feature, without annoying users who do not want to use it. 
In the prior implementation, I had discussed with @positiveimpact what to do about hiding the label for users who didn't need it i.e. in the case a chat had some to its natural conclusion for the time being. We concluded that we could just make this available to certain users like Thomas and avoid needing the ability to turn it off, hence the introduction of the role column. I hope this is a good middle ground :)

positiveimpact:

PR #1154, opened 2022-11-22
Feat: Additional inbox filtering for community building

### LINKED ISSUE #1155: Climate Connect inbox: filtering by "need to reply" messages (that would be for example for me to see all the conversation where the other has been the last one to text me and potentially waiting for an answer from me)
### **_Priority A_**

Key requirements:
* Show tag that says "Need to reply" or "Your turn". ( implemented, at least to some extent by PR #1149
* Filter messages by last message received by the user (Need to reply)
* Filter messages by location

JoshKornfeld:
## Before merging
Review and approve #1149, so we can use the updated labels.

## Description
Closes #1155

This PR aims to add filtering to users of type moderator or admin. Types of filtering are
- Search chats by looking up users by their location. This means that if a location is searched, a list of all chats where participants are from the searched location should be displayed.
- Search chats by looking up if the user was not the last person to send a notification. A list of all chats labeled "Need to reply" from #1149 should be displayed.

Changes backend:
2 new endpoints for each filter.
Refactoring of reused code blocks into functions. 

Changes frontend:
A button to control a menu of 2 extra buttons. When clicked, the 2 extra buttons should be shown or hidden. The 2 extra buttons are used to filter the chat messages by location or the need to reply.
Refactored `inbox.js` into multiple other components. 

## Test plan

1. Sign in on an account with moderator / admin rights
2. Go to inbox
3. Click the arrow icon to open more options
4. Select either search via location or filter by need to reply
5. If location is selected then enter a location into the searchbar, results will load once a location has been selected from the menu
6. If need to reply is selected, the chats will be loaded and displayed

## Before landing

1. PR has meaningful title


ddhanesha (changes requested):
I do not think this is a right way to approach this problem. You are writing different APIs just for filters, which could easily be done by adding filter logic to the existing API. I do not think we had ever established that this requirement is limited to community building moderator or administrator. 


For this specific requirement, I think all you need is a UI the can filter location and `need to reply`. On selecting filter you can call the existing API with query parameters to apply selected filter. I hope this makes sense. I humbly reject this PR.

JoshKornfeld:
Thanks for the feedback. As requested, removed new API calls and moved logic into existing one :) @ddhanesha

positiveimpact:

PR #1160, opened 2022-12-01
Update startup scripts that're used in Azure

piperchester:
## What and Why

Originally I thought these weren't  actually being used, but they're used in Azure I believe. Updated docs to point to that.


positiveimpact (approved):
LGTM, thanks for this addition!

piperchester:
Just to confirm - is `start_frontend.sh` also being used?

positiveimpact:
> Just to confirm - is `start_frontend.sh` also being used @positiveimpact?

yep

PR #1165, opened 2022-12-04
fix docker-compose

piperchester:

### DIFF
--- a/backend/docker-compose.yaml
+++ b/backend/docker-compose.yaml
@@ -1,4 +1,7 @@
-redis:
-    image: redis:6
-    ports:
-      - "6379:6379"
+# version: '2'
+
+services:
+  redis:
+      image: redis:6
+      ports:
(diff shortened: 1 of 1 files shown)


backend/docker-compose.yaml
```diff
@@ -1,4 +1,6 @@
-redis:
-    image: redis:6
-    ports:
-      - "6379:6379"
\ No newline at end of file
+version: '3.8'
+services:
```
piperchester, line 2: It's weird that others haven't run into this sooner - @positiveimpact or @ddhanesha had you?

ddhanesha: @piperchester I think we have to change how this work. Maybe we can move the entire local setup to docker.

ddhanesha: @piperchester Can you create an issue for this?

PR #1166, opened 2022-12-04
Fix Azure web app deploy step on merge

piperchester:
## What and Why

Noticed that we're getting a failure that looks like this

Screenshot 2022-12-04 at 7 58 48 AM

Following https://github.com/Azure/webapps-deploy/issues/98 and then https://github.com/projectkudu/kudu/issues/3042#issuecomment-665659432, it looks like the solve will be scaling up our service plan.

This PR cleans up some formatting in the pre-existing deploy file.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,4 @@
-# Weekly dependency upgrades to the develop branch.
+# Weekly dependency upgrades.
 #
 # Read more about the config here:
 # https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#target-branch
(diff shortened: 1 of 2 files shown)


positiveimpact (approved):
LGTM! This error hasn't happened for a while but sometimes seems to come a bit randomly. Currently we're just within the sponsored credits we get from Azure each year so we would only scale up if really necessary.

positiveimpact:
> this issue still exists. i just got it when trying to deploy a python project via github.

Did you try to deploy this project or a different one?

Ammar-Azman:
Just happened to me. While it seems weird, the the web app is working well when I clicked to the website link. But on Git Action it showed 
```
Error: Failed to deploy web package to App Service.
Error: Deployment Failed, Error: Failed to deploy web package to App Service.
```

PR #1170, opened 2022-12-05
Fix: Sitemap for google indexing

### LINKED ISSUE #1139: Fix sitemap so that german pages are indexed correctly on Google
A large number of our pages is currently not indexed on Google. This is likely because we don't correctly tell Google which english and german pages belong together. 
[Here](https://climateconnect.earth/sitemapindex.xml are our current sitemaps)

And [here](https://developers.google.com/search/docs/specialty/international/localized-versions) Google describes how it needs to be done.

We need to adapt our current sitemap generator script to meet this standard and then resubmit the pages in the search console.

JoshKornfeld:
## What and Why

Closes #1139

PR follows changes laid out by the google documentation in the issue. Implementation can be improved i.e. not easily expandable for other languages, but wanted to at least get it working. Looking forward to some feedback :)

### DIFF
--- a/frontend/pages/sitemap/[language_code_dot_xml].js
+++ b/frontend/pages/sitemap/[language_code_dot_xml].js
@@ -83,7 +83,7 @@ async function createSitemap(
   }
   const BASE_URL = process.env.BASE_URL ? process.env.BASE_URL : "https://climateconnect.earth";
   return `<?xml version="1.0" encoding="UTF-8"?>
-    <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+    <urlset xmlns="https://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xhtml="https://www.w3.org/1999/xhtml">
     ${getStaticPageEntries(staticPages)
       .map((p) => renderEntry(BASE_URL, p.url, p.priority, p.changefreq, p.lastmod))
       .join("")}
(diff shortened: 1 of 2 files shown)


frontend/pages/sitemap/[language_code_dot_xml].js
```diff
@@ -111,14 +111,54 @@ const getStaticPageEntries = (staticPages) => {
...
+    return `<xhtml:link
+    rel="alternate"
+    hreflang="${languageCode}"
+    href="${BASE_URL}${url}"/>
+    <xhtml:link
+    rel="alternate"
+    hreflang="de"
+    href="${BASE_URL}/de${url}"/>`;
+  }
+  const englishURL = BASE_URL + "/" + getEnglishURLFromGerman(url);
```
Comment (piperchester) [inline frontend/pages/sitemap/[language_code_dot_xml].js:135]: Since this is only used within the next condition, WDYT about moving it inside that `if`? 
```suggestion

  const englishURL = BASE_URL + "/" + getEnglishURLFromGerman(url);
```

frontend/pages/sitemap/[language_code_dot_xml].js
```diff
@@ -111,14 +111,54 @@ const getStaticPageEntries = (staticPages) => {
...
+    <xhtml:link
+    rel="alternate"
+    hreflang="en"
+    href="${englishURL}"/>`;
+  }
+};
+
+const getLanguageCode = (url) => {
+  if (url.startsWith("/de/")) return "de";
+  if (url.startsWith("/en/")) return "en";
```
Comment (piperchester) [inline frontend/pages/sitemap/[language_code_dot_xml].js:150]: I might be alone on this one (we should update our ESLint and Prettier rules to take a stance IMO), but I find that always including the braces is more readable and less bug prone. What do you think?
```suggestion
  if (url.startsWith("/de/")) {
    return "de";
  }
  
  if (url.startsWith("/en/")) { 
    return "en";
   }
```

piperchester (commented):
Nice work!

frontend/pages/sitemap/[language_code_dot_xml].js
Comment (JoshKornfeld) [reply frontend/pages/sitemap/[language_code_dot_xml].js:150]: I've always preferred reading if statements with curly braces as you suggested :)

frontend/pages/sitemap/[language_code_dot_xml].js
Comment (JoshKornfeld) [reply frontend/pages/sitemap/[language_code_dot_xml].js:136]: Sounds good to me.

ddhanesha:
@piperchester @positiveimpact  I am not familiar with sitemap google indexing. Do you think this is ready to be merged?

piperchester:
> @piperchester @positiveimpact I am not familiar with sitemap google indexing. Do you think this is ready to be merged?

Will take a look! Adding to my todo list this week to view.

piperchester (commented):
This is looking really great @JoshKornfeld - nice work figuring out the details. Do you know we'd verify how much it ends up improving the crawl coverage?

PR #1173, opened 2022-12-07
Add VSCode dev-container for one-click setup

phiresky:
## What and Why

VSCode Dev containers https://code.visualstudio.com/docs/devcontainers/containers allow setting up a project for development with a single click, regardless of the outside setup.

This PR adds a devcontainer configuration that includes PostGIS and Redis servers.

It also adds a script install_deps.sh and initial_dev_setup.sh that installs dependencies and creates env config files.


install_deps.sh
```diff
@@ -0,0 +1,13 @@
...
+
+cd "$(dirname "$0")"
+pushd frontend
+yarn install
+popd
+
+pushd backend
+python3 -m venv .venv
+source .venv/bin/activate
+pip install -r requirements.txt
```
piperchester, line 12: Nit: could we use?
```suggestion
make install
```

initial_dev_setup.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+set -euo pipefail # http://redsymbol.net/articles/unofficial-bash-strict-mode/
+
+# installs deps and creates the .env files with the minimal configuration for development
+# this script is automatically run in dev containers via postCreateCommand
```
piperchester, line 5:
```suggestion
# this script is automatically run in VS Code Dev Containers via `postCreateCommand`
```

README.md
```diff
@@ -2,7 +2,41 @@
...
+Make sure you have the [Remote Development Extension Pack](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) VSCode extension installed.
+
+1. Open the repo in VS Code
+1. Run the command "Reopen in dev container"
+
+   ![](doc/devcontainer-reopen.png)
+1. Wait for it to be done setting up the dev container and running the setup scripts. This will take 1-5 minutes the first time, 10 seconds after that. It should look something like this:
+   ![](doc/devcontainer-done.png)
+1. Start the frontend dev server using `cd frontend && yarn dev`
+1. Start the backend server using `cd backend && make start`. If you get a "Django is not installed" error run Ctrl+Shift+P "Select python interpreter" -> Recommended so it uses the venv.
```
piperchester, line 32:
```suggestion
1. Start the backend server using `cd backend && make start`. If you get a "Django is not installed" error run `Ctrl+Shift+P` (or `Cmd+Shift+P`) "Select Python interpreter" -> Recommended so it uses the Python venv.
```

piperchester (approved):
Amazing improvement! TIL about Dev Containers. Definitely think this is the direction we should head. Just a handful of small suggestions.

Issue #1144, opened 2022-11-16
Feature: Implement restrictions for cold-contacting users


positiveimpact:
We've had a few cases of users signing up and just messaging 20+ people in one hour with "Hello" or a similar message that provides no value and could be classified as spam. We need to make sure that messages on Climate Connect have value for the receiver and will therefore implement restrictions to prevent this type of behavior.
The following restrictions should be implemented:
- Require a first reach-out message to have at least 10 words. This should only apply to contacting somebody who you've never chatted with before
- For new accounts only allow to cold-contact 5 new people per day. Again this should only apply to contacting somebody who you've never chatted with before.
- Add a field is_user_muted to UserProfile which we can tick through the API to temporarily or permanently stop users from sending messages. Make sure the chat API respects this property and prevents you from sending messages if the value is True.

tobiasrehm:
What is the status of this? Isn't this implemented? @positiveimpact @ddhanesha

positiveimpact:
It's implemented. closed.

PR #1177, opened 2022-12-15
Add restricted_profile boolean field to restrict user access

ddhanesha:
## What and Why

Partially solved #1144. 

I have added a `restricted_profile` boolean flag to restrict access of users who are abusing the platform. I would like to note that it does not solve the problem completely but gives us a breathing room while we work on a permanent  solution.

# Test plan
* Use django shell or django admin, set `user_profile.restricted_profile` to `True`. 
* Go to http://localhost:3000/inbox and start chat with another user. 
* Write a message and try to send it. 
* You should see a message at the top that you account has been restricted. Here is a screenshot.
Screenshot 2022-12-15 at 8 21 17 AM


piperchester (approved):

piperchester:
Looks good!

PR #1178, opened 2022-12-15
First message of the chat should include 10 or more words to start the conversation. 

ddhanesha:
# What and Why

PR partially finishes #1144. It includes a logic that checks total number words of the first message sent by a sender to the climate connect member. This is to restrict users from just sending `Hi` messages to other members of the platform. Merely saying `Hi` does not provide enough information to the message receiver why the person has sent them a message. 

# Test plan
## Plan 1
* Start a new message with another member and type < 10 words in a message. 
* You should see a message at the top the screen. (Attached screenshot)
Screenshot 2022-12-15 at 2 16 18 PM

## Plan 2
* Start a new message with another member and type > 10 words
* Your message should be sent to the receiver.

## Plan 3
* On an existing message thread, type a message < 10 words and send.
* Your message should be sent to the receiver.


positiveimpact:
Oh, actually just realized that we need to send the error message in the correct language

PR #1182, opened 2023-01-13
fix: typo in German translation

MoritzLerch:
## What and Why
fixed a typo ☺️


ddhanesha (approved):
Thanks for the fix! Approved

PR #1186, opened 2023-01-18
Fix/allow exceptions to anti spam limit

positiveimpact:
## What and Why
Our community building managers had trouble with only being able to start 3 new chats per 3 hours. To still keep the anti-spam rules in place for the other users I created this PR giving them special permissions.

## Changes
- Added "mod" user group via permissions that we can use for staff or community mods in the future
- Added a custom permission for mods to create unlimited chats
- Made changes to there is a proper error message when you try to create too many chats
- Changed the limit from 3 to 4 chats per 3 hours to allow users a little bit more flexibility
- Changed `createMuiTheme` to `createTheme` due to frontend warning
- fixed a broken import in backend

I'd appreciate swift review because the community building managers are blocked in their work by not having the special permissions they get after this PR is merged.


ddhanesha:
Screenshot 2023-01-19 at 9 46 20 PM


Hey @positiveimpact I am getting this error while testing. Can't really test it. Can you fix it?

PR #1191, opened 2023-02-06
Prune requirements.txt with pipreqs

piperchester:
## What and Why

This runs [`pipreqs`](https://github.com/bndr/pipreqs) to prune our `requirements.txt` file to install only needed deps.


piperchester:
Closing as it looks like this tool is too good to be true. It pruned Celery and `cryptography` at least which we’re using (and have imports).

positiveimpact:
> Closing as it looks like this tool is too good to be true. It pruned Celery and `cryptography` at least which we’re using (and have imports).

Sucks that it turned out too good to be true :/

PR #1204, opened 2023-02-20
Only allow viewing and accepting project join requests with correct permissions

### LINKED ISSUE #1197: Every user is able to accept join requests for every Project
**Describe the bug**
User join requests can be accepted or declined by every logged in usewr and not only Project admins by simply adding "?show_join_requests=true" to the end of the project URL

**To Reproduce**
1. Go to a project with open requests. e.g.: https://climateconnect.earth/de/projects/kick-off-event
2. add ?show_join_requests=true to the end of the URL 
3. see open requests 
4. accept or decline

**Expected behavior**
Requests should only be answerd by Project admins

positiveimpact:
## What and Why
Fixes #1197
We had a bug where anybody could  accept or decline join requests from any project.

## Changes in this PR
- Fixed and used the permission class `ReadWriteSensibleProjectDataPermission` for viewing, accepting and declining project join requests
- Refactored `project_slug` --> `url_slug` in some places for standardization and because it was breaking some code
- Added a view and endpoint for checking whether you have requested to join a specific project. Before the frontend would just request all requesters and then check if the current user is in there. Had to make the new endpoint because viewing all requesters isn't possible for normal users anymore with the new permissions
- Added proper error and feedback messages when sending, accepting or declining a join request
- Put `getMembershipRequests` function into a lib file because it was used in many places
- ran `yarn format`, `yarn lint` and `black`

## Testing steps

1. Create a project with one account
2. Switch to a different account and navigate to the project you just created
4. Add `?show_join_requests=true` to the end of the project url. You should see a message telling you only administrators can view open join requests
5. Request to join the project 
6. Log in on the account who owns the project
7. Go to the project with `?show_join_requests=true` attached to the url
8. You should be able to accept and decline join requests as expected

### DIFF
--- a/backend/organization/views/project_views.py
+++ b/backend/organization/views/project_views.py
@@ -14,10 +14,8 @@ from climateconnect_api.models import (
     Role,
     Skill,
     UserProfile,
-    ContentShares,
 )
 from climateconnect_api.models.language import Language
-from climateconnect_api.permissions import UserPermission
 from climateconnect_api.utility.translation import (
     edit_translation,
     edit_translations,
@@ -63,7 +61,6 @@ from organization.models.translations import ProjectTranslation
 
 from organization.pagination import (
     MembersPagination,
-    ProjectCommentPagination,
     ProjectPostPagination,
     ProjectsPagination,
     ProjectsSitemapPagination,
@@ -73,7 +70,7 @@ from organization.permissions import (
     ChangeProjectCreatorPermission,
     ProjectMemberReadWritePermission,
     ProjectReadWritePermission,
-    ReadSensibleProjectDataPermission,
+    ReadWriteSensibleProjectDataPermission,
     ApproveDenyProjectMemberRequest,
 )
 from organization.serializers.content import PostSerializer, ProjectCommentSerializer
@@ -1037,6 +1034,7 @@ class ListProjectFollowersView(ListAPIView):
 
 
 class ListProjectRequestersView(ListAPIView):
+    permission_classes = [ReadWriteSensibleProjectDataPermission]
     """This is the endpoint view to return a list of users
     who have requested membership for a specific project, including their request IDs."""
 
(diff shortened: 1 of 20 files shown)


backend/organization/views/project_views.py
```diff
@@ -1066,6 +1064,33 @@ def get_queryset(self):
         return open_membership_requests
 
 
+class HasUserRequested(APIView):
```
ddhanesha, line 1067: Nit: Can you change the class name here? It is confusing to the reviewer. What does `HasUserRequested` means? It can mean multiple things. We should name this class based on its usage.

backend/organization/views/project_views.py
```diff
@@ -1066,6 +1064,33 @@ def get_queryset(self):
...
+    def get(self, request, url_slug):
+        if not request.user.is_authenticated:
```
ddhanesha, line 1069: If you define a permission class before the get request this would automatically be covered. https://www.django-rest-framework.org/api-guide/permissions/#isauthenticated

backend/organization/views/project_views.py
```diff
@@ -1235,15 +1260,15 @@ class ManageJoinProjectView(RetrieveUpdateDestroyAPIView):
     # Keep in mind that when we redefine class attributes at the View level,
     # we're overwriting the default permission and authentication classes
     # that're defined within settings.py.
-    authentication_classes = [TokenAuthentication]
```
ddhanesha, line 1238: Why did you remove authentication_class here?

frontend/src/components/project/ProjectOverview.tsx
```diff
@@ -146,37 +147,22 @@ export default function ProjectOverview({
...
-
-    const requestedMembers = resp.data.results;
+  const getHasRequestedToJoin = async () => {
+    try {
+      const resp = await apiRequest({
+        method: "get",
+        url: `/api/projects/${project.url_slug}/have_i_requested_to_join/`,
+        locale: locale,
+        token: token,
+      });
```
ddhanesha, line 158: Opinion: So the more API requests you make to the backend, it adds more time to load the page. I would like you to think, DO I REALLY NEED AN API TO CHECK FOR `have_i_requested_to_join`? Or can it be done with another fetch call? I really think we are at the point where we should question about our how we write our code. I know its a bug that we want to fix but if it can wait for one extra day, I would definitely spend time thinking about the approach here.

ddhanesha (changes requested):

frontend/src/components/project/ProjectOverview.tsx
positiveimpact: You're right. I have summarized liking, following and joining in one API call

positiveimpact:
> @positiveimpact can you add manual testing steps for this?

Added testing steps

positiveimpact:
@ddhanesha Can this be merged?

ddhanesha (approved):
Let's get this merged!

PR #1210, opened 2023-03-12
Pull privacy policy from webflow to make it updateable without code changes

positiveimpact:
## What and Why
Instead of hardcoding our privacy policy in the repo, we're pulling it from webflow so that we can change it easily without any dev work.


positiveimpact:
Merging this one directly since it just pulls content from webflow and uses the same code that we use in multiple other places for that.

PR #1213, opened 2023-03-19
Refactor: adds types to api operations

Dlyx24:
## Description

This is a follow up from the PR [Convert Frontend to TypeScript](https://github.com/climateconnect/climateconnect/pull/1126) by @phiresky.

It focus on refactoring the file `public/lib/apiOperations.ts` adding all the types missing and refactoring some redundant code.

  - It is now possible to specify the expected return type on the `apiRequest` method which provides with a better development experience with auto-complete and auto compile errors. 
(**Note:** It is still defaulting to any if no type if provided as there are many instances that this method is being used without types and I don't know the expected return type of the API.)

# Comments
Typescript is essential to prevent typing errors. 
While I was implementing these changes I noticed, for instance, that the `sendToLogin` method was using an untyped context and accessing properties of that context such as `.asPath`:

image

The issue is that in all the implementations of this method the context type that is passed is `GetServerSidePropsContext` which doesn't have the `.asPath` attribute. 
While it never actually reaches that code because the input `relativePath` would always be defined (I will explain further) this would cause a FE error that could easily be avoided by having the right types!

The reason why `relativePath` is always defined is because wherever you use this `sendToLogin` method you pass in the `context.resolvedUrl` which is indeed a property of the type `GetServerSidePropsContext` and it is always defined. 
For this reason there is no point in passing in arguments to this function such as `locale` or `relativePath` since you are using attributes from the context itself.

The refactors to the method `apiRequest` include:
 - **Removal of shouldThrowError input:** 
 This input was always defaulting to true and there were usages of `apiRequest` method that were passing it as true, making it redundant. Also the only thing it really did was throw an error that would be catched on the catch block or .catch method when using this `apiRequest`, which is already the case in case the returned promise gets rejected.
  - **Removal of try catch blocks in the return statement:** 
  There was no handling other than console.log the error which is already done on the try catch block whenever this `apiRequest` method is used, making it at worst console.logging twice the error. 
  Also whenever you used a `method` that required a payload you were awaiting for the response of the promise and returning that response but when there was no payload involved you always returned the promise. This is very confusing because sometimes you call this method with a payload expecting it to be a promise and sometimes you call it with a payload but expect it to be a promise fulfilled result. 
  For this reason the return type of this method will be a promise and the developer can choose how to handle it when it is fulfilled or rejected. All the cases where the expected return type was a result were handled either with a `try catch block` or with a `.catch` method.

## Before landing
Make sure all functionality (i.e: api calls) are working as expected on a staging/development environment.


ddhanesha:
Hello @Dlyx24, I apologize for the late response. Do you mind resolving merge conflicts? I will finish the review this week.

Dlyx24:
Hey @ddhanesha
Let me know if you are still interested in this PR. Otherwise feel free to close it!

ddhanesha (commented):
This works looks good to me. Let's get this merged once you fix the merge conflict @Dlyx24. cc @positiveimpact

Dlyx24:
@ddhanesha, Fixed the merging conflicts.

phiresky:
i like 👍 can we merge this?

positiveimpact:
@Dlyx24 Thanks for this very helpful PR! @phiresky Makes sense, I'll review and merge today!

positiveimpact (changes requested):
Thanks for this PR!
There seems to be a problem with the refactoring of http methods from lowercase to uppercase. It is actually breaking the `apiRequest` function in `frontend/public/lib/apiOperations.ts`. Before we were calling `axios[method]` which would either turn out to axios["get"] or axios["post"]. these functions exist and work as expected. However they do not exist when we use uppercase spelling which breaks the function.

Dlyx24:
@positiveimpact 
Apologies for the delay. I am not sure why changing the method from lower case to uppercase would trigger any type of error but since Axios Method type accepts both lower case and uppercase I reverted the changed I did to the HTTP method on all files.
Let me know if this fixes the issue!

frontend/public/lib/apiOperations.ts
```diff
@@ -1,115 +1,110 @@
...
+    headers["Accept-Language"] = acceptLanguageHeadersByLocale[locale];
   }
-}
+  const config: AxiosRequestConfig = tokenConfig(token, headers);
+
+  const requestPromise: Promise<AxiosResponse<T>> = payload
+    ? axios[method](process.env.API_URL + url, payload, config)
+    : axios[method](process.env.API_URL + url, config);
+
+  // Handle promise result wherever this method is called, either on a try catch block or with .then | .catch methods.
```
positiveimpact, line 36: Makes sense :+1: Thanks for improving this function!

positiveimpact (approved):
LGTM! Great work :+1: This is a really helpful PR that greatly improves our code quality. Sorry for the long wait, now we're finally merging :rocket:

Issue #1189, opened 2023-02-02
Create "event" project type and adjust "share project" accordingly


positiveimpact:
We are going to implement an event calendar for the ClimateHubs. To do this we will need additional infos about each project from the "event type". More information will follow.

FynhF:
This would be an awesome change also from an user's perspective! 

It would make it so much more intuitive to use browse a CH, since you could see at first glance when exactly something is happening. 
Right now I am seeing lots of stuff on the startpage or on landing page of a CH, but am having no clue when the events / projects are happening, if they are currently running or if they are already over. At least until I click the element, which makes me land on yet another page so that I have to either go back, or open multiple tabs if I want to find out about other events.

> Questions

    Remove project status completely?

What is the difference between event and project in the usecase of Climate Connect anyway?

positiveimpact:
> This would be an awesome change also from an user's perspective!
> 
> It would make it so much more intuitive to browse a CH, since you could see at first glance when exactly something is happening. Right now I am seeing lots of stuff on the startpage or landing page of a CH, but you cannot find out when the events / projects are happening, if they are currently running or if they are already over, until you open another page.
> 
> > Questions
> > Remove project status completely?
> 
> What is the difference between event and project in the context of Climate Connect anyway?

Hey! Sorry, I totally missed your comment. Thank you for your feedback! Currently "project" is just the umbrella term for any action/project/event/idea that users organize for the climate. Anything from the offer for balcony solar consulting to organizing a climate fair to just a vague idea falls into this currently. We want to change that and make it easier to find what you're looking for. This is why we decided to make events a seperate entity from projects and create an event calendar :)

FynhF:
Ah, I get it, thanks for the clarification and makes sense. :)

HaraldWalker:
Superseded by #2174 — the event calendar feature has shipped (PoC → production iteration → bugfixes). The new issue documents the outcome as built; this one is closed for historical context only. Remaining unchecked items here are addressed by the new design (separate page rather than a 4th Browse tab).

PR #1215, opened 2023-03-27
Feat/adapt share project first step

positiveimpact:
## What and Why
Implements the first step of #1189. It includes the following changes:
- Added new model `ProjectTypes`. The idea is that depending on what type you choose you will be displayed different options during the following steps of share project and your project card/page will look different
- Changed the first step of `Share project` to only include parent organization and name
- Added ProjectTypes data to create_test_data script
- Ran `yarn format` and `black .` which contributed to some unrelated files changing

Testing plan:
- Run migrations
- Run the `create_test_data_script` to populate your local database with project types. 
- Optional but recommended: add an image to each of your new project types in the database using the images from our drive folder (Web development -> testing images -> project_types) 
- Pick any project type and go to the next step


backend/climateconnect_main/utility/general.py
```diff
@@ -56,3 +56,12 @@ def get_allowed_hosts(ALLOWED_HOSTS_ENV):
             if not host in allowed_hosts:
                 allowed_hosts.append(host)
     return allowed_hosts
+
+
+def get_prop_in_language(obj, prop_name: str, language_code: str) -> str:
```
ddhanesha, line 61: Nit: Identify type for `obj` parameter

backend/organization/models/status.py
```diff
@@ -70,3 +70,94 @@ class Meta:
...
+
+    help_text_de = models.TextField(
+        help_text="Translation of help_text column",
+        verbose_name="Helptext DE translation",
+        max_length=512,
+        null=True,
+        blank=True,
+    )
+
+    icon = models.FileField(
```
ddhanesha, line 112: Is icon going to be an image or something else? Can we use imagefield here instead?

backend/organization/models/status.py
```diff
@@ -70,3 +70,94 @@ class Meta:
...
+        help_text="The icon representing the project type in the share project interface",
+        verbose_name="Icon",
+        null=True,
+        blank=True,
+        upload_to=project_type_image_path,
+    )
+
+    PROJECT_PROJECT_TYPE = 0
```
ddhanesha, line 120: Actually, I learned recently that it is easier to have name instead of number for type choices like this. Helps you analyse data more accurately. I would recommend changing this to 
```python
PROJECT_PROJECT_TYPE = 'project'
IDEA_PROJECT_TYPE = 'idea'
EVENT_PROJECT_TYPE = 'event'
```
Rest of its keep the same.

backend/organization/models/status.py
```diff
@@ -70,3 +70,94 @@ class Meta:
...
+
+
+class ProjectTypes(models.Model):
+    name = models.CharField(
+        help_text="Name of the project type",
+        verbose_name="Name",
+        max_length=512,
+        null=True,
+        blank=True,
+    )
```
ddhanesha, line 86: What does this represent here? Can you elaborate?

backend/organization/serializers/status.py
```diff
@@ -34,3 +38,29 @@ def get_status_type(self, obj):
...
+    def get_name(self, obj):
+        return get_project_type_name(obj, get_language())
+
+    def get_original_name(self, obj):
+        return obj.name
+
+    def get_type_id(self, obj):
+        project_type_id = list(
+            filter(lambda x: x[0] == obj.type_id, ProjectTypes.POSSIBLE_PROJECT_TYPES)
+        )[0][1]
```
ddhanesha, line 62: Whenever there are indexes like these, it is difficult for reviewer to understand what it is doing. A better practice is to define these indexes as a variable. For example in this case the `0` index in `lambda x: x[0]` would be something like this. 
```python
        TYPE_ID_INDEX = 0
        PROJECT_TYPE_INDEX = 1
        project_type_id = list(
            filter(lambda x: x[TYPE_ID_INDEX] == obj.type_id, ProjectTypes.POSSIBLE_PROJECT_TYPES)
        )[TYPE_ID_INDEX][PROJECT_TYPE_INDEX]
```
This way it is clear for other to make sense of what is going on.

backend/organization/views/project_views.py
```diff
@@ -846,6 +849,14 @@ def get_queryset(self):
         return ProjectStatus.objects.all()
 
 
+class ListProjectTypeOptions(ListAPIView):
+    permission_classes = [AllowAny]
```
ddhanesha, line 853: Why is the permission here `AllowAny`? Are we allowing another filter to the project page?

frontend/src/components/shareProject/ProjectTypeSelector.tsx
```diff
@@ -0,0 +1,108 @@
...
+    width: "100%",
+  },
+  radioButton: {
+    position: "absolute",
+    right: 0,
+    top: 0,
+  },
+}));
+
+export default function ProjectTypeSelector({ className, value, types, onChange }) {
```
ddhanesha, line 63: Nit: Can you move the default function at the end of the file? Its just easier to read and follow.

ddhanesha (changes requested):

backend/climateconnect_main/utility/general.py
positiveimpact: I couldn't find a good way to type this. These are always instances of a django model. Would you recommend to just use `obj: obj` or is there a way to do something like `obj: ModelInstance`?

backend/organization/models/status.py
positiveimpact: I used file field to allow for svg icons

backend/organization/models/status.py
positiveimpact: Are you referencing the null=True and blank=True? I just took the standard settings but I agree that we should make those mandatory.

backend/organization/views/project_views.py
positiveimpact: I chose AllowAny because this is publically available data. In the future we will introduce a filter for this which will make AllowAny necessary. Would your preference be to change it to isAuthenticated now and return to AllowAny when we need it?

backend/organization/models/status.py
positiveimpact: I tried this but this broke my get_id function so I reverted it.

PR #1224, opened 2023-04-15
Feat/1221 improve navigation climate hub page

MarcoRehmer:
**WIP**

## What and Why

TBD


positiveimpact (approved):
Because of the very short deadline I just added all the additional design elements, some of them not in the prettiest/most generic way. 
It might be worth to revisit this when we talk next week but for now I'd merge it like this.

PR #1225, opened 2023-04-19
Fixing build failure

ddhanesha:
## What and Why

PR includes a draft of fixes we would need to make to material ui upgrade work with our build.


positiveimpact:
LGTM, thanks for making this change! And sorry for breaking the pipeline with the mui PR. I made one small change before mering: I refactored Grid back to the newer version (reasoning explained in mui PR) and removed the `item` property from Grid items to make the build work.

Issue #1229, opened 2023-04-22
Bug: @ mentions in comments: dropdown hidden/creates scroll


tobiasrehm:
**Describe the bug**
When @ ing another user the menu in which to select a user is hidden and creates a scroll inside the comment input field

**To Reproduce**
1. Go to an projects comment section 
2. Type "@[some user name]

**Expected behavior**
The selection field should not stay inside the comment text input but overlap to the "outside" and not create a scroll

**Screenshots**
![image](https://user-images.githubusercontent.com/64414429/233600948-f18fedf7-b04e-4b0c-b24a-c8b23ae004d2.png)

PR #1236, opened 2023-05-09
#1229 fix hidden mention suggestions

MarcoRehmer:
## What and Why

- set overflow to "visible" for mentionsInputContainer

## Additional

- avoid nullish project input for project preview state


ddhanesha (approved):
Good catch! Let's get this merged.

PR #1240, opened 2023-05-11
Fix/1238 adjust organization card height

MarcoRehmer:
## What and Why

- increase the height of organization cards to 390px fix

## Additional

- add "husky" pre-commit hooks and "lint-staged" to only format staged files with prettier (only affects js, ts, jsx, tsx files. Backend files are not be affected by this)
- add installation steps to README.md


positiveimpact (changes requested):
For the the husky pre-commit hook doesn't work yet. 
I ran the ```sh
yarn prepare```command in `climateconnect/frontend`. This creates a `.husky/pre-commit` file which looks like this: 
```#!/usr/bin/env sh
. "$(dirname -- "$0")/_/husky.sh"
cd frontend && npm run pre-commit```
When making a commit it will fail because this file is in the `frontend` folder but tries to `cd` to `frontend` which is not possible because it's already there so I get this error: `frontend/.husky/pre-commit: 5: cd: can't cd to frontend`

MarcoRehmer:
ah yes good catch, also it should not run "npm", instead yarn should be used. I will fix that

positiveimpact:

PR #1241, opened 2023-05-11
fix: make dev container postInstall setup work

phiresky:
enables the venv before attempting to run make migrate.


positiveimpact (approved):
LGTM! Approved.

PR #1242, opened 2023-05-11
Feat/python lint

phiresky:
Changes the python lint to enforce formatting with black and passing ruff.rs lints.

#1243 contains the actual lint fixes, merge that one into this first


positiveimpact (approved):
LGTM! Thanks for this change!

PR #1244, opened 2023-05-13
Save type on step 1 of create organization

### LINKED ISSUE #1209: Create Organisation: Org type selecte in step one is not applied in step 2
**Describe the bug**
When creating an organisation, the user needs to select an organisation type in step one. This type is not carried over to step 2 and needs to be selected again. Step 2 is also missing clear indication of which fields are required. 

**To Reproduce**
1. Got to Profile page and create organisation 
2. Input name, location and org type 
3. click next 
4. see error -> Org type not applied

**Expected behavior**
Org type input in step one should be carried over to step two. In addition, there should be a possibility to go back to step one, which we currently do not have.

**Screenshots**

https://user-images.githubusercontent.com/64414429/221816134-92e7b7a7-ef48-4ea7-baef-013da3595c1d.mp4

positiveimpact:
Fixes #1209. The types you selected in step 1 of "create organization" did not carry over to step 2. You had to select the organization types again. This is now fixed. Additionally I fixed the broken loading animation of the create organization page.


frontend/src/components/account/EditAccountPage.tsx
```diff
@@ -800,6 +800,7 @@ export default function EditAccountPage({
                     infoMetadata
                   ).map((typeObject) => (
                     <Chip
+                      color="secondary"
```
MarcoRehmer, line 803: here you have a double property

PR #1246, opened 2023-05-19
Add python type checking via mypy

phiresky:
## What and Why

This PR adds mypy type checking to the code base and enforces type-correctness via the backend lint.
Based on #1245 to make it easier


phiresky:
I've fixed the ~200 errors that appear with the default settings of mypy.

Strict mode is off in mypy. Enabling strict mode yields 1000 more errors, mostly due to missing type annotations.

I've tried a few tools to auto-add correct type annotations:

* [pyre](https://github.com/facebook/pyre-check) (static inference): works out of the box. adds many type annotations but some fail mypy, needs manual fixes
* [pytype](https://github.com/google/pytype) (static inference): worked with some tweaks. needs manual fixes probably
* [monkeytype](https://github.com/Instagram/MonkeyType) (run-time analysis): Couldn't get it to work, it only wanted to annotate `climateconnect_main.utility.general` nothing else. Possibly some interaction with the django dev server.

backend/chat_messages/consumer.py
```diff
@@ -1,6 +1,6 @@
 import json
 from django.utils import timezone
-from channels.generic.websocket import AsyncWebsocketConsumer
+from channels.generic.websocket import AsyncWebsocketConsumer  # type: ignore
```
positiveimpact, line 3: What do these mean?

backend/climate_match/serializers/question_answer.py
```diff
@@ -34,34 +36,39 @@ class Meta:
...
+                .text  # type: ignore
             )
         return obj.text
 
     def get_answer_type(self, obj: Question) -> str:
+        assert obj.answer_type
         return obj.answer_type.model
 
-    def get_answers(self, obj: Question) -> List:
+    def get_answers(self, obj: Question) -> Any:
```
positiveimpact, line 54: Won't this always be a list? :thinking:

backend/climateconnect_api/management/commands/create_climatematch_data.py
```diff
@@ -136,8 +136,10 @@ def handle(self, *args: Any, **options: Any) -> None:
                 )
                 if "number_of_choices" in question:
                     question_in_db.number_of_choices = question["number_of_choices"]
-                    if "minimum_choices" in question:
-                        question_in_db.minimum_choices = question["minimum_choices"]
+                    if "minimum_choices_required" in question:
+                        question_in_db.minimum_choices_required = question[
```
positiveimpact, line 140: Good catch :+1:

backend/ideas/utility/idea.py
```diff
@@ -125,30 +125,22 @@ def idea_translations(
...
-    ):
-        return (
-            idea.translate_idea.filter(language__language_code=language_code)
-            .first()
-            .name_translation
-        )
+def get_translation(idea: Idea, language_code: str) -> Optional[IdeaTranslation]:
+    if idea.language and language_code != idea.language.language_code:
+        return idea.translate_idea.filter(language__language_code=language_code).first()
+    return None
```
positiveimpact, line 131: We should definitely make these generic all over the code base but for this PR it's okay to do it like this.

backend/organization/management/commands/organization_translation.py
```diff
@@ -36,7 +36,7 @@ def translate_organization(organization: Organization) -> None:
             if "school" in texts:
                 org_translation.school_translation = texts["school"]
             if "about" in texts:
-                org_translation.about = texts["about"]
+                org_translation.about_translation = texts["about"]
```
positiveimpact, line 39: Good catch!

backend/organization/management/commands/project_translation.py
```diff
@@ -7,7 +7,7 @@
 
 
 def translate_project(project: Project) -> None:
-    data = {
+    data: dict[str, Any] = {
```
positiveimpact, line 10: Does this mean the first element should be a string and then there can be any number of following elements of any type? Or does it simply mean all elements in the dict should either be of type "str" or "Any"?

backend/climateconnect_main/utility/general.py
```diff
@@ -10,7 +11,9 @@
 
 
 # source: https://dev.to/ageumatheus/creating-image-from-dataurl-base64-with-pyhton-django-454g
-def get_image_from_data_url(data_url, resize=False, base_width=500):
+def get_image_from_data_url(
+    data_url, resize=False, base_width=500
+) -> tuple[ContentFile, tuple[str, Any]]:
```
positiveimpact, line 16: Here I'm getting  this error when trying to run the backend: 
```
File "/home/chris/climateconnect_env/climateconnect/backend/climateconnect_main/utility/general.py", line 16, in <module>
    ) -> tuple[ContentFile, tuple[str, Any]]:
TypeError: 'type' object is not subscriptable
```

positiveimpact: I'm getting the following hint from vscode to fix this: `Subscript for class "list" will generate runtime exception; enclose type annotation in quotes`

positiveimpact: Same thing in organization/utility/project.py at this position: `def get_project_helpful_connections(
    project: Project, language_code: str
) -> Optional[list[str]]:`

positiveimpact:
```suggestion
) ->"tuple[ContentFile, tuple[str, Any]]":
```

backend/organization/utility/project.py
```diff
@@ -66,61 +67,44 @@ def create_new_project(data: Dict, source_language: Language) -> Project:
...
-        return project.translation_project.get(
-            language__language_code=language_code
-        ).helpful_connections_translation
+        ).first()
+    return None
+
+
+def get_project_helpful_connections(
+    project: Project, language_code: str
+) -> Optional[list[str]]:
```
positiveimpact, line 82:
```suggestion
) -> Optional["list[str]"]:
```
To fix the error I mentioned above.

positiveimpact (commented):
LGTM except one small issue I mentioned in the comments! Once this is fixed this can be merged!

positiveimpact (commented):
LGTM except one small issue I mentioned in the comments! Once this is fixed this can be merged!

backend/organization/management/commands/project_translation.py
phiresky: it means that all keys in the dict are `str` and all values are `any`. That's what most dicts are but python allows many things as dict keys apart from strs

backend/chat_messages/consumer.py
phiresky: it tells mypy that it's ok that the given library does not have type hints (and all things pulled from the library will have type `any` instead of erroring)

backend/climate_match/serializers/question_answer.py
phiresky: probably but it can return Serializer.data which would need some other change to make mypy understand returns a list in this case

backend/climateconnect_main/utility/general.py
positiveimpact: I'm not sure how much effort and risk a python minor version upgrade is. Would you recommend to go for the upgrade or is this a bigger project? What are your thoughts on this @ddhanesha ?

PR #1284, opened 2024-04-01
Added code for adding script that helps with animation

ddhanesha:


positiveimpact:
We could try this option you just found: https://docs.developers.webflow.com/data/docs/devlink-documentation-and-usage-guide

PR #1288, opened 2024-05-16
Pull blog posts from webflow and show them in sitemap

positiveimpact:
## What and Why

As @tobiasrehm pointed out, so far our blog posts were not shown in our sitemap and could therefore not be found on Google. This is because they have dynamic urls and are pulled from webflow. I fixed this, they can now be found in our sitemap and will therefore be visible on Google

## Changes
- Added a function to pull all blogposts in a language from the webflow API
- introduced a new env variable `WEBFLOW_TOKEN` to store our webflow API token

# Danger! add WEBFLOW_TOKEN env variable to server before pushing this update!


frontend/public/lib/webflowOperations.ts
```diff
@@ -0,0 +1,31 @@
+import axios from "axios";
+import { CcLocale } from "../../src/types";
+
+export async function getAllBlogPosts(locale: CcLocale) {
+  const api_token = process.env.WEBFLOW_API_TOKEN
+  const WEBFLOW_LANGUAGES = {
+    de: "e8f987c48323e8e2485564e99263ac0e",
+    en: "08ea600611a6f0978326919b604c81c1"
```
ddhanesha, line 8: Are these random or some keys?

frontend/pages/sitemap/[language_code_dot_xml].tsx
```diff
@@ -155,27 +160,38 @@ const getEntries = async (entryTypePlural, locale, language_code_for_url) => {
...
+
+  if(entryTypePlural === "post") {
+    const blogPosts = await getAllBlogPosts(language_code_for_url ? language_code_for_url : locale);
+    return parseEntries(entryTypePlural, blogPosts, language_code_for_url);
+  }
+
+  try {
+    const resp = await apiRequest({
+      method: "get",
+      url: "/api/sitemap/" + entryTypePlural + "/?page_size=1000",
```
Comment (ddhanesha) [inline frontend/pages/sitemap/[language_code_dot_xml].tsx:173]: Don't we need the full URL here? Otherwise it would use the default URL right?

ddhanesha (approved):

Issue #666, opened 2021-07-14
If Org has no project add this Vector grafic and text


tobiasrehm:

tobiasrehm:
![image](https://user-images.githubusercontent.com/64414429/125600172-467e55df-ee5c-4350-9fbb-6a7396fe50ab.png)

xd link:  https://xd.adobe.com/view/43e90872-89e3-4bf0-8542-0ed6330bcb85-16ba/

tobiasrehm:
bonus: Have multiple images and texts that are shown randomly. Ideally we could add new ones through the backend

positiveimpact:
Voting to keep this.
@ddhanesha @tobiasrehm Could you react to this comment with :+1: (Keep open) or :-1: (Close)?

tobiasrehm:
not on roadmap

PR #1290, opened 2024-05-20
Adding image for no Project/s and Member/s yet in an Org

AnneMerlyn:
## What and Why

- Added the image without background both for the project and member section.
- Remove the old text and added the wrapper to center the image.

This will solve this following issue #666


positiveimpact:
Thanks for submitting this PR! There is a few things we still need to change, some of them not code-related:
- The vector graphic should be displayed only on the logged in user's profile page, not on the organization page. The PR shows the correct image for that but mistakenly describes that it should be on the organization page. That doesn't make sense because orgs are by definition not lonely sailors but multiple people
- The vector graphic should not include text as the text should adapt to the language that the user has chosen (Please fix this @tobiasrehm )

tobiasrehm:
Gruppe 7794

Here is just the sailboat without text as an SVG. 

maybe we could even implement it in a way that we choose from mulitple image an text combos randomly when the page is loading. Similar to how Slack dose it when you have no unread messages:
image

Issue #1279, opened 2024-03-05
Bug: Date indication on events overlaps with mobile menus


positiveimpact:
**Describe the bug**
Both on `/browse` and on location hub pages (e.g. `/hubs/erlangen`) the date and time indicator on projects of the type "event" will overlap with the bottom menu on mobile.

**To Reproduce**
1. Make sure you have added at least one location hub manually through `localhost:8000/admin/`. For this you will need at least one location in your database. Locations are automatically retrieved from OpenStreetmaps and added to our database when you search for them through the filters on the browse page or create a project with a location manually through the GUI.
2. Make sure you have ran the script to create test data and manually created at least one project of the project type "event" which is part of the Location hub you created
3. Switch to mobile view in your browse
4. Scroll down to see the event time and date indication overlap with the mobile menu on the bottom of the screen like in the screenshot

**Expected behavior**
The bottom menu should be displayed on top of the date and time indication. Please make sure that there aren't any unintended consequences such as the bottom menu suddenly being displayed over popup windows.

**Screenshots**
![Screenshot from 2024-03-05 11-03-48](https://github.com/climateconnect/climateconnect/assets/10531079/ee135fb7-0acb-4284-a952-83d1dadbc32e)

PR #1291, opened 2024-05-21
Fixing EventDateIndicator Tooltip that overlaps the MobileBottomMenu

AnneMerlyn:
#### These updates will fix this bug #1279 
- Updated the z-index for the `EventDateIndicator`,  lowering to `10`
- Updated also the z-index for the `MobileBottomMenu`, bumping it up to `20`

#### Additional fix : Found an issue when scrolling down the content: where the Footer's height is still higher than the mobile bottom menu and can still be seen 
- Added also height for the `Footer` in relative position, with `49px`
  
  #### See Screenshot Below of the Issue: 
  
    image

### DIFF
--- a/frontend/src/components/browse/MobileBottomMenu.tsx
+++ b/frontend/src/components/browse/MobileBottomMenu.tsx
@@ -14,7 +14,7 @@ const useStyles = makeStyles(() => ({
     bottom: 0,
     left: 0,
     right: 0,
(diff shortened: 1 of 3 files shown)


ddhanesha (approved):
Thanks for the submission. Let's get this merged

positiveimpact (approved):
LGTM! Thanks for making this change, merged!

PR #1331, opened 2024-10-05
[security] Fix Django `DEBUG` setting

hartwork:
The current code makes `DEBUG=False` activate debug because `bool("False")` is `True` in Python. In general, `DEBUG` should be disabled in production and made to gravitate towards `False` (rather than current `True`) for security.

### DIFF
--- a/backend/climateconnect_main/settings.py
+++ b/backend/climateconnect_main/settings.py
@@ -38,7 +38,7 @@ BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 SECRET_KEY = env("SECRET_KEY")
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = env("DEBUG", False)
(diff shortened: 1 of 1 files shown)


hartwork:
@positiveimpact this is urgent, surfing to https://api.climateconnect.earth/ shows that debug mode is still active, and all that an attacker needs is one single error 500 crash for any of the endpoints (which is rather easy, it could point to a specific view that's easy to crash) and they get read access to Django setting `CELERY_BROKER_URL` including the username and password through the crash report page due to debug mode, and then with those credentials they have write access to Redis, control both the cache content and the celery queue and can likely execute arbitrary code and more on your server. That's about as bad as it gets in my book. All it takes is this pull request here and a fresh soon deployment. Please act soon, thanks for your attention!

positiveimpact (approved):
LGTM. Thank you for catching this major issue and creating this PR! 
This is a huge help and prevents potential future damage! 
Thank you also for creating the other PRs, I will review them shortly!

PR #1334, opened 2024-10-09
Whitelisted params

Idontker:
## What and Why
We shall not trust user input (like e.g.) when constructing urls to connect to our backend.
To address this issue a whitelist approach was used: 
- Added a whitelist for characters within [getDataOperations.ts](https://github.com/climateconnect/climateconnect/compare/whitelisted-params?expand=1#diff-d3a5194184ff628bc925f2f4192b79a3bd7bc323f82644ade95cbd4c7a31b738) 
- added `TYPES_BY_TAB_VALUE.indexOf(newHash) != -1` to check if the # read by the BrowseComponent.ts is a valid value.


Idontker:
This PR is not ready yet contains a bug: Some filter values might include other important symbols as e.g. spaces.

Idontker:
Changes were not complete and did not solve an issue
=> Closed.

PR #1345, opened 2024-10-10
Refactoring BrowseComponent, and its parents relation to Filters

Idontker:
## Beaware
This is still a draft. It appears to change the fundamental premise of responsibilities of the filters state and its state management. Therefore, I expect this PR change a lot of lines without the ability of splitting it into smaller chunks.
I will try to add comments to the code and PR to explain changes.
  

## What and Why
There are lot of dependencies and shared responsibilities  regarding the filtering feature within the `BrowseComponent.ts`, its parents and its children. This leads to a higher complexity and duplicated code. This PR would like to present an alternative premise and, based on this, simplify state management and bundle it into fewer components.

### Previously
- The pages `browse.tsx` and `[hubUrls].tsx` were the owner of the `filters` state, as they need to load the initial filters and all possible filters options (which filters are allowed).
- The `filters` state was updated in sync with URL by `BrowseComponent.tsx`.
- props regarding the update of state and onSubmit hooks were implemented, extended, used or passed to ones children throughout up to 5 component levels (e.g.  `browse.tsx` -> `BrowseComponent.tsx` -> `FitlerContent`  -> `Filters`). 

### New Premises 
- The page components only provide the initial values and filter options/definitions. These values have to be passed through to the `FilterSection` - the new owner of the `filters` state.
- `BrowseComponent.tsx` need to extract filters from the URL on load (UX feature). Therefore,  `BrowseComponent` can use the URL as the *only source* to derive the current state of the filters. This leads to:
  - `BrowseComponent` does not  actively change the `filters` state
  -  `BrowseComponent` needs to listen to URL changes and reload data
  -  `BrowseComponent` is responsible for reloading and caching query results locally.
  -  `FilterSection` owns the `filters` state and it and its children are the only ones managing/changing it. 
  - `FilterSection` needs to keep the URL in sync. 
  - `FilterSection` needs to notify `BrowseComponent` with a custom event detached from `window.location` as (1) react-router is not used (2) react does not provide a hook for the location history (3) there is no native event triggered by updating the history without browser level navigation.


frontend/pages/hubs/[hubUrl].tsx
```diff
@@ -3,9 +3,8 @@ import makeStyles from "@mui/styles/makeStyles";
 import parseHtml from "html-react-parser";
 import Head from "next/head";
 import React, { useContext, useEffect, useRef, useState } from "react";
-import Cookies from "universal-cookie";
```
Comment (Idontker) [inline frontend/pages/hubs/[hubUrl].tsx:6]: unused after changes

frontend/pages/hubs/[hubUrl].tsx
```diff
@@ -335,20 +277,13 @@ export default function Hub({
           {!isLocationHub && <BrowseExplainer />}
           <BrowseContext.Provider value={contextValues}>
             <BrowseContent
```
Comment (Idontker) [inline frontend/pages/hubs/[hubUrl].tsx:279]: removed filters related hooks but added the initalFilters and filterChoices as these might need values provided by SSR

Issue #1360, opened 2024-10-25
Bug: Loading filters from URL fails on full (re)load 


Idontker:
**Describe the bug**
Filters are saved within the URL but on first load / page reload they will be removed from the url and not applied.

**To Reproduce**
1. Go to https://climateconnect.earth/browse?&search=Energy&
2. Search text will be empty and URL will change to https://climateconnect.earth/browse

**Expected behavior**
Filters should correctly be loaded from the URL

PR #1361, opened 2024-10-25
fixed the bug (wrong object was used to get the filters from the url)

Idontker:
## What and Why
fixing #1360 
Wrong object was used to get the filters from the URL (oneliner)


positiveimpact (approved):
LGTM

PR #1365, opened 2024-11-07
improve the supporters subtitle

fateme-ramezanpour:
## What and Why
 In this PR, the subtitle style for the supporter’s section is updated.

## Changes
Added a CSS style to set `font-weight: normal `for the subtitle in the supporter's section.


positiveimpact (approved):
LGTM, thanks for the quick change! :)

Issue #1368, opened 2024-11-18
[Custom Hub Feature] Allow setting custom primary and secondary colors for specific hubs


positiveimpact:
## Why and What
We plan to customize the hub theme for certain external hubs. This customization can be handled in one of two ways:

**1.  Frontend-Only Approach:**

- Define the theme colors directly in the `Theme.ts` file.
- Use Next.js's useRouter hook to determine the current route and dynamically update the theme based on the route.

**2. Integration of Backend and Frontend:**

- The backend can provide an API for administrators to manage theme settings.

- The frontend can use this API to fetch the theme data and apply it dynamically via the `ThemeProvider`.

_These steps are required to achieve this feature._

- [x] Implement `hubTheme` and `hubThemeColor` tables in the backend:

       The `HubTheme` table will store foreign key relationships to the `HubThemeColor` and `Hub` tables.
       The `HubThemeColor` table will store color information related to the theme.

- [x] making Api to get Theme data

- [x] Integrated theme API data on the Front-end and dynamically generated the theme object

fateme-ramezanpour:
[Design link](https://xd.adobe.com/view/1c7e3ef7-9ebf-4b92-9ac1-1b5bae1fddff-492d/)

PR #1379, opened 2024-11-25
make HubThemeColor && HubTheme models and view

fateme-ramezanpour:
## What and Why
This PR addresses feature [1368](https://github.com/climateconnect/climateconnect/issues/1368)

## Solution
**Integration of Backend and Frontend:**

- The backend can provide an API for administrators to manage theme settings.

- The frontend can use this API to fetch the theme data and apply it dynamically via the ThemeProvider.

_These steps are required to achieve this feature._
- [x] Implement `hubTheme` and `hubThemeColor` tables in the backend:

       The `HubTheme` table will store foreign key relationships to the `HubThemeColor` and `Hub` tables.
       The `HubThemeColor` table will store color information related to the theme.

- [x] making Api to get Theme data

- [x] Integrated theme API data on the Front-end and dynamically generated the theme object

Here’s what it currently looks like:
Screenshot 2024-12-04 at 11 53 13 AM

Screenshot 2024-12-04 at 11 50 18 AM

Project page currently looks like:

Screenshot 2024-12-04 at 11 45 18 AM


backend/hubs/models/hub.py
```diff
@@ -415,27 +415,37 @@ class HubTheme(models.Model):
...
+        related_name="secondary",
         help_text="secondary_color",
         verbose_name="secondary_color",
         on_delete=models.CASCADE,
         max_length=1024,
         null=True,
         blank=True,
     )
-    
+    text = models.ForeignKey(
```
fateme-ramezanpour, line 438: Added a text field to set the theme's font color.

frontend/pages/hubs/[hubUrl].tsx
```diff
@@ -288,101 +294,103 @@ export default function Hub({
...
-              navigationRequested={navRequested}
-            />
-          )}
-          {!isLocationHub && (
-            <HubHeaderImage
-              image={getImageUrl(image)}
-              source={image_attribution}
-              onClose={closeHubHeaderImage}
-              isLocationHub={isLocationHub}
+        <ThemeProvider theme={hubThemeData ? transformThemeData(hubThemeData) : theme}>
```
Comment (fateme-ramezanpour) [inline frontend/pages/hubs/[hubUrl].tsx:297]: Used ThemeProvider to customize the hub page theme.

frontend/pages/hubs/[hubUrl].tsx
```diff
@@ -32,6 +32,8 @@ import AddIcon from "@mui/icons-material/Add";
 import { Theme } from "@mui/material/styles";
 import theme from "../../src/themes/theme";
 import BrowseContext from "../../src/components/context/BrowseContext";
+import { ThemeProvider } from "@mui/material/styles";
+import { transformThemeData } from "../../src/themes/transformThemeData";
```
Comment (fateme-ramezanpour) [inline frontend/pages/hubs/[hubUrl].tsx:36]: Implemented a function to transform theme data received from the API into a structured theme object for use in the frontend.

frontend/pages/hubs/[hubUrl].tsx
```diff
@@ -460,3 +468,18 @@ const getHubSupportersData = async (url_slug, locale) => {
     return null;
   }
 };
+
+const getHubTheme = async (url_slug) => {
```
Comment (fateme-ramezanpour) [inline frontend/pages/hubs/[hubUrl].tsx:472]: fetch Theme data

frontend/src/components/hub/HubSupporters.tsx
```diff
@@ -27,7 +19,7 @@ type HubSupporter = {
 
 const useStyles = makeStyles((theme) => ({
   root: {
-    backgroundColor: "#207178",
+    backgroundColor: theme.palette.primary.main,
```
fateme-ramezanpour, line 22: Made the background color dynamic

frontend/src/themes/transformThemeData.ts
```diff
@@ -0,0 +1,29 @@
+import defaultTheme from "./theme";
+
+export const transformThemeData = (data) => {
```
fateme-ramezanpour, line 3: transform theme data received from the API into a structured theme object

backend/hubs/models/hub.py
```diff
@@ -382,20 +382,35 @@ class HubThemeColor(models.Model):
...
+        help_text="extraLight color",
+        verbose_name="extraLight color",
         max_length=200,
         null=True,
         blank=True,
     )
-    extraLight = models.CharField(
-        help_text="extraLight color",
-        verbose_name="extraLight color",
+    contrastText = models.CharField(
```
fateme-ramezanpour, line 392: To store the text color, a `contrastText` field was added to the model.

backend/hubs/models/hub.py
```diff
@@ -382,20 +382,35 @@ class HubThemeColor(models.Model):
...
+        help_text="contrastText color",
+        verbose_name="contrastText color",
         max_length=200,
         null=True,
         blank=True,
     )
+    background_default = models.CharField(
```
fateme-ramezanpour, line 399: We need a separate color for the Navbar. 
The `background_default` field allows us to store this color in the database.
We also plan to fill the background object in the Theme like this:
`background: { default: "", paper: "" }`
The `background_paper` field will define the background color for elements like the project preview.

backend/hubs/models/hub.py
```diff
@@ -435,17 +450,6 @@ class HubTheme(models.Model):
         null=True,
         blank=True,
     )
-    text = models.ForeignKey(
```
fateme-ramezanpour: Removed the `text` field and replaced it with the `contrastText` field.

frontend/pages/projects/[projectId].tsx
```diff
@@ -242,53 +249,55 @@ export default function ProjectPage({
...
-                requestedToJoinProject={requestedToJoinProject}
-                handleJoinRequest={handleJoinRequest}
-                hubSupporters={hubSupporters}
-                hubPage={hubPage}
-              />
-            </div>
-            <div className={classes.secondaryContent}>
-              {!smallScreenSize && (
-                <ProjectSideBar
+        <ThemeProvider theme={hubThemeData ? transformThemeData(hubThemeData) : theme}>
```
Comment (fateme-ramezanpour) [inline frontend/pages/projects/[projectId].tsx:251]: We need an API call to fetch theme data in `[projectId].tsx`, as this page also includes custom theme settings. However, the API call should be conditional and only trigger when we are on specific hubs (after deciding which hub types require theme customization).

frontend/pages/projects/[projectId].tsx
```diff
@@ -460,3 +469,18 @@ function parseProjectMembers(projectMembers) {
     };
   });
 }
+
+const getHubTheme = async (url_slug) => {
```
Comment (fateme-ramezanpour) [inline frontend/pages/projects/[projectId].tsx:472]: fetch Theme data

frontend/src/components/hub/HubTabsNavigation.tsx
```diff
@@ -19,14 +19,16 @@ const useStyles = makeStyles((theme) => ({
   },
   tab: {
     textTransform: "none",
-    color: "white",
+    color: theme.palette.primary.contrastText,
```
fateme-ramezanpour, line 22: Make the color dynamic.

frontend/src/themes/theme.ts
```diff
@@ -21,13 +21,14 @@ const coreTheme = createTheme({
     primary: {
       main: "#207178",
       light: "#66BCB5",
-      lightHover: "#7dd1ca",
```
fateme-ramezanpour, line 24: We no longer use `lightHover`, so I removed this line.

frontend/src/themes/transformThemeData.ts
Idontker: Unsure about the name as  I missunderstood functions "task" when only reading the name :) But I do not have a better name. I'll just add some comment litterly describing it with your comment :D 
// transform theme data received from the API into a structured theme object

Additionally, maybe one could combine the `getHubTheme` with this function, as they are always called together

Idontker (commented):
In general, a great and elegant solution to the imo difficulty task.
- Future frontent development can ignore the theme most of the time, as it is done (more or less) automatically. 
- Adding and switching colors is esay as well by changing or inserting new colors in the database. 
- Feature is implemented as an opt in feature. If the backend request fails or colors are missing, the frontend will fallback to our basic palette 👍 
I'll add some minor things and approve it afterwards :) 🚀

Idontker (changes requested):
@fateme-ramezanpour After taking a second look: Do you mind hopping into another call? I spotted some smaller issues and would like solutions :)

backend/hubs/migrations/0021_auto_20241120_0955.py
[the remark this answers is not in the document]
positiveimpact: As we discussed, it doesn't really make a difference and we can leave them as they are :)

backend/hubs/models/hub.py
positiveimpact: I think we should use the help_text attributes to specify what background_default and background_paper does

frontend/src/components/hub/HubTabsNavigation.tsx
```diff
@@ -10,7 +10,7 @@ import HubsDropDown from "../indexPage/hubsSubHeader/HubsDropDown";
 
 const useStyles = makeStyles((theme) => ({
   root: {
-    background: theme.palette.primary.main,
+    background: theme.palette.background.default,
```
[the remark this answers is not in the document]
positiveimpact: The section of the post that you posted doesn't reference primary/background at all or am I missing something?
It seems to be an argument to use `background.main` but I don't see how it would be an argument to use `primary.main`

frontend/src/components/hub/HubTabsNavigation.tsx
```diff
@@ -19,14 +19,16 @@ const useStyles = makeStyles((theme) => ({
...
     "&.Mui-selected": {
       paddingLeft: theme.spacing(1),
       paddingRight: theme.spacing(1),
       "& .tabLabel": {
-        color: theme.palette.primary.main,
-        background: "white",
+        color: theme.palette.primary.main, // The color is set to "theme.palette.primary.main" as a default. Ideally, this should use `theme.palette.secondary.contrastText`.
+        // However, in the ClimateConnect hubs, the contrastText color is white.
+        background: "white", // The background is set to "white" as a default. Ideally, this should use `theme.palette.secondary.main`.
+        // However, in the ClimateConnect hubs, the secondary color is black, which results in an undesirable black background.
```
positiveimpact, line 31: Why should it use `secondary.main`?

frontend/src/themes/transformThemeData.ts
```diff
@@ -0,0 +1,29 @@
+import defaultTheme from "./hubTheme";
+
+export const transformThemeData = (data) => {
```
positiveimpact, line 3:
```suggestion
// transform theme data received from the API into a structured theme object
export const transformThemeData = (data) => {
```

backend/hubs/models/hub.py
```diff
@@ -367,3 +366,94 @@ class Meta:
 
     def __str__(self):
         return "%s" % (self.name)
+
+class HubThemeColor(models.Model): 
+    main = models.CharField(
+        help_text="main color",
```
positiveimpact, line 372: Which formats can we use here? hex and rgb? It would be helpful to specify that in the help text, for example like in this suggestion
```suggestion
        help_text="main color (use hex code, e.g. #FFF)",
```

frontend/src/themes/transformThemeData.ts
```diff
@@ -0,0 +1,29 @@
...
+      },
+      secondary: {
+        main: data?.secondary?.main,
+        light: data?.secondary?.light,
+        extraLight: data?.secondary?.extraLight,
+        contrastText: data?.secondary?.contrastText,
+      },
+      background: {
+        default: data?.primary?.background_default,
+        paper: data?.primary?.background_paper,
```
positiveimpact, line 23: This is a bit confusing because secondary.background_default and secondary.background_paper is never used. Maybe it's more intuitive if we directly add those to HubTheme without a foreign key between?

backend/hubs/models/hub.py
```diff
@@ -367,3 +366,94 @@ class Meta:
...
+
+    class Meta:
+        app_label = "hubs"
+        verbose_name = "Hub Theme Color"
+        verbose_name_plural = "Hub Theme Color"
+
+    def __str__(self):
+        return f"ThemeColor {self.id} - Main: {self.main}"
+
+class HubTheme(models.Model):
```
positiveimpact, line 422: It would be helpful to add a `name` field here which is also shown in django admin. Otherwise it will be really hard to tell them apart when we have many entries here. Name could for example be `Erlangen primary color`.

positiveimpact (approved):
Great job on this PR! Added some small comments, will test locally now

frontend/pages/hubs/[hubUrl].tsx
Comment (positiveimpact) [reply frontend/pages/hubs/[hubUrl].tsx:297]: Why don't we wrap `WideLayout` with `ThemeProvider` but only put it on this level?

backend/hubs/views/hub_views.py
```diff
@@ -81,3 +82,24 @@ def get(self, request, url_slug):
...
+                {"message": "Hub not found: {}".format(url_slug)},
+                status=status.HTTP_404_NOT_FOUND,
+            )
+        try:
+            hub_theme = HubTheme.objects.get(hub=hub)
+        except HubTheme.DoesNotExist:
+            return Response(
+                {"message": "Hub theme not found: {}".format(url_slug)},
+                status=status.HTTP_404_NOT_FOUND,
+            )
```
Idontker, line 101:
```suggestion
            # returning an nearly empty object indicating no custom theme data for a valid hub without customization.
            return Response(
                {"message": "Hub theme not found: {}".format(url_slug)},
                status=status.HTTP_200_NOT_FOUND,
            )
```
I would prefer to send a 200 back, so that in the frontend we can distinguish an faulty request (hub not found) from a no custom data :)

frontend/src/themes/transformThemeData.ts
```diff
@@ -0,0 +1,29 @@
+import defaultTheme from "./hubTheme";
+
+export const transformThemeData = (data) => {
+  const { palette, ...restOfDefaultTheme } = defaultTheme;
+  const customTheme = {
```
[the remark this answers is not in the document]
Idontker: additionally including the change, ts complains about us using `extraLight` as it is not part of the mui color/theme palatte type.

backend/hubs/views/hub_views.py
positiveimpact: Imo since there is no hub theme found, 404 seems more reasonable because we are trying to retrieve the hub theme. Either way it can be distinguished by the error message. I'd vote to keeping the 404

frontend/src/components/dashboard/Dashboard.tsx
```diff
@@ -104,6 +104,12 @@ const useStyles = makeStyles((theme: Theme) => {
     climateHubOption: {
       width: "100%",
     },
+    buttonLableColor: {
```
positiveimpact, line 107:
```suggestion
    buttonLabelColor: {
```

frontend/src/components/dashboard/Dashboard.tsx
```diff
@@ -135,7 +141,7 @@ const HoverButton = ({ items, label, startIcon }) => {
     <>
       <Button
         aria-haspopup="true"
-        className={classes.HoverButtonButton}
+        className={classes.buttonLableColor}
```
positiveimpact, line 144:
```suggestion
        className={classes.buttonLabelColor}
```

positiveimpact (approved):
Made some small changes, LGTM now. Great job on this PR, this was a quite difficult and very crucial task :clap:

PR #1400, opened 2024-12-23
fix yarn.lock changes 

fateme-ramezanpour:
## What and Why

The `yarn.lock` file had unexpected changes because a third-party library was manually removed from the package.json without using the appropriate Yarn command. 
This caused inconsistencies in the lock file.

## Solution
To fix this issue, the following command was run to properly remove the library and update:
`yarn remove [package_name]
`


ddhanesha:
@fateme-ramezanpour closing this PR for because I fixed it with my changes. https://github.com/climateconnect/climateconnect/pull/1409

PR #1404, opened 2025-01-03
Signup hub backend

Idontker:
## What
If a user signs up from a url containing the "hub" query parameter, e.g. "hub=Erlangen", the backend will try to look for a hub which url_slug matches the query parameter. If yes, the relation will be noted. Otherwise, nothing new happens (old behavior will be performed). 

## Why
This is a backend preparation task for an upcoming that allows a user to visit a hub specific signup page and then sign up to the whole platform and the specific hub.


backend/climateconnect_api/views/user_views.py
```diff
@@ -156,6 +156,12 @@ def post(self, request):
             user_profile.last_completed_tutorial_step = request.data[
                 "last_completed_tutorial_step"
             ]
+        if "hub" in request.data:
+            hub = Hub.objects.filter(url_slug=request.data["hub"])
```
ddhanesha, line 160: I would recommend doing this

```python
hub = Hub.objects.filter(url_slug=requrst.data["hub"]).first()
if hub:
     user_profile.related_hubs.add(hub)
```
Reduced the number of queries you will be making to the database.

ddhanesha (changes requested):

backend/climateconnect_api/views/user_views.py
Idontker: good point. adopted changes

positiveimpact (approved):
LGTM, great work!

PR #1405, opened 2025-01-07
login-visual-update

Idontker:
## What and Why
updated the login/sign in page to match the updated version of the sign up version. It uses the same building blocks like the sign up process.


ddhanesha (approved):
Looks good to me, let's get this merged!

positiveimpact (approved):
Great work on desktop but unfortunately this and the sign up page looked completely broken on mobile. Please always check if your changes also work on mobile.

PR #1402, opened 2025-01-01
Background on wide screen layout

Idontker:
## Dependencies
This PR depends on changes applied within:
- #1403 

Please, merge these branch first.

## What and Why
Added custom backgrounds for customer on signup, login page (will be present, as soon as its layout will be updated) and browse page (see images).
### Implementation 
Create a custom background component, that owns all the responsibility (hardcoded) on which pages and for which hubs it will show a custom background. Currently the implementation is hacky/hardcoded. As more custom hubs are created (and requirements for their backgrounds become clearer) one would have to update the code to be more dynamic.

For the auth pages, a query url param will be added for the custom Hub signUp:
`http://localhost:3000/signup?hub=prio1`

### Missing
A design and implementation of the background on Mobil is missing.

image
image

PR #1406, opened 2025-01-08
Updated colors on the singup page for the prio1 hub

Idontker:
## Dependencies
This branch relies on changes within
- #1402 
- #1403 (as #1402 depends on it) 

please, merge these branches first.

## What and Why
Updated the colors on the signup page for the prio1 hub. During this update, I added a color `contrast` to general and customHub theme, which is based on the `theme.palette.background.default_constrast`. This addition allows an easier usage of this color when while using a mui component:
```ts
<TextField color="primary"/>
<TextField color="contrast"/>
```

Each usage of "contrast" on a MUI requires a type extension like:
```ts
import { TextFieldProps } from "@mui/material/TextField";

declare module "@mui/material/TextField" {
  interface TextFieldPropsColorOverrides {
    contrast: true;
  }
}

```

## Follow Up Changes
The icon/image on the right side has not been updated yet. Additionally, I have not updated the login page or the account creation page. Changes to these will be addressed in a subsequent PR.


positiveimpact:
I merged this into #1407 and finished it over there, therefore closing this PR

PR #1411, opened 2025-01-13
Feature 1397 cusom color in project

fateme-ramezanpour:


positiveimpact (approved):
Overall decent work, some things weren't working properly on mobile yet. Also I don't think we should use background.default as text color if it's not absolutely necessary because I don't think this will scale very well with future custom hubs

PR #1412, opened 2025-01-15
Update README.md (reinstalling db)

Idontker:
## What and Why
Had some troubles after wiping while setting up the db again and wanted to document these for the future.


positiveimpact (approved):
LGTM, thanks for this update!

Issue #1378, opened 2024-11-20
Bug:Add a new hub type for hubs unrelated to location or sector hubs


fateme-ramezanpour:
**Describe the bug**
In our project, we have some additional hubs that currently do not have a proper type in the hub table.
To resolve this, we should consider adding a new hub type to the hub model to support these hubs.

**Proposed solution**
- Add foreign key pointing to the hub table from UserProfiles, Organizations, Projects
- Add button allowing a user to join a hub

PR #1413, opened 2025-01-16
added a new hub type "custom hub" that behaves like a location hub (for now)

Idontker:
like a location hub for now preparing prio1.

## What and Why
Started work / prepared #1378 by creating a custom hub type for the prio1 page that behaves like the locations hubs on the browse pages.


frontend/src/components/browse/BrowseContent.tsx
```diff
@@ -146,6 +147,8 @@ export default function BrowseContent({
   const [hash, setHash] = useState<string | null>(null);
   const [tabValue, setTabValue] = useState(hash ? TYPES_BY_TAB_VALUE.indexOf(hash) : 0);
 
+  const isLocationHubFlag = isLocationHubLikeHub(hubData?.hub_type);
```
positiveimpact, line 150: I think we should just call this `isLocationHubLikeHub` to avoid confusion

positiveimpact (approved):
LGTM, great work! After this PR custom hubs will still show projects from the location you add to them. I'm not sure yet whether that's a bug or a feature :D

PR #1416, opened 2025-01-16
Sharing projects on custom hubs

Idontker:
## Dependencies
This branch relies on changes within
- #1404 
- #1413 

please, merge these branches first.

## What and Why
Adjusted the share feature to allow linkage to custom hubs. Additionally, if one redirects to this page from the [hubUrl] or a ?hubName=... path, the hub will be pre-selected  (if valid).

Implements parts of the share/create feature of  #1414 

## Missing/Future Work of #1414 
1. Add custom theme to the share page.
2. Add ?hubName=... to the share link within the header

### Checkbox for the Prio1 selection
image

PR #1407, opened 2025-01-08
Signin/Login page: custom hub recoloring

Idontker:
## Dependencies
This branch relies on changes within
- #1405
- #1406 (depends on #1402 and #1403)

please, merge these branches first.

## What and Why
Added the background for the prio1 login page  and updated the colors on the signup page for the prio1 hub. 

## Follow Up Changes
The icon/image on the right side has not been updated yet. 
image
image

PR #1417, opened 2025-01-17
Custom hub theme on share project page

Idontker:
## Dependencies

This branch relies on changes within
- #1407
- #1416 

please, merge these branches first.

## What and Why
Currently, the share page uses the cc colors. On a custom hub (like prio1) it shall use the correct colors. This PR updates the colors to use the correct colors.
Additionally, I performed minor fixed. I also changed to color of the back button to a darker gray so that the primary color can be used. 

### Problems 

I struggled to change/fix the color of the Date and Time pickers on the 'Project Details' page.
A simple 
```js
 <DateTimePicker  color="contrast" {...} />
```
does not work as the component does not support setting the color this way via  
https://v5.mui.com/x/react-date-pickers/getting-started/
https://mui.com/x/react-date-pickers/date-picker/

Additionally, I realized that each `TextField` is using the primary color per default. One always needs to change it to contrast on pages related to custom hub. This might be fine for now but is prone to errors in the future. But, I do not have a proper solution yet.
### Now

image
image
image
image

### Previously 
image
image
image


positiveimpact (approved):
Good work, finished up the PR, ready to merge now :)

PR #1440, opened 2025-02-10
Feature dynamic link prio1

fateme-ramezanpour:
## Checked the following


positiveimpact (approved):
LGTM now, great work!

Issue #1441, opened 2025-02-11
Bug: login link on Project and organisation page with wrong redirect


tobiasrehm:
**Describe the bug**
When clicking login on the organistion or prooject page, after logging in it redirects to the general browse page instead of the organisation or project you were just looking at.

**To Reproduce**
1. Go to any project or organisation page while logged out
2. Klick login e.g. here: 

![Image](https://github.com/user-attachments/assets/aaecd218-1a64-4066-b931-3793ad9e6e84)
3. see error 

**Expected behavior**
Useres should be redirected to the page they were on before loggin in

PR #1450, opened 2025-02-21
Fix bug related to redirection on signup and signin pages

fateme-ramezanpour:
## Description
This PR addresses the issue [1441](https://github.com/climateconnect/climateconnect/issues/1441)


frontend/pages/signup.tsx
```diff
@@ -22,16 +21,17 @@ import { Container, Theme, useMediaQuery } from "@mui/material";
...
 import CustomAuthImage from "../src/components/hub/CustomAuthImage";
+import { extractHubFromRedirectUrl } from "../public/lib/hubOperations";
 
 export async function getServerSideProps(ctx) {
-  const hubUrl = ctx.query.hub;
-
-  const hubThemeData = await getHubTheme(hubUrl);
+  const { redirect } = ctx.query;
+  const hubUrl = extractHubFromRedirectUrl(redirect);
+  const hubThemeData = hubUrl ? await getHubTheme(hubUrl) : null;
```
Idontker, line 29: this breaks the normal link without a redirect.
```
http://localhost:3000/de/signin?hub=prio1
```
Additionally, when trying to access the login page via the nav bar from the hub-browse page, it uses 
```
http://localhost:3000/de/signin?redirect=%2Fhubs%2Fprio1&hub=prio1
```
and breaks, as it is comming from `/de/hubs/prio1` which does not contain a url param.

frontend/src/components/general/LoginNudge.tsx
```diff
@@ -23,28 +23,33 @@ type Props = {
...
+
   return (
     <div className={`${fullPage && classes.loginNudge} ${className}`}>
       <Typography className={fullPage ? classes.loginNudgeText : undefined}>
         {texts.please}{" "}
         <Link
           underline="always"
           color="primary"
-          href={`${getLocalePrefix(locale)}/signin${hub ? `?hub=${hub}` : ""}`}
+          href={`${getLocalePrefix(locale)}/signin?redirect=${encodedRedirectUrl}`}
```
Idontker, line 44: Not too sure, but I think we could also just use 
```suggestion
          href={`${getLocalePrefix(locale)}/signin?hub=${hub}&redirect=${encodedRedirectUrl}`}
```
to keep track of the current hub we are at, removing the need of `extractHubFromRedirectUrl` on the signin, signup and account created pages :) 

information may be redundant, but it should be simpler to implement :smile:

frontend/pages/signin.tsx
```diff
@@ -11,11 +11,13 @@ import { Container, Link, Theme, useMediaQuery } from "@mui/material";
...
 import { transformThemeData } from "../src/themes/transformThemeData";
 import Login from "../src/components/signup/Login";
+import { extractHubFromUrl } from "../public/lib/hubOperations";
 
 export async function getServerSideProps(ctx) {
-  const hubSlug = ctx.query.hub;
   const message = ctx.query.message;
   const message_type = ctx.query.message_type;
+  const { redirect } = ctx.query;
+  const hubSlug = extractHubFromUrl(redirect);
```
Idontker, line 20:
```suggestion
  const { hub: hubSlug } = ctx.query;
```

frontend/pages/signin.tsx
```diff
@@ -11,11 +11,13 @@ import { Container, Link, Theme, useMediaQuery } from "@mui/material";
 import getHubTheme from "../src/themes/fetchHubTheme";
 import { transformThemeData } from "../src/themes/transformThemeData";
 import Login from "../src/components/signup/Login";
+import { extractHubFromUrl } from "../public/lib/hubOperations";
```
Idontker, line 14:
```suggestion

```

frontend/public/lib/hubOperations.ts
```diff
@@ -16,3 +16,19 @@ export async function getAllHubs(locale: any, just_sector_hubs?: boolean) {
...
+  if (hubsPathMatch) {
+    return hubsPathMatch[1];
+  }
+  // Check if there's a query parameter hub= in the URL
+  const hubQueryMatch = decodedRedirect.match(/[?&]hub=([^&#]+)/);
+  if (hubQueryMatch) {
+    return hubQueryMatch[1];
+  }
+  return null;
+}
```
Idontker, line 34:
```suggestion

```
I do not see the upsite vs. just using `ctx.query`

frontend/src/components/general/LoginNudge.tsx
```diff
@@ -23,28 +25,32 @@ type Props = {
...
 export default function LoginNudge({ whatToDo, fullPage, className }: Props) {
   const classes = useStyles();
   const { locale } = useContext(UserContext);
   const texts = getTexts({ page: "general", locale: locale });
-  const urlParams = new URLSearchParams(window.location.search);
-  const hub = urlParams.get("hub");
+  const router = useRouter();
+  const currentPath = router.asPath;
+  const encodedRedirectUrl = encodeURIComponent(currentPath);
+  const hubUrl = extractHubFromUrl(currentPath);
```
Idontker, line 36: either, we can be sure, that we can get the hub via: 
```
const urlParams = new URLSearchParams(window.location.search);
const hub = urlParams.get("hub");
```
or the parent component should be dealing with that, shouldn't it? It might know it's hub property anyways.

frontend/src/components/general/LoginNudge.tsx
Idontker: I think, the issue still holds and it was not addressed.

Idontker (changes requested):
Not sure aboutt the `extractHubFromUrl` function. It might be not helpful

fateme-ramezanpour:
Last time I checked, this approach was working without any issues. 
If you believe your approach is better, I'm totally open to it — it would be great if you could create a separate PR with your suggested changes so we can move forward more easily.

Idontker:
Sure, i will create an alternative suggestion so we can decide which version to keep :)

fateme-ramezanpour:
I closed this PR. We decided to go with a different approach, so Karol created another PR for it.

PR #1454, opened 2025-02-24
fixed the shared organization url for prio1

fateme-ramezanpour:
## Checked the following


frontend/src/components/communication/MessageContent.tsx
Idontker, line 1: Do not have to context of this change. Is there a good reason why the context will be null here?

Idontker (approved):
Aside some minor things, LGMT.
Added a type for hubUrl in SocialMediaShareButtonProps, as it was missing.

Can be merged after adressing the comments

PR #1461, opened 2025-02-26
Fix custom prio one background

Idontker:
## Checked the following
- [x] Pages affected by this change work on mobile when logged in
- [x] Pages affected by this change work on mobile when logged out
Did not test the last yet. It is custom for the Prio1 Hub and should only show up on its page.

## Pages affected:
- /de/hubs/prio1
- /en/hubs/prio1
- /hubs/prio1

## What and Why

Updated background for the prio1 browse hub page. Now includes:
- correctly renders the triangle / trapezoid shape
- background height is no longer hard coded
- includes Prio1 Icon if logged in and no supporters are present
- includes Prio1 Icon if not logged in and screen width is greater then 1550px (sufficient space)

## Note 
One could improve Browse Page when logged out and (770px < screen width < 900px)

image

### DIFF
--- a/frontend/pages/hubs/[hubUrl].tsx
+++ b/frontend/pages/hubs/[hubUrl].tsx
@@ -361,7 +361,12 @@ export default function Hub({
           </BrowseContext.Provider>
         </div>
         {isSmallScreen && (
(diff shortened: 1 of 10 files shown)


positiveimpact (approved):
LGTM, good work!

PR #1470, opened 2025-02-28
Remove project status filter

Idontker:
## Checked the following
- [x] Pages affected by this change work on mobile
- [x] Pages affected by this change work when logged out
- [x] Pages affected by this change work when logged in
- [x] Pages affected by this change work with custom theme and default theme

## What and Why

Removed `status` from the browse content page and browse hub page

PR #1471, opened 2025-02-28
Remove project status share and edit project

Idontker:
## Checked the following

## What and Why
Merge it after #1470.

Fully removed the project status from the front end. Added a dummy value for the back end so that the backend does not break until we have not deleted project status from the backend.


Idontker:
LGTM

HaraldWalker (approved):
Looking good. Tested locally. 
Found bug #1685 but is not related to his PR.

PR #1473, opened 2025-03-06
Remove collaboration filter frontend

Idontker:
## Checked the following

## What and Why

Removed collaboration filter as it was broken and did not really provide impact. It was only removed within the frontend. There were no changes in the backend, as the information may be useful in the future


positiveimpact (approved):
LGTM

PR #1474, opened 2025-03-06
Remove project status backend

Idontker:
## Checked the following

## What and Why
Project status is not used anymore and has generally no use. Removed it from the backend.
- Project status table is not activly used anymore
- Allowed project status within the project objects/table to be null
- Removed it in any referenced position (e.g. serializers, gen_test_data, ...)

## Note
- It depends on #1471. Please merge this first.
- The ProjectStatus table will be removed in the future. As this is a destructive change, it will be postponed. Otherwise a reviewer would delete the ProjectStatus reference of all his local projects during a review without an option of recovery.
- I've checked it locally on my machine and it works fine. But, I would really appreciate a proper second look, as it affects the database and projects in general  :)


frontend/src/components/project/ProjectContent.tsx
```diff
@@ -267,25 +267,17 @@ export default function ProjectContent({
...
-                })}
-              </Typography>
-            )
-          }
-          {
-            // TODO: update remove
-            project.end_date && project.status.key === "cancelled" && (
-              <Typography>{texts.cancelled} :(</Typography>
-            )
-          }
```
Idontker: image
 
hopefully, github will update their diffing soonish ...

HaraldWalker (approved):

Issue #1394, opened 2024-12-18
Update FAQ page header 


tobiasrehm:
The Header on the FAQ page should be in the same style as all other about pages. 
Same Header grafic and Heading style as on the Team page: https://climateconnect.earth/team

nirnaeth:
I'll pick up this one as spoken over video call :)

PR #1494, opened 2025-03-23
1394 - adjust FAQ page as similar as possible to generated pages

nirnaeth:
## Checked the following

## What and Why

**Image**
Took the image from one of the other pages. I duplicated it in such a way that fills the whole screen. I tried to reproduce the look and feel of the image as close as I could to the generated ones. 

**Content**
Added padding to the top of the text container so that the content looks as close as possible to the content of the other static generated pages.

Issue #1394


frontend/pages/faq.tsx
```diff
@@ -14,53 +15,37 @@ import HeaderImage from "../src/components/staticpages/HeaderImage";
...
-      height: "100%",
-      [theme.breakpoints.down("lg")]: {
-        background: "rgba(255, 255, 255, 0.8)",
-      },
+      top: "45%",
+      left: "10%",
+      maxWidth: "400px",
+      minWidth: "250px",
+      borderRadius: "5px",
+      backgroundColor: "#207178",
```
Idontker, line 25: position absolute can easily break on different screen sizes. I would rather prefer a different solution, which inserts into the normal layout and then uses `transform` to push it up:

```suggestion
            display: "flex",
      alignItems: "left",
      transform: "translateY(-50%)",
 ```
In this case, using this class and its related div as a wrapper for the row segment makes more sense. Therefore, I would move the `backgroundColor` and `borderRadius` to the div below.

File: frontend/pages/faq.tsx [hunk 2]
```diff
@@ -14,53 +15,37 @@ import HeaderImage from "../src/components/staticpages/HeaderImage";
...
     },
     headerTextInnerContainer: {
       height: "100%",
       display: "flex",
```
Idontker, line 29:
```suggestion
      borderRadius: "5px",
      backgroundColor: "#207178",
      paddingRight: "25px",
```
See above. This is now the "new card" while the one above is the line wrapper.

frontend/pages/faq.tsx
```diff
@@ -2,6 +2,7 @@ import { Container, Typography } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
 import Cookies from "next-cookies";
 import React, { useContext } from "react";
+import WebflowPage from "../src/components/webflow/WebflowPage";
```
Idontker, line 5:
```suggestion
```
not used, or is it?

frontend/pages/faq.tsx
```diff
@@ -70,6 +55,7 @@ const useStyles = makeStyles((theme) => {
         fontSize: 14,
         fontWeight: "normal",
       },
+      paddingTop: "70px",
```
Idontker, line 58: If you agree with my improvements, the paddingTop is redundant and can be removed
```suggestion

```

Idontker (changes requested):
Hey there. I really liked your work. CSS can sometimes be a bit messy/annoying to work with. I suggest changing some CSS and omit using position absolute. 

Also, I liked the rework on the header. Kinda smart solution to archive having only one sun :smile:

With my suggestions, It looks like this (see image). Feel free to ask if you have questions :smile: 
image

frontend/pages/_app.tsx
```diff
@@ -14,7 +14,7 @@ import UserContext from "../src/components/context/UserContext";
 import theme from "../src/themes/theme";
 import { CcLocale } from "../src/types";
 import * as Sentry from "@sentry/react";
-import "../devlink/global.css";
+//import "../devlink/global.css";
```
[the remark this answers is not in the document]
nirnaeth: oh sorry, this is something I forgot! thank you for catching it :)

nirnaeth:
thank you, I haven't touched css in ages, so it was a bit tough. 
I'll apply the changes you suggested and clean the PR and/or add comments on why certain things where necessary. and thank you for your patience!

Idontker (changes requested):
See the comments I made :)
(needed to write a command to set the status to reviewed and awaiting changes)

PR #1495, opened 2025-03-23
hotfix for locaiton lookup, instable api call no longer breaks the search

Idontker:
## Draft
Currently, I am using a POST request that will redirect to code of the GET request. This way, we can send the geojson coordinates form the client to the backend which is used for locations that are unknown to the backend. 

Another option that has been (semi-)discarded is to rely on the GET implementation alone and add and encode the location data in the URL itself. This might result in inconsistent behavior due to the number of chars in the URL reaching a length of over 1000. Usually, 1000 chars is fine, but some infrastructure might cut of the URL after a length of 2000. Due to the numerous technologies that handles URL and search params, this might be inconsistent. 

Additionally, GET requests shall not / can not contain a body. Therefore, we can not use the body to transmit data. The POST approach is using the body to transmit data.    

## Checked the following

## What and Why

hotfix for locaiton lookup, instable api call no longer breaks the search

### DIFF
--- a/frontend/public/lib/getDataOperations.ts
+++ b/frontend/public/lib/getDataOperations.ts
@@ -25,16 +25,6 @@ export async function getDataFromServer({
     // &category=Lowering%20food%20waste&
     url += urlEnding;
   }
-  // if (location instanceof Object && !(location instanceof String)) {
-  //   // use base64 encoding to pass the location object
-
-  //   console.log("help me please, this is a mess");
-  //   const encodedLocation = btoa(JSON.stringify(location));
-  //   const encodedLocation2 = encodeURIComponent(JSON.stringify(location));
-  //   console.log(encodedLocation.length, encodedLocation2.length);
-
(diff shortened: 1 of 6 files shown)


frontend/public/lib/getDataOperations.ts
```diff
@@ -16,10 +25,24 @@ export async function getDataFromServer({ type, page, token, urlEnding, hubUrl,
...
+
+  //   url += `&location=${encodeURIComponent(encodedLocation)}`;
+  // }
 
   try {
     console.log(`Getting data for ${type} at ${url}`);
-    const resp = await apiRequest({ method: "get", url, token, locale });
+
+    const resp = location
+      ? await apiRequest({ method: "post", url, payload: location, token, locale })
```
Idontker, line 43: if the location filter is set, search for locations via post

positiveimpact (approved):
LGTM

PR #1506, opened 2025-03-26
Handle browse content api error with user feedbacked

Idontker:
## Checked the following

## What and Why
The user is now getting informed, if the search on a browse page fails. Additionally, performed some refactoring.

### DIFF
--- a/frontend/public/lib/filterOperations.ts
+++ b/frontend/public/lib/filterOperations.ts
@@ -241,6 +241,9 @@ export async function applyNewFilters({
     };
   } catch (e) {
     console.log(e);
-    throw e;
+    // TODO: in the future, throw the error
+    // but make sure that the calling component catches
+    /// the error and gives feedback to the user
+    // throw e;
   }
 }
--- a/frontend/public/lib/filterOperations.ts
+++ b/frontend/public/lib/filterOperations.ts
@@ -5,6 +5,7 @@ import { getDataFromServer } from "./getDataOperations";
 import { membersWithAdditionalInfo } from "./getOptions";
 import { getInfoMetadataByType, getReducedPossibleFilters } from "./parsingOperations";
 import { encodeQueryParamsFromFilters } from "./urlOperations";
+import getTexts from "../texts/texts";
 
 const getLocationFilterUrl = (location) => {
   /*Pass place id. If the place id is found in our db we can use it's polygon,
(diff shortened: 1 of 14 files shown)


frontend/public/lib/filterOperations.ts
```diff
@@ -241,6 +242,9 @@ export async function applyNewFilters({
     };
   } catch (e) {
     console.log(e);
-    throw e;
+    const texts = getTexts({ page: "general", locale: locale });
+
+    handleSetErrorMessage(texts.error_during_search);
+    throw new Error(texts.error_during_search_short);
```
Idontker, line 248: get texts that are displayed via the setErrorMessage.

frontend/public/lib/locationOperations.ts
```diff
@@ -354,5 +354,6 @@ export async function getLocationFilteredBy(query) {
     return full_location;
   } catch (e) {
     console.log(e);
+    return null;
```
Idontker, line 357: changed return undefined to return null, if the backend fails to lookup the location by place_id.

frontend/public/texts/general_texts.json
Idontker, line 1: added texts that will be displayed in case of an error during search

frontend/src/components/browse/BrowseContent.tsx
```diff
@@ -69,7 +69,6 @@ export default function BrowseContent({
   initialOrganizations,
   initialProjects,
   customSearchBarLabels,
-  errorMessage,
```
Idontker, line 72: `errorMessage` is provided by the FilterContext. It does not need to be propagated

frontend/src/components/browse/BrowseContent.tsx
```diff
@@ -103,12 +102,9 @@ export default function BrowseContent({
 
   const token = new Cookies().get("auth_token");
 
-  const {
-    filters,
-    handleUpdateFilterValues,
-    handleSetErrorMessage,
-    handleApplyNewFilters: applyNewFilters,
-  } = useContext(FilterContext);
```
Idontker, line 111: Same here, `handleUpdateFilterValues` is provided by FilterContext and not used in BrowseConent. It was only used to propagate down to components. 
Therefore, it can be removed and looked up on demand.

frontend/src/components/browse/BrowseContent.tsx
```diff
@@ -200,7 +196,8 @@ export default function BrowseContent({
       };
       setNonFilterParams(splitQueryObject.nonFilters);
       if (splitQueryObject?.nonFilters?.message) {
-        showFeedbackMessage({
+        // ? is needed due to the context type {showFeedbackMessage?: (..) => void;}
+        showFeedbackMessage?.({
```
Idontker, line 200: added a proper typing for showFeedbackMessage. This resulted in erros due to showFeedbackMessage being optional within the context. Maybe one could this fix in the future, in a better way. Not the main focus of this PR.

PR #1527, opened 2025-04-14
Issue 1441 loginlink prio1 v2

Idontker:
## What and Why
Alternative solution to the #1441 issue that lets the page components do query / extracting of the `hubUrl` instead of performing it inside components lower down.

### DIFF
--- a/frontend/src/components/communication/CommentInput.tsx
+++ b/frontend/src/components/communication/CommentInput.tsx
@@ -8,7 +8,7 @@ import {
   useMediaQuery,
 } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
-import React, { useContext, useState } from "react";
+import React, { useContext, useEffect, useState } from "react";
 import { getImageUrl } from "../../../public/lib/imageOperations";
 import getTexts from "../../../public/texts/texts";
 import theme from "../../themes/theme";
@@ -68,6 +68,12 @@ function CommentInput({
   const texts = getTexts({ page: "communication", locale: locale });
   const isSmallScreen = useMediaQuery<Theme>(theme.breakpoints.down("md"));
   const [showSendHelper, setShowSendHelper] = useState(false);
+
+  let hubUrl = "";
+
+  useEffect(() => {
+    hubUrl = new URLSearchParams(window.location.search).get("hub") ?? "";
+  }, [window.location.search]);
 
   const onCurCommentChange = (e) => {
     setCurComment(e.target.value);
@@ -184,7 +190,7 @@ function CommentInput({
         </form>
       </div>
     );
-  else return <LoginNudge whatToDo={texts.to_write_a_comment} />;
+  else return <LoginNudge whatToDo={texts.to_write_a_comment} hubUrl={hubUrl} />;
 }
 
 export { CommentInput as default, INFO_TEXT_SIZES };
--- a/frontend/src/components/communication/CommentInput.tsx
+++ b/frontend/src/components/communication/CommentInput.tsx
@@ -61,6 +61,7 @@ function CommentInput({
   hasComments,
   infoTextSize,
   useIconButton,
+  hubUrl,
 }: any) {
   const classes = useStyles();
   const [curComment, setCurComment] = React.useState("");
@@ -68,12 +69,6 @@ function CommentInput({
   const texts = getTexts({ page: "communication", locale: locale });
   const isSmallScreen = useMediaQuery<Theme>(theme.breakpoints.down("md"));
   const [showSendHelper, setShowSendHelper] = useState(false);
-
-  let hubUrl = "";
-
-  useEffect(() => {
-    hubUrl = new URLSearchParams(window.location.search).get("hub") ?? "";
-  }, [window.location.search]);
 
   const onCurCommentChange = (e) => {
     setCurComment(e.target.value);
--- a/frontend/src/components/general/LoginNudge.tsx
+++ b/frontend/src/components/general/LoginNudge.tsx
@@ -5,7 +5,6 @@ import { getLocalePrefix } from "../../../public/lib/apiOperations";
 import getTexts from "../../../public/texts/texts";
 import UserContext from "../context/UserContext";
 import { useRouter } from "next/router";
-import { extractHubFromUrl } from "../../../public/lib/hubOperations";
 
 const useStyles = makeStyles((theme) => {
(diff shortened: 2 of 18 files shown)


frontend/src/components/general/LoginNudge.tsx
```diff
@@ -24,34 +23,37 @@ type Props = {
...
   const texts = getTexts({ page: "general", locale: locale });
   const router = useRouter();
   const currentPath = router.asPath;
   const encodedRedirectUrl = encodeURIComponent(currentPath);
-  const hubUrl = extractHubFromUrl(currentPath);
+
+  let path_to_signin = `${getLocalePrefix(locale)}/signin?`;
+  let path_to_signup = `${getLocalePrefix(locale)}/signup?`;
+
+  path_to_signup += `redirect=${encodedRedirectUrl}`;
```
fateme-ramezanpour, line 40: We don't need a redirect after signup, since the user should check their email to verify their account.

frontend/src/components/layouts/ContentImageSplitLayout.tsx
```diff
@@ -2,7 +2,15 @@ import React from "react";
 import { Grid, Theme, useMediaQuery } from "@mui/material";
```
fateme-ramezanpour, line 2: Is this file related to the redirect url?

fateme-ramezanpour:
I think adding extra hub params to url make it redundant but it make code simpler for now, I'm ok with move forward to this way.

frontend/src/components/general/LoginNudge.tsx
fateme-ramezanpour: and there are more LoginNudge component that should be contain hubUrl,
(LoginNudge that we have inside the edit project tabs and ....)

frontend/src/components/layouts/ContentImageSplitLayout.tsx
Idontker: Not really. You added rightfully the ToDo's :+1:  and I figured that we will forget about ToDo comments so I moved the style changes.

frontend/src/components/general/LoginNudge.tsx
Idontker: Both valid points. My bad. I think, my intelligent sense did not spot all the occurrences of the Login Nudge. Ill fix it :smile:

Idontker:
@fateme-ramezanpour i have updated this PR. What do you think about the changes? Can we merge it :+1:

frontend/src/components/communication/CommentInput.tsx
```diff
@@ -69,6 +69,12 @@ function CommentInput({
   const isSmallScreen = useMediaQuery<Theme>(theme.breakpoints.down("md"));
   const [showSendHelper, setShowSendHelper] = useState(false);
 
+  let hubUrl = "";
```
fateme-ramezanpour, line 72: you're retrieving hubUrl inside the useEffect, but since the parent component (CommentsContent) already has access to hubUrl, would it make sense to pass it down as a prop instead?
That might make the code a bit simpler since you already have the value. What do you think?

frontend/src/components/general/LoginNudge.tsx
```diff
@@ -24,34 +23,36 @@ type Props = {
...
-  const hubUrl = extractHubFromUrl(currentPath);
+
+  let path_to_signin = `${getLocalePrefix(locale)}/signin?`;
+  let path_to_signup = `${getLocalePrefix(locale)}/signup?`;
+
+  path_to_signin += `redirect=${encodedRedirectUrl}`;
+
+  if (hubUrl && hubUrl !== "") {
+    path_to_signin += `&hub=${hubUrl}`;
+    path_to_signup += `&hub=${hubUrl}`;
```
fateme-ramezanpour, line 44: I think this line could be simplified to`path_to_signup += `hub=${hubUrl}`;`(just removing &)
since we’re only appending the hub query param. 
Do we have some problem here that we need to keep `&` ?
currently signup link look likes:` de/signup?&hub=prio1`

frontend/src/components/communication/CommentInput.tsx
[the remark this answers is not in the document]
Idontker: True, I will fix it tomorrow 👍

frontend/src/components/general/LoginNudge.tsx
Idontker: Yeah, you are right. It was copy & paste on my side. & does not hurt but there is no reason for it

Idontker:
@fateme-ramezanpour fixed your comments :+1:

frontend/src/components/general/LoginNudge.tsx
```diff
@@ -24,34 +23,36 @@ type Props = {
...
   const classes = useStyles();
   const { locale } = useContext(UserContext);
   const texts = getTexts({ page: "general", locale: locale });
   const router = useRouter();
   const currentPath = router.asPath;
   const encodedRedirectUrl = encodeURIComponent(currentPath);
-  const hubUrl = extractHubFromUrl(currentPath);
+
+  let path_to_signin = `${getLocalePrefix(locale)}/signin?`;
+  let path_to_signup = `${getLocalePrefix(locale)}/signup?`;
```
fateme-ramezanpour, line 38: When we don't have a hub for signup, after click on signup user redirect to :
`http://localhost:3000/signup?`
I think we should remove `?` from this line:
`let path_to_signup = `${getLocalePrefix(locale)}/signup?`;`
 and put it before hub:
`path_to_signup += `?hub=${hubUrl}`;`

fateme-ramezanpour:
In `manageProjectMembers` page, when there is no hub, after the user clicks on sign-in or sign-up, the link generated is:
`signin?redirect=......Fthis-is-a-text-for-draft&hub=undefined`

Here, the hub parameter is incorrectly set to undefined.
(To see the  manageProjectMembers.tsx page , please open one project, Go to the Team tab and click on MANAGE MEMBERS button)

Idontker:
fixed the ? comment. 
But, I am not able to reproduce the `hub=undefined` bug. It should be not possible / might have resulted in another spot setting the hub param to the string `"undefined"`. Otherwise it would not pass the line:
```js
  if (hubUrl && hubUrl !== "") {
    ...
  }
```
As you can see in the dev-tools, the string `"undefined"` is truefy.
image

Issue #1526, opened 2025-04-14
Bug: Settings page does not inherit hub specific theme


Idontker:
**Describe the bug**
See title.

**To Reproduce**
1. Go to https://climateconnect.earth/settings?hub=prio1

**Expected behavior**
Custom hub style is used.
 
**Screenshots**

![Image](https://github.com/user-attachments/assets/609c81a8-64a5-42c7-879e-38f28d34f69a)

PR #1528, opened 2025-04-15
settings page now inherits custom hub design

Idontker:
## Checked the following

## What and Why
Adresses #1526

### DIFF
--- a/frontend/pages/settings.tsx
+++ b/frontend/pages/settings.tsx
@@ -7,17 +7,27 @@ import SettingsPage from "../src/components/account/SettingsPage";
 import UserContext from "../src/components/context/UserContext";
 import LoginNudge from "../src/components/general/LoginNudge";
 import Layout from "../src/components/layouts/layout";
+import WideLayout from "../src/components/layouts/WideLayout";
+import getHubTheme from "../src/themes/fetchHubTheme";
+import { transformThemeData } from "../src/themes/transformThemeData";
 
 export async function getServerSideProps(ctx) {
   const { auth_token } = NextCookies(ctx);
+  const { hub } = ctx.query;
+
+  const hubThemeData = await getHubTheme(hub);
+  console.log("hubThemeData", hubThemeData);
+
   return {
     props: {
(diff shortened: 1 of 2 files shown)


frontend/pages/settings.tsx
```diff
@@ -26,20 +34,30 @@ export default function Settings({ settings }) {
   const texts = getTexts({ page: "settings", locale: locale });
   if (user)
     return (
-      <Layout title={texts.settings} message={message} noSpacingBottom>
```
fateme-ramezanpour, line 29: Wouldn't it be better to update the existing `Layout` component here and use that instead of switching to `WideLayout`?
In this [PR](https://github.com/climateconnect/climateconnect/pull/1518) , I updated the `Layout` component so we could reuse it in cases like this.

frontend/pages/settings.tsx
```diff
@@ -26,20 +34,30 @@ export default function Settings({ settings }) {
...
+      </WideLayout>
     );
   else
     return (
-      <Layout title={texts.please_log_in} hideHeadline>
+      <WideLayout
+        title={texts.please_log_in}
+        message={message}
+        hubUrl={hubUrl}
+        customTheme={hubThemeData ? transformThemeData(hubThemeData) : undefined}
```
fateme-ramezanpour, line 57: We also need this line to change Header backgroundColor: 

  `headerBackground={hubUrl === "prio1" ? "#7883ff" : "#FFF"}`

fateme-ramezanpour:
These 3 buttons: `Change preferences` , `Change cookie settings` and `Edit profile page` appear stretched.
and It would be great to update the redirect URLs for buttons, as users lose the current hub context after clicking them.

frontend/pages/settings.tsx
Idontker: True. Lets get the other PR first merged and then ill revert it to the `Layout` component.

Idontker:
Ill push it back for now, change it and reopen it as soon as #1518 is merged

PR #1529, opened 2025-04-15
removed raffleterms

Idontker:
## What and Why
Raffleterms were not used anymore. Therefore, they've been removed


positiveimpact (approved):
LGTM

Issue #1587, opened 2025-08-04
Make sub hub filters work properly with organizations and members


positiveimpact:
1. Sub hubs are displayed like sector hubs (menu layout changes between main hub and sub hub -> should stay in location hub mode
2. Filters on organizations and members only take into account sub hub filter (location filter is missing)
3. When you are on the "organizations" tab and click on a hub, it will automatically switch tot he projects tab. It should stay in the current tab

PR #1589, opened 2025-08-04
 fixed parent hub search for organization view @Idontker

Idontker:
## Checked the following

## What and Why
Fixed bug #1587 for organizations


positiveimpact (approved):
Looks good to me! I added some code to also address these issues from #1599:

1. Sub-Hubs use their parent hubs layout now - if the parent hub is a location hub it will use location hub's tab and filter layout. Before this switching from parent hub to child hub would change the layout
2. You now don't switch back to the project tab when you change to a different sub-hub.

PR #1590, opened 2025-08-04
removed unused sector filter, and updated loction filter to be more 

Idontker:
…esistant if the hubs location is missing e.g. countryor the multi_polygon.

## Summary by CodeRabbit

* **Refactor**
  * Improved location-based filtering for Location Hub to conditionally apply country, geographical area, and distance criteria based on available data.
  * Removed separate sectors filtering pathway from query parameters.


HaraldWalker:
Added a new feature request #1688 for filtering topics. This PR should then leave the sector filter but can still provide the location filter improvement.

PR #1612, opened 2025-08-25
added query paramas for redirect after signin

fateme-ramezanpour:
## Description
This PR fixes the redirection link after a user logs in. 
They will now be redirected back to the specific page they were trying to access before logging in.

This issue was previously addressed in PR [#1527](https://github.com/climateconnect/climateconnect/pull/1527), which was merged but did not successfully integrate the changes into the main branch. This PR reintroduces those changes to resolve the redirection issue permanently.

**Another problem**
When redirecting users, the /de/ locale was being duplicated in the URL.
For example, a redirect from `/de/some-page` would go to `.../de/de/new-page`.
This was caused by the redirect function not properly stripping the locale from `window.location.pathname.`

## Changes

- We no longer need a redirect after signup, as it's now handled by the `accountcreated` page.

- I've moved the redirect URL code to `apiOperations` so it can be used in other places.

## Checked the following

## Summary by CodeRabbit

- New Features
  - Locale-aware login redirects across Chat, Inbox, and Project pages for a smoother sign-in experience.
  - Login prompt now returns users to their current page after sign-in and preserves hub context when available.
- Bug Fixes
  - Prevented improper URL encoding when hub is undefined.
- Refactor
  - Centralized redirect URL generation via a shared helper for consistent behavior.
- Behavior Changes
  - Some signup prompts no longer include automatic redirect back or hub parameters, simplifying the flow.

### DIFF
--- a/frontend/pages/chat/[chatUUID].tsx
+++ b/frontend/pages/chat/[chatUUID].tsx
@@ -1,7 +1,7 @@
 import NextCookies from "next-cookies";
 import React, { useContext, useEffect } from "react";
 import Cookies from "universal-cookie";
-import { apiRequest, redirect, sendToLogin } from "../../public/lib/apiOperations";
+import { apiRequest, redirect, sendToLogin, getRedirectUrl } from "../../public/lib/apiOperations";
 import { getMessageFromServer } from "../../public/lib/messagingOperations";
 import getTexts from "../../public/texts/texts";
 import MessagingLayout from "../../src/components/communication/chat/MessagingLayout";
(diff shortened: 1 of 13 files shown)


positiveimpact (changes requested):
The redirect is still broken when I'm on the german version of the platform and try to send the project a message. In this case I get redirected to "https://climateconnect.earth/de/de/projects/<project_slug>". The "/de/de" causes a 404 error.
It would be great to include this in this PR

positiveimpact:
When logging in on any page on the live website, I get successfully redirected to that page again after login

HaraldWalker (approved):
LGTM and verified locally after reproducing the issues on production.

PR #1620, opened 2025-09-12
fix: devlink without webflow token (part 1)

KathiAurina:
Addresses https://github.com/climateconnect/climateconnect/issues/1635 

## Checked the following

## What and Why
if people don't have a webflow token and still want to collaborate, this is the fix of devlink imports in 3 of the 4 neccessary files

## Summary by CodeRabbit

- New Features
  - Introduced a feature flag to enable Devlink-powered pages/components.
  - Added a fallback UI when Devlink is disabled or tokens are missing, including a Browse button.
- Refactor
  - Switched to environment-gated, runtime loading for Devlink components and provider to keep the default experience unchanged unless enabled.
- Style
  - Loads Devlink-specific styles only when the feature flag is on.
- Chores
  - Exposed the ENABLE_DEVLINK environment variable to the client via configuration.

### DIFF
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -15,12 +15,13 @@ import theme from "../src/themes/theme";
 import { CcLocale } from "../src/types";
 import * as Sentry from "@sentry/react";
 
-
 //only bundle global.css if ENABLE_DEVLINK in .env is true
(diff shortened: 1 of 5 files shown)


Idontker:
@KathiAurina 

the lint check failed. I fixed it via `yarn format`. This is helpful, as it reduces the number of changes a reviewer has to read due to format changes.

PR #1628, opened 2025-09-30
added hamlet type to the first part of search name

fateme-ramezanpour:
## Checked the following


positiveimpact:
Should we use exact location names everywhere, including in the search bar?

positiveimpact:
We need to check whether it makes sense to just not show the middle part in cases where it's the same as the first part -> we need to test this with different locations

Issue #1653, opened 2025-10-22
Add support for PNG hub logos in header


HaraldWalker:
**Is your feature request related to a problem? Please describe.**
Some SVG hub logos are not being rendered nicely on all browsers (esp. Safari has some issues with some SVGs). Also sometimes the SVG is actually larger than the PNG version depending on its complexity, increasing load time.

**Describe the solution you'd like**
Make it possible to also use PNGs logos for a custom hub. 

**Describe alternatives you've considered**
With the SVG that had a rendering issue in Safari we tried multiple ways to create or improve the SVG without success. Files were too large and still had some rendering issues. 

Part of this task is also the use a PNG logo for the hub where we saw the issue.

PR #1655, opened 2025-10-23
Replace Perth SVG logo with optimized version

HaraldWalker:
## Checked the following
- [X] Pages affected by this change work on mobile

Verified the new version also on mobile. This PR only swaps the SVG and does not introduce any other code changes. 

## What and Why
Addresses Safari rendering and file size issue of old perth_logo.svg. 
The new logo renders fine in Safari and other browsers and is much smaller (27 KB instead of 410 KB). 

Going for this solution instead of implementing PNG support just for this case as originally considered in #1653. 

Old logo version in Safari:
perth-logo-svg-old-version-safari

New logo version in Safari:
perth-logo-svg-new-version-safari


frontend/public/images/hub_logos/ch_perth_logo.svg
HaraldWalker: Don't know why Github shows  `W: 300px | H: 132px` as dimension. In the SVG itself it is: `width="272px" height="120px"` (twice the height of what we use)

PR #1709, opened 2025-12-17
Location translation

KathiAurina:
## Checked the following

## What and Why

Implementing changes for #1666

### DIFF
--- a/backend/climateconnect_api/management/commands/add_locale_to_location_translations.py
+++ b/backend/climateconnect_api/management/commands/add_locale_to_location_translations.py
@@ -16,11 +16,12 @@ from climateconnect_api.models.language import Language
 
 # from django.db import transaction
 from location.models import Location, LocationTranslation
+from django.conf import settings
+
 from location.utility import format_location_name
 from location.utility import format_translation_data
 
 NOMINATIM_DETAILS_URL = "https://nominatim.openstreetmap.org/lookup"
-CUSTOM_USER_AGENT = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36"
 MAPPING_TABLE_PATH = Path(__file__).parent / "osm_lookup_tables" / "mapping.csv"
 
 
@@ -88,7 +89,7 @@ def translate_locations(locs: list["Location"], locale: str):
             "accept-language": locale,
         }
 
-        headers = {"User-Agent": CUSTOM_USER_AGENT}
+        headers = {"User-Agent": settings.CUSTOM_USER_AGENT}
 
         # Retry-Logic for Nominatim (503 errors may occur when too many requests are sent)
         max_retries = 3
--- a/backend/climateconnect_api/management/commands/osm_data_retrieval_scripts/create_osm_class_data.py
+++ b/backend/climateconnect_api/management/commands/osm_data_retrieval_scripts/create_osm_class_data.py
@@ -10,7 +10,7 @@ import requests
 from tqdm import tqdm
 
 LOCATIONS_URL = "https://nominatim.openstreetmap.org/lookup"
-HEADERS = {"User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36"}
+HEADERS = {"User-Agent": "ClimateConnect/1.0 (contact@climateconnect.earth)"}
 
 
 def discover_osm_class(osm_ids: list) -> tuple[dict, set] | None:
--- a/backend/climateconnect_api/management/commands/osm_data_retrieval_scripts/create_osm_type_data.py
+++ b/backend/climateconnect_api/management/commands/osm_data_retrieval_scripts/create_osm_type_data.py
@@ -6,7 +6,7 @@ import requests
 from tqdm import tqdm
 
 LOCATIONS_URL = "https://nominatim.openstreetmap.org/lookup"
-HEADERS = {"User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36"}
+HEADERS = {"User-Agent": "ClimateConnect/1.0 (contact@climateconnect.earth)"}
 
 
 def discover_osm_type(osm_ids: list) -> tuple[dict, set] | None:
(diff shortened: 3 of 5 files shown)


backend/location/models.py
```diff
@@ -1,9 +1,94 @@
...
 from django.contrib.gis.db import models
+from django.db import IntegrityError, transaction
 
 from climateconnect_api.models.language import Language
+from location.tasks import create_name_from_translation_data
+
+logger = logging.getLogger(__name__)
+
+
+class LocationManager(models.Manager):
```
HaraldWalker: Class and function don't seem to be used anywhere.

HaraldWalker (changes requested):
See comments. Looks like some changes are needed.

backend/climateconnect_api/management/commands/add_locale_to_location_translations.py
```diff
@@ -0,0 +1,305 @@
...
+    for i in tqdm(range(0, len(locs), batch_size)):
+
+        batch_locations = locs[i : i + batch_size]
+        osm_ids = set()
+        unique_translations = {}
+
+        for loc in batch_locations:
+            if loc.osm_id and loc.osm_type:
+                osm_type = loc.osm_type[0].upper()
+                if osm_type is None:
```
HaraldWalker: Dead None check on a value that can never be None. Line 97 already checks and `upper()` would throw an error if it is not something it can convert.

backend/climateconnect_api/management/commands/add_locale_to_location_translations.py
```diff
@@ -0,0 +1,305 @@
...
+from location.models import Location, LocationTranslation
+from location.utility import format_location_name
+from location.utility import format_translation_data
+
+NOMINATIM_DETAILS_URL = "https://nominatim.openstreetmap.org/lookup"
+CUSTOM_USER_AGENT = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36"
+MAPPING_TABLE_PATH = Path(__file__).parent / "osm_lookup_tables" / "mapping.csv"
+
+
+def load_osm_mapping() -> dict[str, list[int]]:
```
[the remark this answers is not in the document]
KathiAurina: you are right, I removed it, this was part of an older implementation, now this is done through Django directly without any mapping tables whatsoever

backend/location/tasks.py
```diff
@@ -0,0 +1,107 @@
...
+                )
+                continue
+
+            address = data[0].get("address", {})
+            translation_data["city_translation"] = (
+                address.get("city") or address.get("town") or address.get("village")
+            )
+            translation_data["state_translation"] = address.get("state")
+            translation_data["country_translation"] = address.get("country")
+            translation_data["name_translation"] = data[0].get("localname")
```
HaraldWalker, line 71: As previously mentioned we have to use our own logic to get the name. 
`translation_data["name_translation"] = format_location_name(data[0]).get("name")`

backend/location/tasks.py
```diff
@@ -0,0 +1,107 @@
...
+
+        except requests.exceptions.RequestException as e:
+            logger.error(
+                f"error while retrieving translation data from nominatim for location id {instance.id}: {e}"
+            )
+            raise self.retry(exc=e, countdown=60 * (self.request.retries + 1))
+
+        if not translation_data.get("name_translation"):
+            formatted_translation = format_translation_data(translation_data)
+            translation_data["name_translation"] = format_location_name(
```
HaraldWalker, line 81: `format_location_name` needs the full original nominatim response to create the correct result. With formatted_translation is does not return the expected name.

backend/climateconnect_api/management/commands/add_locale_to_location_translations.py
```diff
@@ -0,0 +1,278 @@
...
+            # this is for the locations in palestine (id=2667, localname = None,
+            # the only address tag that is provided is "disputed")
+            if translation_data["name_translation"] is None:
+                translation_data["name_translation"] = address.get("disputed")
+
+            for loc in matching_locations:
+                name = translation_data["name_translation"]
+                if not name:
+                    formatted_translation = format_translation_data(translation_data)
+                    formatted_name_data = format_location_name(formatted_translation)
```
HaraldWalker, line 179: Same issue as in task.py. `format_location_name` needs the nominatim result, not something that is already processed. And not as fall back but as default to get the translated name.

backend/location/utility.py
HaraldWalker, line 242: Could define TypedDict as this function expects a specific format. Would have caused a type error when passing something different to it. 

```
class NominatimResult(TypedDict):
    type: str           # required — format_location_name uses this
    display_name: str   # required — used for country names
    address: NominatimAddress  # required — used for city/state/country

    # lat, lon, place_id, licence, osm_id, boundingbox, extratags, ...
    # all omitted — they're in the real response but we don't declare them
    # as we don't need them

class NominatimAddress(TypedDict, total=False):
    city: str
    town: str
    village: str
    state: str
    country: str
    country_code: str
    # etc. — only declare what your code actually reads
```

backend/climateconnect_api/management/commands/add_locale_to_location_translations.py
```diff
@@ -0,0 +1,278 @@
...
+from location.models import Location, LocationTranslation
+from location.utility import format_location_name
+from location.utility import format_translation_data
+
+NOMINATIM_DETAILS_URL = "https://nominatim.openstreetmap.org/lookup"
+CUSTOM_USER_AGENT = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36"
+MAPPING_TABLE_PATH = Path(__file__).parent / "osm_lookup_tables" / "mapping.csv"
+
+
+def create_name_from_translation_data(
```
HaraldWalker, line 28: not used and should not be used.

backend/climateconnect_api/management/commands/add_locale_to_location_translations.py
```diff
@@ -0,0 +1,278 @@
...
+            )
+            if not matching_locations.exists():
+                print(
+                    f"warning: no Location found for OSM identifiers: type={osm_type}, id={osm_id}"
+                )
+                continue
+
+            address = result.get("address", {})
+            translation_data = {
+                "name_translation": result.get("localname"),
```
HaraldWalker, line 163: needs to use format_location_name
`translation_data["name_translation"] = format_location_name(result).get("name")`

backend/location/tasks.py
HaraldWalker: Correction: The function doesn't exactly use the raw nominatim response but it needs a lot of its data. The format is slightly different (e.g. no nested address object). The current function uses the data as it is now being sent by the frontend, for instance when changing the location on a user profile page.  Best to create a dict based on what the function actually expects.

katharinaauer: I found out, that the format_location_name function does indeed use and accept the raw nominatim response and fixed/tested the code accordingly

Issue #1718, opened 2026-01-06
Bug: Type error when switching language on a project page


HaraldWalker:
**Describe the bug**
When switching the language on a project page it throws a type error.

**To Reproduce**
1. Go to a project page
2. Click on other language
3. See error in console (or on screen when in dev mode)

**Expected behavior**
No error, language is changed as expected

**Additional context**
In production the user is barely impacted, for a flash second it shows an error but then renders the page correctly. In dev mode you see it more clearly. 
Check the PR for the next.js upgrade. I had fixed same/similar errors in one of the last commits. 
Not critical since user can still use the application and change language and only seems to happen on this page.

Root cause, Next.js and Reaact upgrade. This React version is more picky. It tries to call a function but that function doesn't exist.

PR #1720, opened 2026-01-07
Fix for type error destroy is not a function

HaraldWalker:
## Checked the following
- [X] Pages affected by this change work on mobile
- [X] Pages affected by this change work when logged out
- [X] Pages affected by this change work when logged in
- [X] Pages affected by this change work with custom theme and default theme
- [X] Run within `/frontend`: `yarn format && yarn lint`

## What and Why
Fix for #1718

### DIFF
--- a/frontend/src/components/project/ProjectTeamContent.tsx
+++ b/frontend/src/components/project/ProjectTeamContent.tsx
@@ -66,8 +66,16 @@ export default function TeamContent({ project, handleReadNotifications, hubUrl }
   const texts = getTexts({ page: "project", locale: locale });
   const classes = useStyles();
 
-  useEffect(async () => {
-    await handleReadNotifications(NOTIFICATION_TYPES.indexOf("project_join_request_approved"));
+  useEffect(() => {
+    const readNotifications = async () => {
+      try {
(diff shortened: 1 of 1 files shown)


HaraldWalker:
Explanation of the error root cause:
https://www.dhiwise.com/post/how-to-fix-the-error-destroy-is-not-a-function-in-react

frontend/src/components/project/ProjectTeamContent.tsx
```diff
@@ -66,8 +66,16 @@ export default function TeamContent({ project, handleReadNotifications, hubUrl }
   const texts = getTexts({ page: "project", locale: locale });
   const classes = useStyles();
 
-  useEffect(async () => {
-    await handleReadNotifications(NOTIFICATION_TYPES.indexOf("project_join_request_approved"));
+  useEffect(() => {
```
HaraldWalker: Explanation:
This prevents the error, because the effect doesn't return a Promise. The code launches an async task inside the effect and uses void to ignore its Promise, so useEffect itself returns undefined (no cleanup by calling destroy). If you instead made the effect callback async (so it returns a Promise), React would receive a Promise where it expects a cleanup function — that mismatch can lead to runtime errors like destroy is not a function.

PR #1742, opened 2026-01-15
fixing some Route (next/router) linting warnings

fateme-ramezanpour:
## Checked the following


frontend/public/lib/dateOperations.ts
```diff
@@ -1,5 +1,4 @@
-import { format, isThisYear, isToday } from "date-fns";
-import { de, enUS } from "date-fns/locale";
```
fateme-ramezanpour: I tried to fix this line using automated linting (yarn format --fix), but it introduced an error. I've removed the line in the subsequent commit instead.

Issue #1683, opened 2025-11-24
Bug: Image crop dialog does't fit into small horizontal screen


HaraldWalker:
**Describe the bug**
On a small horizontal screen like on a tablet device the upload image dialog requires more height than typically available. This works for mobile devices in vertical mode but not in horizontal mode.

**To Reproduce**
1. Go to any page with image upload feature (like user profile edit)
2. Upload an image
3. Reduce the width of the window and notice of the crop dialog changes style. You can also use some browser features to simulate a mobile device in horizontal mode but you can see the effect also my just resizing the window.
4. See how to bottom with the slide get's cut off. 

**Expected behavior**
The style adjustment based on fold should only be applied on vertical mode. 

**Screenshots**
<img width="1052" height="711" alt="Image" src="https://github.com/user-attachments/assets/6064f401-1c2f-4106-8f3a-9386fc3e9710" />

https://github.com/user-attachments/assets/d0568b95-292d-461a-b446-bffe3fb362b1

**Desktop (please complete the following information):**
Originally seen on iPad but as described can also be reproduced in any browser.

PR #1777, opened 2026-02-10
fixed height problem on changing Avatar

fateme-ramezanpour:
## Description
This PR addresses issue [1683](https://github.com/climateconnect/climateconnect/issues/1683)
## Checked the following


frontend/public/lib/dateOperations.ts
```diff
@@ -1,5 +1,4 @@
-import { format, isThisYear, isToday } from "date-fns";
-import { de, enUS } from "date-fns/locale";
+import { format, isThisYear, isToday, de, enUS } from "date-fns";
```
HaraldWalker: I know `yarn format --fix` does this and it breaks the code. I then get module not found.

fateme-ramezanpour:
> Wondering if the crop slider can be fixed in the bottom of the modal while the image area is responsive. Maybe we can also make this modal a bit larger to give it more room.

removed the max-height, so the slider problem got fixed.

HaraldWalker (approved):
Looking good now. 👍🏻 

Let me merge it so I can coordinate the deployment.

PR #1795, opened 2026-02-21
Fix: Keep locale when redirecting from hub landing page to hub browse page

positiveimpact:
## The bug
When you visit a link to a hub landing page (e.g. `/de/hubs/kassel`): If no landing page exists, you get redirected to the hub browse page. In this case the locale is currently lost, so you get redirected to the english browse page (`/hubs/kassel/browse`)

## The fix
Now the locale is kept on redirect. `/de/hubs/kassel` now redirects to `/de/hubs/kassel/browse` instead of `/hubs/kassel/browse` if there is no landing page.


positiveimpact:
Thank you!

Issue #1782, opened 2026-02-11
Bug: Memeber skills filter throws application error


HaraldWalker:
**Describe the bug**
When trying to filter members on skills it throws a client application error in the browser.

The smae skills filter is working correctly when browsing on hub level.

**To Reproduce**
1. Go to 'https://climateconnect.earth/browse?&#members'
2. Click on 'Skills' filter button
3. See error

**Expected behavior**
On click it opens the skills filter and user can select skills and show members that match those skills.

PR #1813, opened 2026-03-05
Fix Bug - Member skills filter throws application error

HaraldWalker:
## Checked the following

## What and Why

Fix for #1782


frontend/public/data/possibleFilters.ts
```diff
@@ -73,20 +73,22 @@ const getSearchFilter = () => {
 
 const getIdeasFilters = (filterChoices, texts) => [...getLocationFilters(texts)];
```
Idontker: Are ideas still in use - not related to this pr?

Issue #1849, opened 2026-03-26
Member can see the events they registered for


HaraldWalker:
### User Story

As a platform user
I want to see all the events I registered for
So that have an overview, can make changes and don't forget about an event.

### Description & Context
- The user can see all upcoming event the user has registered for on the user profile page
- Out of scope: seeing past events, ical calendar feed

### Acceptance Criteria

- On the user profile page at `profiles/[user-slug] ` the events the user has registered for are shown.
  - The events are only visible to the logged in user. 
  - If the user does not have any registered events it shows a text (same behavior as for my projects on the same page)
  - The events are shown in a grid with project preview, just like the projects of the user. 
  - The user can then click on the event to see the details and change registration (normal preview card behavior)
  - Due to current limitations max. 12 upcoming events are shown.
  - The events are ordered by event start date in ascending order, starting at the current date.
  - Events that are in the past are not being shown
- On an event page the user registered for the user can 
  - see that they are already registered

### Technical Notes
[Technical specs](https://github.com/climateconnect/climateconnect/blob/master/doc/spec/20260309_1100_member_see_registered_events.md)

PR #1896, opened 2026-04-14
Registered event in user profile

fateme-ramezanpour:
## Checked the following

## What and Why

implements #1849


frontend/pages/profiles/[profileUrl].tsx
```diff
@@ -14,18 +14,26 @@ import UserContext from "./../../src/components/context/UserContext";
...
+    projectTypes,
+    hubThemeData,
+    registeredEvents,
+  ] = await Promise.all([
     getProfileByUrlIfExists(profileUrl, auth_token, ctx.locale),
     getOrganizationsByUser(profileUrl, auth_token, ctx.locale),
     getProjectsByUser(profileUrl, auth_token, ctx.locale),
     getProjectTypeOptions(ctx.locale),
     getHubTheme(hubUrl),
+    auth_token ? getRegisteredEvents(auth_token, ctx.locale) : null,
```
Comment (HaraldWalker) [inline frontend/pages/profiles/[profileUrl].tsx]: Medium: 
`getRegisteredEvents` is called on every SSR render where auth_token exists — including when an authenticated user browses someone else's profile. Because the endpoint filters by self.request.user, it returns the viewing user's own real registered events, not an empty result or an error. That personal data is then serialised into the SSR props for a profile page that has nothing to do with them, and silently discarded by the `isOwnAccount` guard in `ProfileRoot`. No security issue, but it is a real (if small) data fetch and payload that serves no purpose.

A simple fix: do not fetch in SSR at all and instead trigger a client-side fetch (inside `ProfileRoot`) after confirming `isOwnAccount`. Alternatively, accept it as a known limitation for now.

frontend/src/components/profile/ProfileRoot.tsx
```diff
@@ -232,6 +233,18 @@ export default function ProfileRoot({
           </Typography>
         )}
       </Container>
+      {isOwnAccount && (
```
HaraldWalker: High:
New section is not gated behind the EVENT_REGISTRATION feature toggle

The fix is straightforward — add the hook call to `ProfileRoot` and wrap the section:

```tsx
const { isEnabled } = useFeatureToggles();
const isEventRegistrationEnabled = isEnabled("EVENT_REGISTRATION");

// ...

{isOwnAccount && isEventRegistrationEnabled && (
  <Container ...>
    ...
  </Container>
)}
```

HaraldWalker (changes requested):
Backend is good, frontend needs some improvements. I forgot to mention explicitly in the specs that the new section on the profile page needs to be behind the feature toggle. Otherwise, users would now already see this new section before.

HaraldWalker (approved):
Looking good now. 👍🏻

Issue #1901, opened 2026-04-16
Show if a user is already registered for an event on browse projects and similar projects


HaraldWalker:
### User Story ###
As a user 
I want to see on the project browser page if I have registered already for an event
so that I can quickly identify what I registered for already

### Description & Context ###
- on the project browse page and similar projects sidebar when showing events (projects) with registration show 'Registered' instead of the register button if the user is already registered and the registration is still open    

### Technical notes ###
- avoid making the browse query slower by also fetching registration data
- suggestion: fetch the user's event registrations once and then match the projects in the project preview cards (be aware that paging might be an issue)

PR #1915, opened 2026-04-21
Change user registration button on browse and sidebar projects

fateme-ramezanpour:
## Description
This PR addresses issue [https://github.com/climateconnect/climateconnect/issues/1901](https://github.com/climateconnect/climateconnect/issues/1901)
## Checked the following


frontend/pages/projects/[projectId]/index.tsx
```diff
@@ -172,11 +172,16 @@ export default function ProjectPage({
...
   const [numberOfLikes, setNumberOfLikes] = useState(project?.number_of_likes);
   const [numberOfFollowers, setNumberOfFollowers] = useState(project?.number_of_followers);
-  const { CUSTOM_HUB_URLS, locale } = useContext(UserContext);
+  const { CUSTOM_HUB_URLS, locale, user } = useContext(UserContext);
   const texts = getTexts({ page: "project", locale: locale, project: project });
   const [showSimilarProjects, setShowSimilarProjects] = useState(true);
   const [projectTypes, setProjectTypes] = useState([]);
 
+  // Get registered event slugs from user profile for showing "Registered ✓" status
+  const registeredEventSlugs = React.useMemo(() => {
```
Comment (HaraldWalker) [inline frontend/pages/projects/[projectId]/index.tsx]: `import {useMemo} from 'react'`

frontend/src/components/hub/HubBrowsePage.tsx
```diff
@@ -313,6 +318,7 @@ export default function HubBrowsePage({
                 linkedHubs={linkedHubs}
                 isLocationHub={isLocationHub}
                 fromPage="hub"
+                registeredEventSlugs={registeredEventSlugs}
```
HaraldWalker: As `registeredEventSlugs` are available in the `UserContext` I am wondering if we need to pass it through to all the components that need it or if it would keep it simpler by just fetching it from the context where it is needed.

HaraldWalker:
The state of registered events doesn't get refreshed when the user registers or cancels for an event, so the browse page or similar projects list has outdated data and shows the wrong state.  The easiest might be to reload the my_profile data after event registration changes.

Add refreshUser to UserContext
In the provider, refreshUser calls /api/my_profile/ and updates user state
After registration/cancellation success in modals, call refreshUser()
Components can then access refreshed user?.registered_event_slugs directly from context.

HaraldWalker (changes requested):

doc/spec/20260420_1400_show_registered_status_on_browse_and_similar.md
```diff
@@ -0,0 +1,277 @@
...
+- On the **similar projects sidebar** (shown on individual event pages):
+  - Apply the same "Registered ✓" indicator for events the user is registered for.
+  - This provides context when browsing related events.
+
+- **Performance requirement**: The browse query must not be slowed down by fetching registration data per project. The solution must avoid N+1 queries or per-row registration checks.
+
+**Explicitly Out of Scope (this iteration):**
+- Showing registration status to unauthenticated users (not applicable — you must be logged in to register).
+- Showing how many events a user is registered for (aggregate count).
+- Real-time updates when registration status changes (acceptable for the user to refresh the page).
```
HaraldWalker: I disagree. For a good UX it is important that the user sees the effect of actions as confirmation.

frontend/pages/browse.tsx
```diff
@@ -77,7 +82,11 @@ export default function Browse({ filterChoices, hubs, initialLocationFilter, pro
             locale={locale}
             token={token}
           >
-            <BrowseContent filterChoices={filterChoices} allHubs={hubs} />
+            <BrowseContent
+              filterChoices={filterChoices}
+              allHubs={hubs}
+              registeredEventSlugs={registeredEventSlugs}
```
HaraldWalker: The current implementation threads registeredEventSlugs through multiple component layers:
browse.tsx → BrowseContent → ProjectPreviews → ProjectPreview → ProjectMetaData

Each component that needs to check registration status could directly access user.registered_event_slugs from context and compute the Set locally.

Benefits of accessing from UserContext directly:

1. Eliminated prop drilling - removes registeredEventSlugs from 5 component interfaces
2. Simpler code - components don't need to accept and pass through props they don't use
3. Consistency - all components that need user data already use UserContext
4. Less re-render risk - no additional props means less chance of unnecessary re-renders from prop changes

Suggested refactor pattern:

```
// In ProjectPreview.tsx - compute locally instead of receiving as prop
const { user } = useContext(UserContext);
const registeredEventSlugs = React.useMemo(() => {
  return new Set(user?.registered_event_slugs || []);
}, [user?.registered_event_slugs]);

// Then use registeredEventSlugs.has(project.url_slug) for the check
```

frontend/pages/projects/[projectId]/index.tsx
```diff
@@ -172,11 +172,16 @@ export default function ProjectPage({
   const [likingChangePending, setLikingChangePending] = useState(false);
   const [numberOfLikes, setNumberOfLikes] = useState(project?.number_of_likes);
   const [numberOfFollowers, setNumberOfFollowers] = useState(project?.number_of_followers);
-  const { CUSTOM_HUB_URLS, locale } = useContext(UserContext);
+  const { CUSTOM_HUB_URLS, locale, user } = useContext(UserContext);
   const texts = getTexts({ page: "project", locale: locale, project: project });
   const [showSimilarProjects, setShowSimilarProjects] = useState(true);
   const [projectTypes, setProjectTypes] = useState([]);
 
+  // Get registered event slugs from user profile for showing "Registered ✓" status
```
Comment (fateme-ramezanpour) [inline frontend/pages/projects/[projectId]/index.tsx:180]: Implemented prop drilling from index to fix the `ProjectSidebar` flashing bug. This ensures stable state synchronization during the initial render.

HaraldWalker (approved):
All good now!
Thanks 🙇🏻

Issue #1960, opened 2026-05-06
Register for an event with extra fields


HaraldWalker:
Event registration V2

### User Story ###
As a platform user  
I want to register for an event that collects additional information using additional custom fields
so I can attend the event and provide the right information

### Description & Context ###
- When registering for an event the user can enter values for additional fields.
- What fields are being shown depends on the event registration configuration. 
- The fields are shown in the order as specified in the event registration configuration. 
- In total max. 5 custom fields get shown
- The form validates that required custom fields are being filled in.
- When completing the registration the system stores the values in the platform.

As extra custom fields types we have:
- Checkbox that shows a description (formatted rich text) and if the field is required or not.
- Single select options with a title and 1-X options with a title and in the order as specified and if the field is required or not.
    
Note: 
If the modal is getting too small we can allow scrolling and should follow the guidelines for scrollable modals in Material Design (hint: the buttons need to be sticky at the bottom so that they are always visible). This might require some UI changs to version 1 of event registration.

More details can be found in #1880 and the tech specs for it.

PR #1977, opened 2026-05-12
Register event with extra field

fateme-ramezanpour:
## Description
This PR addresses issue [1960](https://github.com/climateconnect/climateconnect/issues/1960)

## Checked the following


frontend/src/components/project/RegistrationOptionSelectField.tsx
```diff
@@ -0,0 +1,75 @@
...
+import makeStyles from "@mui/styles/makeStyles";
+import { RegistrationField } from "../../types";
+
+const useStyles = makeStyles((theme: Theme) => ({
+  root: {
+    marginBottom: theme.spacing(2),
+    paddingLeft: "10px",
+  },
+  label: {
+    fontSize: "0.875rem",
```
HaraldWalker, line 21: Should avoid hard-coding font styles like this in components and use MUI theme typography when possible.

frontend/src/components/project/RegistrationCheckboxField.tsx
```diff
@@ -0,0 +1,85 @@
...
+  textBlock: {
+    flex: 1,
+    minWidth: 0,
+  },
+  required: {
+    color: theme.palette.error.main,
+    marginRight: theme.spacing(0.5),
+  },
+  descriptionHtml: {
+    fontSize: "0.875rem",
```
HaraldWalker, line 28: The ideal MUI pattern is to use **theme typography variants** (e.g. `<Typography variant="body2">`) rather than hardcoding font sizes. MUI's default `body2` variant already has `fontSize: "0.875rem"` (14px) — exactly what these components need. Using the variant also gives you the theme's full type scale, responsive behaviour, and colour token for free.

frontend/src/components/project/RegistrationOptionSelectField.tsx
```diff
@@ -0,0 +1,75 @@
...
+  RadioGroup,
+} from "@mui/material";
+import { Theme } from "@mui/material/styles";
+import makeStyles from "@mui/styles/makeStyles";
+import { RegistrationField } from "../../types";
+
+const useStyles = makeStyles((theme: Theme) => ({
+  root: {
+    marginBottom: theme.spacing(2),
+    paddingLeft: "10px",
```
HaraldWalker, line 18: The project consistently uses the MUI theme spacing system (`theme.spacing(N)`) for all padding/margin values. A raw pixel string breaks theme adaptability and is inconsistent with every other component in the project. Examples of the correct pattern: `ProjectMetaData.tsx` uses `theme.spacing(1.5)`, `EventRegistrationModal.tsx` uses `theme.spacing(2)`, etc.

The intent is to align radio buttons visually with the checkbox content, but this should use a value derived from the theme, e.g. `theme.spacing(1.25)` (10px at 8px base) or the same left-offset applied to the checkbox field itself.

**Recommendation**: Replace `"10px"` with a `theme.spacing(...)` value. If alignment with the checkbox is the goal, apply a consistent `paddingLeft` (or `marginLeft`) to *both* field types using the same theme-derived value, or let the `FormControlLabel` and `Checkbox` components handle their own indentation naturally through MUI defaults.

HaraldWalker (approved):
Approved but needs some UI improvements we can do in another iteration.

Issue #2044, opened 2026-06-09
Allow negative project rating values


HaraldWalker:
## Problem Statement

The project ranking system allows admins to manually adjust a project's position in the list via the `rating` field on the `Project` model. This field is added directly to the computed ranking score in `ProjectRanking.calculate_ranking()`.

Currently, `rating` is a `PositiveSmallIntegerField` (range 0–32767) with a default of `100`. This means:

- Every project starts with a **+100 baseline** contribution to its ranking score.
- An admin can **lower** a project's rating to reduce its boost (e.g. set it to `0`), but can never make it **negative** — i.e. can never push a project below the baseline of other projects.

### Why This Matters Now

Newly created upcoming events are dominating the project list. A brand-new event with zero engagement receives ~75 points from the `created_at` factor alone (weight 5 × recency score 15). An established project with real engagement (comments, likes, followers) totals ~74 points. The manual `rating` field is the only lever admins have to correct this, but since it can only go down to `0`, it cannot overcome the `created_at` advantage of new events.

Allowing **negative** `rating` values would let admins set a project's rating to e.g. `-50`, which subtracts 50 from its total ranking score — enough to sink unwanted projects below those with genuine engagement.

PR #2045, opened 2026-06-09
Allow negative project ratings

HaraldWalker:
## Checked the following

## What and Why

Implements #2044 2044


backend/organization/views/project_views.py
```diff
@@ -1493,7 +1493,7 @@ class ListFeaturedProjects(ListAPIView):
 
     def get_queryset(self):
         return Project.objects.filter(
-            rating__lte=99, is_draft=False, is_active=True
+            rating__gte=0, rating__lte=99, is_draft=False, is_active=True
```
HaraldWalker, line 1496: This is actually unused code - tech dept. 
I added a spec file to clean this old API endpoint up.

Issue #2033, opened 2026-06-03
Filter Similar Projects by Hub Context


HaraldWalker:
## Overview

Enhance the similar projects recommendation in the project sidebar so that when a user is navigating within a climate hub context, only projects from that same hub are shown — rather than similar projects from the entire platform regardless of hub membership.

## Problem Statement

Currently, the "Similar Projects" sidebar on project detail pages (`GET /api/projects/{slug}/similar/`) returns globally similar projects without considering the user's hub context. When a user is browsing inside a climate hub (e.g., `?hub=berlin`), projects from unrelated cities or sectors appear alongside relevant local ones, diluting the hub-centric discovery experience that the rest of the platform already supports.

This inconsistency is especially noticeable for **location hubs** and **custom hubs**, where users expect to see work relevant to their regional or organizational community.

## User Stories

- As a user browsing a project within the Berlin hub, I want to see similar projects that are also in Berlin so I can discover local climate action.
- As a user browsing a project within a custom hub (e.g., Prio1), I want to see similar projects associated with that hub.
- As a user browsing a project without any hub context, I should continue to see globally similar projects (no behaviour change).

## Proposed Solution

Filter the `GET /api/projects/{slug}/similar/` response by the hub supplied via the `?hub=` query parameter, applying the same hub-location-to-project filtering logic already used by `ListProjectsView`.

The frontend already passes `?hub=` to this endpoint for serializer context purposes, so no frontend code changes are required — the backend simply needs to act on that parameter at the queryset level.

PR #2050, opened 2026-06-10
Filter similar projects by hub context

HaraldWalker:
## Checked the following

## What and Why

Implements #2033


backend/organization/views/project_views.py
```diff
@@ -250,64 +249,7 @@ def get_queryset(self):
                 )
 
         if "hub" in self.request.query_params:
-            # retrieve hub and its parents
```
HaraldWalker, line 253: Extracted into utility so we can re-use it in other places. The behavior of the get project enpoint did not change.

PR #2060, opened 2026-06-15
Mui v 6

fateme-ramezanpour:
## Checked the following


frontend/src/components/editProject/EditProjectOverview.tsx
HaraldWalker: Still using old Grid, should now be Grid2

frontend/src/components/editProject/EditProjectOverview.tsx
```diff
@@ -461,7 +461,7 @@ const InputSectors = ({
           />
         ))}
         <Grid container>
-          <Grid xs={12} sm={8} md={5} lg={5} item>
+          <Grid size={{ xs: 12, sm: 8, md: 5, lg: 5 }} item>
```
HaraldWalker: uses already the Grid2 size prop while the item is a Grid v1 prop. Should use Grid2 only (see previous comment about import). This mix of grid v1 and v2 would directly have impact on how this component gets rendered.

frontend/package.json
```diff
@@ -24,12 +24,13 @@
...
-    "@mui/styles": "^5.11.13",
-    "@mui/x-data-grid": "6",
-    "@mui/x-date-pickers": "^6.0.3",
+    "@mui/icons-material": "^6.5.0",
+    "@mui/material": "^6.5.0",
+    "@mui/styles": "^6.5.0",
+    "@mui/x-data-grid": "^7.0.0",
+    "@mui/x-date-pickers": "^7.0.0",
     "@next/bundle-analyzer": "^15.5.19",
+    "@react-google-maps/api": "^1.13.0",
```
HaraldWalker: This dependency is back. We had just removed it.

frontend/package.json
```diff
@@ -40,6 +41,7 @@
     "@tiptap/extension-link": "3.26.0",
     "@tiptap/extension-table": "3.26.0",
     "@tiptap/extension-text-align": "^3.26.0",
+    "@tiptap/extensions": "3.26.0",
```
HaraldWalker: Not imported anywhere

frontend/src/components/shareProject/CheckboxFieldEditor.tsx
```diff
@@ -41,7 +41,7 @@ const EXTENSIONS = [
   }),
   LinkBubbleMenuHandler,
   CharacterCount.configure({ limit: CHARACTER_LIMIT }),
-];
+] as any;
```
HaraldWalker: after
`import type { Extensions } from "@tiptap/core";`
this should be the type and that should avoid the need for the casting below. 

Same applies to the PorganizerMessageEditor.tsx change.

HaraldWalker (changes requested):
Most of it is correct. 
One critical issue and two minor improvements.

frontend/package.json
fateme-ramezanpour: I added this package because I was getting this error:
Screenshot 2026-06-23 at 11 18 11 AM

fateme-ramezanpour: I removed `@tiptap/extensions `package.
I had a yarn.lock conflict with master and removed .lock file, which threw the error above. I fixed it by restoring the lockfile from master.

HaraldWalker (approved):

PR #2069, opened 2026-06-18
Fix failing build

HaraldWalker:
## Checked the following

## What and Why

CI / CD fixes
- Addresses a known Microsoft Azure issue that broke the build
- Fixes the sentry integration for the frontend. Needed to get useful error reports.

### DIFF
--- a/.github/workflows/backend_tests.yml
+++ b/.github/workflows/backend_tests.yml
@@ -49,7 +49,9 @@
          uses: actions/checkout@v6
 
        - name: Install GDAL
-         run: sudo apt-get update && sudo apt-get install -y gdal-bin libgdal-dev
+         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/backend_tests.yml
```diff
@@ -49,7 +49,9 @@
          uses: actions/checkout@v6
 
        - name: Install GDAL
-         run: sudo apt-get update && sudo apt-get install -y gdal-bin libgdal-dev
+         run: |
+           sudo rm -f /etc/apt/sources.list.d/azure-cli.sources
```
HaraldWalker, line 53: Totall ridiculous error. The ubuntu image has azure-cli by default and then apt-get update fails with an error trying to update azure-cli. We don't need it in this workflow, so it is save to remove it and that fixes the error on update.

Issue #1630, opened 2025-10-02
Bug: Bug: Losing hub info when going to a sponsor organization


HaraldWalker:
**Describe the bug**
When going to a sponsor organization within a hub, we leave the hub context and lose the styling and navigation.

**To Reproduce**
1. Go to
 https://climateconnect.earth/projects/frederick-und-seine-mauseabenteuer?hub=wuerzburg 
or https://climateconnect.earth/en/hubs/wuerzburg/browse
2. Click on one of the 'is supported by' organizations.

**Expected behavior**
The user stays with the ClimateHub

Related issue #1588

PR #2096, opened 2026-06-30
Losing hub info supporting org

fateme-ramezanpour:
## Description
This PR addresses issue [1630](https://github.com/climateconnect/climateconnect/issues/1630)
## Checked the following


HaraldWalker (approved):
Looking good. 

At some point I'd like to refactor how we handle the hub context so that we don't have to do this chain of passing through props and maybe automatically add it to links to avoid bugs like this.

PR #2105, opened 2026-07-02
proof of concept for location iq integration for autocomplete requests

KathiAurina:
## Checked the following


backend/Makefile
```diff
@@ -19,6 +19,12 @@ celery_beat:
 	$(info )
 	celery -A climateconnect_main beat -l INFO
 
+celery_lookup:
```
HaraldWalker, line 34: on staging and production the application uses start_backend.sh, here we are now missing the celery lookup que worker

backend/Makefile
```diff
@@ -19,6 +19,12 @@ celery_beat:
 	$(info )
 	celery -A climateconnect_main beat -l INFO
```
HaraldWalker, line 30: also start the lookup celery here so that a developer only has to run `make celery`. Otherwise the platform will not work properly. We can keep the new `celery_lookup` for when we only want to work on the lookup feature.

backend/climateconnect_main/settings.py
```diff
@@ -357,9 +403,98 @@
...
+# doc/spec/20260720_1400_locationiq_rate_limited_queue_design.md.
+CELERY_BROKER_TRANSPORT_OPTIONS = {
+    "socket_connect_timeout": 2,
+    "socket_timeout": 2,
+}
 NOMINATIM_LOOKUP_URL = "https://nominatim.openstreetmap.org/lookup"
 CUSTOM_USER_AGENT = "ClimateConnect/1.0 (contact@climateconnect.earth)"
 
+# LocationIQ autocomplete
+LOCATIONIQ_API_KEY = env("LOCATIONIQ_API_KEY", "")
```
HaraldWalker, line 449: Needs to be added to `initial_dev_setup.sh`

backend/climateconnect_main/settings.py
```diff
@@ -357,9 +403,98 @@
...
+# and the fallback is what runs whenever LocationIQ is skipped, which includes
+# the case where the LOCATIONIQ_AUTOCOMPLETE toggle can't be read at all (a DB
+# blip degrades to default=False). OSM's usage policy allows 1 req/s per
+# application from one IP, so while the toggle is off in production this should
+# be "1/s" — see the risk table in
+# doc/spec/20260804_1202_locationiq_feature_toggle_and_result_caching.md.
+# Lowering it is NOT a one-line change: LOCATIONIQ_SENTINEL_TTL_S and
+# LOCATIONIQ_STALE_PENDING_S below are both derived from this rate and have to
+# be re-derived with it.
+LOCATIONIQ_MAX_RATE = "2/s"
```
HaraldWalker, line 470: And if we change to a different plan we can adjust the value here 👍🏻

backend/climateconnect_main/settings.py
```diff
@@ -357,9 +403,98 @@
...
+LOCATIONIQ_PENDING_CAP = 16  # max distinct in-flight lookups before 503 (backpressure)
```
HaraldWalker: Those settings (LOCATIONIQ_RECLAIM_LOCK_S, LOCATIONIQ_SENTINEL_TTL_S, LOCATIONIQ_STALE_PENDING_S, LOCATIONIQ_PENDING_CAP, etc.) are all about the queue/cache infrastructure — the rendezvous pattern, the pending sentinel lifecycle, the LRU eviction cap. They would be identical regardless of whether the backend was calling LocationIQ or some other geocoding API.

The LOCATIONIQ_ prefix is misleading — it suggests these are provider-specific knobs when they are actually generic infrastructure settings for the autocomplete proxy. Only a handful are genuinely LocationIQ-specific:
• LOCATIONIQ_API_KEY — provider credential
• LOCATIONIQ_AUTOCOMPLETE_URL — provider endpoint
• LOCATIONIQ_TIMEOUT — provider-specific timeout
• LOCATIONIQ_MAX_RATE — the 2 req/s limit that LocationIQ imposes
• LOCATIONIQ_DAILY_BUDGET — quota management for the paid API

Everything else (SENTINEL_TTL_S, STALE_PENDING_S, RECLAIM_LOCK_S, PENDING_CAP, RESULT_TTL_S, MAX_CACHE_AGE_S, CACHE_MAX_ENTRIES, STATS_TTL_S) belongs to the proxy layer and should arguably be prefixed AUTOCOMPLETE_ or LOCATION_PROXY_.

HaraldWalker (changes requested):
In general solid first round. Needs some improvements. See comments and what I shared via Slack with you.

HaraldWalker (approved):
The design is sound, the code is clean, test coverage is excellent and it works as specified. 👍🏻

Issue #2104, opened 2026-07-02
A super-admin can delete an organisation


HaraldWalker:
### User Story
As a super-admin of an organisation
I want to be able to delete an organisation
So that I can clean up and not depend on support for doing it for me

### Requirements
- When a super admin is editing an organisation there is a delete option (similar to projects) allowing the user to delete the organisation
- It shows a confirmation dialog where the user has to confirm the action. 
- If the organisation has projects it shows a warning and the user **can not** delete. The backend API also checks this. Warning text: The organisation has X projects. Please first delete these projects or assign them to other organisations or users. 
- It cleans up organisation members, followers, likes etc (Django should handle this automatically)
- After delete it goes back to the organisation tab and shows a confirmation message.

PR #2126, opened 2026-07-13
Delete organization by supper admin

fateme-ramezanpour:
## Checked the following

## What and Why

Implements #2104


HaraldWalker (changes requested):
a suggested change, see review document

HaraldWalker (changes requested):
For a better UX experience the check if the organisation has projects and a message should happen before the user confirms the delete action. The little API error text in the corner is easily missed. The confirmation dialog should already show the text and instead of No/Yes show a cancel button. On the project page we were already showing the projects, so can either use the count from there or call the API on the edit page to get the count (or make a new endpoint with count). 

There is a mismatch between the count the backend returns if there are projects and the projects the user can actually see on the organisation page. 

The count on the delete check also counts draft and inactive (`is_active=False`) projects. These are not visible on the organisation page, so the user would be stuck. The draft ones the user could find on his user profile (only if the same user is also the owner). But I would argue that draft projects and inactive projects don't need to be preserved, so we can use the count that the list projects endpoint returns (excluding draft and inactive) and draft/inactive projects can be deleted together with the organisation.

HaraldWalker (approved):
All previous review issues have been addressed. It is good now!

Issue #2123, opened 2026-07-13
Improve project detail page navigation and extend it to organisations and members


HaraldWalker:
**Is your feature request related to a problem? Please describe.**
The project detail page has a 'Go back' navigation. This now always goes back to the project browse page (either the global one or the one of the hub). But that's not always where the user came from and especially once we introduce the event calendar, we would want to take the user back to the event calendar. 

Additionally the same go back button should work also for organisations and members, so after visiting those detail pages, the user can go back also here to where the user came from. No hard-coding on project browse as back url!

**Describe the solution you'd like**
- Add the sub-navigation bar as seen on project detail page also to the organisation detail page
- Add the sub-navigation bar as seen on project detail page also to the member profile page
  - Do not show the sub-navigation when the user goes to My profile (My projects/My Organisations).  
- The 'Go back' should actually go back to where the user came from, so act the same as a browser back button.

E.g. 
User profile page shows the projects of the user, user clicks on the projects, then go back takes the user back to the projects on the user profile page. 
A project detail page shows that a project was created by user A, click on the link to see the user profile, then go back and it shows the project detail page again.

PR #2129, opened 2026-07-15
Go back navigation

fateme-ramezanpour:
## Description
This Pr addresses issue [2123](https://github.com/climateconnect/climateconnect/issues/2123)

## Checked the following


HaraldWalker (approved):
Looking good. 

There is one edge case without a 'nice' solution. When the user changes language, the route/history back then takes you back to the previous language path. 
But sounds like an edge case where a user changes languages and then tries to go back multiple times. That user can then still change language again. 
Could theoretically detect if go back target = current page in other language and in that case go back to the default pages.

Issue #2130, opened 2026-07-15
Refactor handling of hub context


HaraldWalker:
## Problem Statement

Climate Connect shows projects, organisations, members, and dedicated hub landing pages. All of these can be scoped to a specific climate hub, but the frontend has no clean concept of "the active hub". Instead, the hub is identified in two parallel ways (the `?hub=<slug>` query parameter on non-dedicated pages and the `/hubs/<slug>/...` path on dedicated hub pages), and that information is then drilled through many layers of components under inconsistent names (`hubUrl`, `hubPage`, `hubName`, `hubData`, `fromPage`).

The practical consequences we see today:

- **Lost hub context on navigation.** Because the `?hub=` query string is open-coded in ~40+ files via patterns like `hubUrl ? "?hub=" + hubUrl : ""`, it is easy to forget it on a new link. When it is missing, the user is silently dropped back to the global site even though everything they were looking at was hub-scoped.
- **Lost locale context on navigation.** The same pattern of open-coding `getLocalePrefix(locale) + "/path"` appears in 200+ places. When a developer forgets it, the user lands in the English site even though they were browsing in another language — another "silent" failure.
- **No single source of truth for the hubs list.** Every page that needs hubs calls `getAllHubs(locale)` in its own `getServerSideProps`, refetching on every navigation, and the result is passed down as props (sometimes as `hubs`, sometimes `allHubs`).
- **Inconsistent prop naming.** The same logical value is renamed as it goes down the tree — e.g. `hubUrl` (SSR) → `hubPage` (ProjectPage → ProjectPageRoot) → `hubUrl` again (descendants). On org pages the value is also stored as `hubName` and `hubData` inconsistently.
- **No reactive active-hub context.** `UserContext.hubUrl` is computed once in `pages/_app.tsx` from `router.query` and is the only app-wide source — but it carries only the slug string, no hub data, and component authors routinely read it from a URL snapshot rather than derive it where they need it.
- **Hub-data fetches are repeated per page.** `getHubData(slug, locale)` is called independently by every page that needs hub data, and `getHubTheme(slug)` is called separately on top of that. The two are never correlated.

[...]

PR #2136, opened 2026-07-16
Implement phase 3 of hubcontext refactor

HaraldWalker:
## Checked the following

## What and Why

Part 3 of #2130 

This part shows how to use the new concept and directly fixes a link bug where the hub context was lost because calling components forgot to set the hub property.

### DIFF
--- a/frontend/src/components/footer/Footer.tsx
+++ b/frontend/src/components/footer/Footer.tsx
@@ -1,4 +1,4 @@
-import { Box, Link, Theme, useMediaQuery } from "@mui/material";
+import { Box, Theme, useMediaQuery } from "@mui/material";
 import makeStyles from "@mui/styles/makeStyles";
 import FacebookIcon from "@mui/icons-material/Facebook";
 import FavoriteIcon from "@mui/icons-material/Favorite";
@@ -6,7 +6,7 @@ import GitHubIcon from "@mui/icons-material/GitHub";
 import InstagramIcon from "@mui/icons-material/Instagram";
 import YouTubeIcon from "@mui/icons-material/YouTube";
 import React, { useContext } from "react";
-import { getLocalePrefix } from "../../../public/lib/apiOperations";
+import AppLink from "../general/AppLink";
 import getTexts from "../../../public/texts/texts";
 import UserContext from "../context/UserContext";
 import SocialMediaButton from "../general/SocialMediaButton";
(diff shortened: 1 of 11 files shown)


frontend/src/components/organization/MiniOrganizationPreview.tsx
```diff
@@ -63,16 +63,13 @@ export default function MiniOrganizationPreview({
   nolink,
   doNotShowName,
   inline,
-  hubUrl,
 }: any) {
-  const { locale } = useContext(UserContext);
-  const queryString = hubUrl ? "?hub=" + hubUrl : "";
```
HaraldWalker, line 69: It was here but if no hubUrl was provided as prop it had no effect. 
The caller should not need to know. 
Even this component should not have to know how to construct hub compatible links. We can leave that to our AppLink component.

frontend/src/components/organization/MiniOrganizationPreview.tsx
```diff
@@ -63,16 +63,13 @@ export default function MiniOrganizationPreview({
...
   if (!nolink)
     return (
-      <Link
+      <AppLink
         className={className}
         color="inherit"
-        href={getLocalePrefix(locale) + `/organizations/${organization.url_slug}${queryString}`}
```
HaraldWalker, line 75: These string concatenations (harder to read, error-prone) are no longer needed).

frontend/src/components/general/LoginNudge.tsx
```diff
@@ -29,31 +29,26 @@ export default function LoginNudge({ whatToDo, fullPage, className }: Props) {
...
   const currentPath = router.asPath;
   const encodedRedirectUrl = encodeURIComponent(currentPath);
-  let path_to_signin = `${getLocalePrefix(locale)}/signin?`;
-  let path_to_signup = `${getLocalePrefix(locale)}/signup`;
-  path_to_signin += `redirect=${encodedRedirectUrl}`;
-
-  if (hub && hub !== "") {
-    path_to_signin += `&hub=${hub}`;
-    path_to_signup += `?hub=${hub}`;
-  }
```
HaraldWalker, line 44: Don't have to do this locale prefix and hub query parameter work any longer in all places that link somewhere. Less boilerplate code, fewer errors.

frontend/src/components/footer/Footer.tsx
```diff
@@ -163,15 +163,15 @@ const SmallFooter = ({
     >
       <Box className={classes.flexContainer}>
         <Box className={classes.leftBox}>
-          <Link href={getLocalePrefix(locale) + "/imprint"} color="inherit" underline="hover">
+          <AppLink href="/imprint" leaveHub color="inherit" underline="hover">
```
HaraldWalker, line 166: When we have global links that intentionally leave the hub context, set the leaveHub property. This property also makes it easy to find these places later.

Issue #2122, opened 2026-07-13
When switching hubs stay on the same tab context


HaraldWalker:
**Is your feature request related to a problem? Please describe.**
When a user who is looking at organisations or members is switching to a different hub or all places, the user is being redirected to the project browse page. This will be especially annoying when we introduce the event calendar page.

**Describe the solution you'd like**
When switching hub or going to all places stay in the current tab (sub-navigation) context (projects, organisations, members, event calendar). 

Example:
https://climateconnect.earth/de/hubs/potsdam/browse?&#organizations
then navigates to 
https://climateconnect.earth/de/hubs/erlangen/browse?&#organizations
or
https://climateconnect.earth/de/browse?&#organizations

**Additional context**
The event calendar doesn't use the anchor tab link but instead is a real page url but the hub url pattern is the same.
E.g. /hubs/kassel/events for a hub event calendar and /events for the global event calendar.

PR #2144, opened 2026-07-20
losing tab context after switching hub

fateme-ramezanpour:
## Description
This PR addresses issue [2122](https://github.com/climateconnect/climateconnect/issues/2122)
## Checked the following


frontend/src/components/indexPage/hubsSubHeader/HubsDropDown.tsx
HaraldWalker: On 'All locations' the tab or event page switch is not implemented. Also here when the user goes from Potsdam event to all locations the user would like to stay on the events calendar.

HaraldWalker (changes requested):
Implementation not complete. See comment.

HaraldWalker (approved):
Looking good now! 👍🏻

Issue #2175, opened 2026-07-30
Bug: Frontend calls the am_i_following API endpoint for guest users on organisation pages


HaraldWalker:
**Describe the bug**
The frontend makes a call to the am_i_following API endpoint for users that are not logged in, causing a large amount of error logs in the backend. 

**To Reproduce**
1. As guest user go to an organisation detail page.
2. See error in browser console and backend logs

**Expected behavior**
the `/api/organizations/[org slug]/am_i_following/` endpoint is only being called for authorised (loged in) users

PR #2190, opened 2026-08-04
Call am_i_following api when user is login

fateme-ramezanpour:
## Description
this PR addresses issue [2175](https://github.com/climateconnect/climateconnect/issues/2175)
## Checked the following

### DIFF
--- a/frontend/pages/organizations/[organizationUrl].test.tsx
+++ b/frontend/pages/organizations/[organizationUrl].test.tsx
@@ -1,0 +1,33 @@
+import { getIsUserFollowing } from "../../public/lib/organizationOperations";
+import { apiRequest } from "../../public/lib/apiOperations";
+
+jest.mock("../../public/lib/apiOperations", () => ({
+  apiRequest: jest.fn(),
+}));
+
+const mockedApiRequest = apiRequest as jest.MockedFunction<typeof apiRequest>;
+
+describe("getIsUserFollowing", () => {
+  beforeEach(() => {
+    mockedApiRequest.mockReset();
+  });
+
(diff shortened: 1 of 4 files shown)


HaraldWalker (approved):
Looking good!

frontend/pages/organizations/[organizationUrl].tsx
```diff
@@ -100,7 +101,7 @@ export async function getServerSideProps(ctx) {
     getMembersByOrganization(organizationUrl, auth_token, ctx.locale),
     getOrganizationTypes(),
     getRolesOptions(auth_token, ctx.locale),
-    getIsUserFollowing(organizationUrl, auth_token, ctx.locale),
+    auth_token ? getIsUserFollowing(organizationUrl, auth_token, ctx.locale) : null,
```
Comment (HaraldWalker) [inline frontend/pages/organizations/[organizationUrl].tsx:104]: this guard is duplicate. The function already handles it when token is null

frontend/pages/organizations/[organizationUrl].test.tsx
```diff
@@ -0,0 +1,33 @@
+import { getIsUserFollowing } from "../../public/lib/organizationOperations";
+import { apiRequest } from "../../public/lib/apiOperations";
+
+jest.mock("../../public/lib/apiOperations", () => ({
+  apiRequest: jest.fn(),
+}));
+
+const mockedApiRequest = apiRequest as jest.MockedFunction<typeof apiRequest>;
```
Comment (HaraldWalker) [inline frontend/pages/organizations/[organizationUrl].test.tsx:8]: This new page test was the first page test and broke the build. Would require some additional next.js configuration.
 
But it is actually obsolete. The important code to test is in the function and the test should be in organizationOperations.test.ts.

Issue #1730, opened 2026-01-12
Remove the climate match feature from the backend


HaraldWalker:
**Describe the solution you'd like**
After the climate match feature is no longer being used by the front-end, remove its implementation from the backend. 

- clean up api, serializers, models, etc.

PR #2251, opened 2026-08-26
remove climate_match from backend

KathiAurina:
## Checked the following

## What and Why
#1730


HaraldWalker (approved):
Thanks for cleaning up! 👍🏻

PR #2271, opened 2026-09-07
Upgrade mui to v 7

fateme-ramezanpour:
## Description
This PR addresses issue [2241](https://github.com/climateconnect/climatehub/issues/2241)
## Checked the following


frontend/src/components/layouts/ContentImageSplitLayout.tsx
HaraldWalker, line 60: leftGridSizes and rightGridSizes can't be used like this in grid (grid2) and are actually obsolete as the size property already sets the values the only calling element uses.

frontend/package.json
```diff
@@ -25,11 +25,11 @@
     "@emotion/react": "^11.14.0",
     "@emotion/styled": "^11.14.1",
     "@floating-ui/dom": "^1.0.0",
-    "@mui/icons-material": "^6.5.0",
-    "@mui/material": "^6.5.0",
+    "@mui/icons-material": "^7.3.11",
+    "@mui/material": "^7.3.11",
     "@mui/styles": "^6.5.0",
```
HaraldWalker, line 30: I see, there is no Mui styles 7.x and removing it should be a separate task.

HaraldWalker (changes requested):
One critical issue and one improvement.

frontend/package.json
fateme-ramezanpour: yes, we can only remove it after all `makeStyles` and `withStyles` in our code are refactored.

HaraldWalker: I added #2290 as task.

HaraldWalker (approved):
Looking good

PR #2287, opened 2026-09-14
Allow image upload via drag & drop

fateme-ramezanpour:
## Description
This PR addresses issue [2240](https://github.com/climateconnect/climatehub/issues/2240)
## Checked the following


HaraldWalker:
There is an issue with the drag leave. It doesn't properly clean up the highlight of the drag zone. 
Screenshot 2026-09-17 at 09 44 34

HaraldWalker:
The highlight frame is 1-2 pixels inside of the image. 
Screenshot 2026-09-17 at 09 57 08

HaraldWalker:
Change the drag highlight styling to from MUI default blue (#1976d2, solid) to project primary teal (#207178, dashed)
Screenshot 2026-09-17 at 10 00 03

HaraldWalker:
On the project edit form we already have a dashed line, so the drag highlight adds a double highlight. 
Screenshot 2026-09-17 at 10 47 35
Instead of showing the extra highlight we can toggle the existing one.

HaraldWalker:
As the avatar is getting a round crop it would be nice to show a round drag area highlight. E.g. like this:
Screenshot 2026-09-17 at 11 02 29

frontend/src/components/shareProject/AddPhotoSection.tsx
```diff
@@ -72,8 +75,7 @@ export default function AddPhotoSection({
     handleSetOpen({ [dialogName]: true });
   };
 
-  const onImageChange = async (event) => {
-    const file = event.target.files[0];
+  const handleImageFile = async (file: File) => {
     if (!file || !file.type || !ACCEPTED_IMAGE_TYPES.includes(file.type)) {
       alert(texts.please_upload_either_a_png_or_a_jpg_file);
```
HaraldWalker, line 80: Please don't use standard windows.alerts. Those are ugly and not in line with MUI or Material Design. 

We could show an error in the dropzone itself but that won't work with the avatar drop zone as it is too small. 

Best second option is to use a Snackbar and we already to this on other forms to show non-critical errors.

HaraldWalker (changes requested):
Functional ok but needs some UX/UI improvements.

